PR #105, opened 2023-08-24
remember saveable DateTimeFields

sorenoid:
add savers as needed to save state across orientation changes for DateTimeFields. One thing I figured out was that when a state object holds a `State<>` property, that state object needs to be remembered with a saver so that we don't remember the `State<>` objects individually -- a new one gets created when the app goes through pause/resume.

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/DateTimeField.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/DateTimeField.kt
@@ -54,9 +54,8 @@ import androidx.compose.ui.unit.dp
 import com.arcgismaps.toolkit.featureforms.R
 import com.arcgismaps.toolkit.featureforms.components.datetime.picker.DateTimePicker
 import com.arcgismaps.toolkit.featureforms.components.datetime.picker.DateTimePickerInput
-import com.arcgismaps.toolkit.featureforms.components.datetime.picker.DateTimePickerState
 import com.arcgismaps.toolkit.featureforms.components.datetime.picker.DateTimePickerStyle
-import com.arcgismaps.toolkit.featureforms.components.datetime.picker.dateTimePickerStateSaver
+import com.arcgismaps.toolkit.featureforms.components.datetime.picker.rememberDateTimePickerState
 import com.arcgismaps.toolkit.featureforms.utils.PlaceholderTransformation
 
 @Composable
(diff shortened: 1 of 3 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePickerState.kt
```diff
@@ -246,12 +243,33 @@ internal fun DateTimePickerState(
...
+/**
+ * a StateSaver for the DateTimePickerState.
+ *
+ * @param initialValue the value needed to initialize a DateTimePickerState
+ * @return a StateSaver
+ * @since 200.3.0
+ */
+internal fun dateTimePickerStateSaver(initialValue: Long?): Saver<DateTimePickerState, Any> = listSaver(
+    save = {
+        listOf(it.pickerStyle, it.minDateTime, it.maxDateTime, initialValue, it.label, it.description, it.activePickerInput.value)
```
kaushikrw, line 267: I think it would be easy to read if these arguments and the ones below at L:273 were formatted to wrap in a new line instead.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/DateTimeField.kt
```diff
@@ -64,12 +67,15 @@ internal fun DateTimeField(
     val isEditable by state.isEditable
     val isRequired by state.isRequired
     val epochMillis by state.value
-    val pickerStyle = if (state.shouldShowTime) {
+    val shouldShowTime = rememberSaveable {
```
kaushikrw, line 70:
```suggestion
    val shouldShowTime = rememberSaveable {
```
Does this need to be remembered?

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/DateTimeField.kt
```diff
@@ -192,14 +198,15 @@ internal fun DateTimeField(
     }
 
     if (openDialog) {
-        val pickerState = remember {
+        val pickerState = rememberSaveable(saver = dateTimePickerStateSaver(epochMillis)) {
```
kaushikrw, line 201: It would be good to provide a default `rememberDateTimePickerState(..)` function that does the saving internally.

kaushikrw (commented):
This looks good! Just some minor comments below.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/DateTimeField.kt
sorenoid: nope. at one point it seemed necessary, but not now.

kaushikrw (approved):
LGTM!

PR #106, opened 2023-08-25
date time fixes

sorenoid:
apollo issue [#138](https://devtopia.esri.com/runtime/apollo/issues/138)

Working with the date picker is a bit complicated. It expects an input datetime epoch millseconds, but will provide back the datetime as of midnight of the selected date. So if an attribute has a date time other than midnight, we open the picker, then tap ok, the attribute's value will be set incorrectly.

Furthermore, we already know that the date picker operates in millis, and also displays dates that correspond to those milliseconds. So to correctly display the date in the current, non-UTC timezone, that millis passed to it cannot be epoch milliseconds, they must be epoch milliseconds augmented by an offset. Likewise the value it eventually produces need to have the offset taken away by subtracting it.

This PR mainly makes our internal `DateTime` class always hold UTC values
* it is renamed to `UtcDateTime`.
* it provides getters and factory functions that can be used to interact with the inputs and outputs of our DateTimePicker. 
* It adds an `epochMillis` field
* renames the `dateTime` field to `date` and also makes clear that the `date` field is in fact the millis at midnight, and so is settable from the output of the picker.
* new properties for use in populating the date time picker
* a new factory method for creating an instance from the output of the date time picker

Also
* several helper extensions on Long.


kaushikrw (approved):
Looks good, thanks for fixing this! Just some minor comments with naming.

PR #107, opened 2023-08-26
Featureforms micro-app update to cache portalitems

kaushikrw:
### Summary of changes

The microapp now loads a list of authenticated `PortalItems` from a specified `Portal` with a given folder name under the users content. It will also load any public portal items that do not need authentication. This is done by implementing an `ItemRepository` which can load both locally specified public portal items and authenticated items which are fetched over the network. Additionally, a synchronization mechanism exists which uses `room` database provider where all data fetched over the network are first synced into the local database. This means a single `flow` can be subscribed to to observe any new data.

A second layer of `PortalItemRepository` provides a mapping to the `ItemRepository` which also transforms the data into loaded `PortalItems`. This repository layer also employs a cache to store the `PortalItem`s for faster subsequent fetching. 

To refresh the data, to options are provided in a dropdown menu -
- Refresh - fetches new data from the network.
- Clear Data - clears the local cache and fetches new data from the network.

Note: The `PortalItemRepository` can be optimized further to also cache the `PortalItem` thumbnails which currently are loaded every time hence isn't as optimized as it could be.


microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/data/network/ItemRemoteDataSource.kt
```diff
@@ -0,0 +1,61 @@
...
+        const val portalUri = "https://www.arcgis.com"
+
+        /**
+         * Folder under the portal to fetch the portal items from.
+         */
+        const val portalFolder = "Apollo"
+    }
+
+    suspend fun fetchItemData(): List<ItemData> = withContext(dispatcher) {
+        return@withContext itemApi.fetchItems()
```
sorenoid, line 59: could just be

```suggestion
       itemApi.fetchItems()
```

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/data/PortalItemRepository.kt
```diff
@@ -0,0 +1,92 @@
...
+    private val dispatcher: CoroutineDispatcher,
+    private val itemRepository: ItemRepository,
+    private val itemCacheDao: ItemCacheDao,
+) {
+    // in memory cache of loaded portal items
+    private val portalItems: MutableMap<String, PortalItem> = mutableMapOf()
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    private val portalItemFlow: Flow<List<PortalItem>> =
+        itemRepository.observe().mapLatest {
```
sorenoid, line 31: what happens if the item repo grows so large that the first item is loaded and cached (i.e it is emitted by the call to `observe()`)  before the last item begins emitting? Will there be two emissions, one for that first item and one for the last item?

this is my only real concern with this PR. everything else is just nitpicks.

sorenoid (changes requested):
This looks really good. I have just one concern and some small nitpicks. My concern is about the `portalItemFlow` in the `PortalItemRepository`. Please see my question inline.

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/data/ItemRepository.kt
```diff
@@ -18,33 +18,54 @@
...
-        } else {
-            mutex.withLock { itemData }
-        }
+    suspend fun refresh() = withContext(dispatcher) {
+        // get local items
+        val localItems = refreshLocalItems()
+        // get network items
+        val remoteItems = remoteDataSource.fetchItemData()
+        // purge existing items and add the updated items
+        localDataSource.deleteAndInsert(localItems + remoteItems)
```
sorenoid, line 58: I see the `@Transaction` annotation on `deleteAndInsert`, which I assume means that the call is thread safe, in addition to indicating that it is a transactional "all or nothing" operation I just want to make sure that this is true and that the calls to `refreshLocalItems` and `remoteDataSource.fetchItemData()` are also thread safe?

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/di/PersistenceModule.kt
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * The provider of the ItemCacheDatabase.
+     */
+    @Singleton
+    @Provides
+    fun provideItemCacheDataBase(@ApplicationContext context: Context): ItemCacheDatabase {
+        return Room.databaseBuilder(
+            context.applicationContext,
```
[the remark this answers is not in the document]
kaushikrw: I think `context.applicationContext` is redundant, changed it to just `context`.

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/data/ItemRepository.kt
kaushikrw: Yes, that is true. All operations defined on the room `Dao` interfaces are thread-safe. Since the source of data is always reactive to database changes which are observed via flows I believe `datasource.refresh()` should be both thread and main safe. 

I think there is no need for any additional synchronization based on my testing. Perhaps there could be a mechanism to only perform a single `refresh` to prevent duplicate fetches but this operation would ultimately result in synchronized writes to the db and such 'single refresh' is implemented in the UI layer.

kaushikrw:
@sorenoid Thanks for the feedback. I merged in #110 and its ready for another look!

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/data/PortalItemRepository.kt
```diff
@@ -20,73 +31,129 @@ import kotlinx.coroutines.withContext
...
+            loadAndCachePortalItems(localItems + remoteItems)
+        }
     }
 
     /**
      * Returns the number of items in the repository.
      */
     suspend fun getItemCount(): Int = withContext(dispatcher) {
-        return@withContext itemRepository.getCount()
+        return@withContext itemCacheDao.getCount()
```
sorenoid, line 86:
```suggestion
    itemCacheDao.getCount()
```

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/data/PortalItemRepository.kt
```diff
@@ -20,73 +31,129 @@ import kotlinx.coroutines.withContext
...
+    private suspend fun createThumbnail(name: String, bitmap: Bitmap): String =
+        withContext(Dispatchers.IO) {
+            val thumbsDir = File("$filesDir/thumbs")
+            if (!thumbsDir.exists()) thumbsDir.mkdirs()
+            val file = File("${thumbsDir.absolutePath}/${name}.jpg")
+            file.createNewFile()
+            FileOutputStream(file).use {
+                bitmap.compress(Bitmap.CompressFormat.JPEG, 100, it)
+            }
+            return@withContext file.absolutePath
```
sorenoid, line 147:
```suggestion
    file.absolutePath
```

sorenoid (approved):
There are a couple small inconsistencies in the use or `return` in `withContext` that should be straightened out. Other than that this looks good. Please see my private message about the SDK work that needs to be merged before this can be merged.

PR #112, opened 2023-08-30
Reinstate onWindowFocusChanged in OAuthUserSignInActivity

hud10837:
Without this fix, pressing the close button on the custom tab or the back button while the OAuth prompt is displayed will result in a blank screen and the OAuth sign in will never complete, because these two actions don't trigger `onNewIntent` when they return back to `OAuthUserSignInActivity`.


eri9000 (approved):
LGTM 👍🏼

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/OAuthUserSignInActivity.kt
```diff
@@ -108,6 +109,24 @@ public class OAuthUserSignInActivity : ComponentActivity() {
         }
     }
 
+    override fun onWindowFocusChanged(hasFocus: Boolean) {
```
gunt0001, line 112: As discussed, lets add doc about this function being called when the close or back button is pressed.

gunt0001 (approved):
@hud10837 - looks good, one minore comment.

PR #114, opened 2023-09-01
now freeze fix

sorenoid:
when setting the value of the `dateTime` property of type`UtcDateTime` on our DateTimePickerState object, we were keying the m3 DatePickerState on the instance. This was causing some stale caching/remembering in the m3 DatePicker, I think. 

* removed the keying, and explicitly call `datePickerState.setSelection(...)` each time the DateTimePicker is recomposed. 

This fixes the freezing of the year picker when the now() button is pressed.

* fixed an issue where the initial value was always reset in the DateTimePickerState saver function.


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePicker.kt
```diff
@@ -132,6 +133,11 @@ internal fun DateTimePicker(
             DisplayMode.Picker
         )
     }
+    
+    //reset the selection if/when the dateTime changes
+    datePickerState.setSelection(dateTime.dateForPicker)
+    
+    //datePickerState.setSelection(dateTime.dateForPicker)
```
kaushikrw, line 140:
```suggestion
```

kaushikrw (approved):
Looks good!

PR #116, opened 2023-09-06
datetime min max

sorenoid:
various fixes to date and time validation

* make a distinction between validating dates and validating date times. This is because we want to allow a date to be pickable when the min or max falls somewhere on that date. This fixes issues with the min date
* convert to UTC before evaluating in the dateTimeValidator
* use the dateTimeValidator to enable the OK button. use the date validator everywhere else


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePicker.kt
```diff
@@ -345,6 +355,32 @@ private object DateTimePickerDialogTokens {
...
+/**
+ * A lenient operator which "ignores" if the receiver is null. This is used
+ * to short circuit a comparison if the receiver is null, meaning the comparison
+ * isn't valid or necessary.
+ *
+ * @param other the right hand side of the comparison
+ * @return -1 if the this receiver is strictly less than other, 1 if the this receiver is strictly greater than other
+ * and 0 if the two values are equal, or the receiver is null.
+ */
+private operator fun Long?.compareTo(other: Long): Int = this?.compareTo(other) ?: 0
```
kaushikrw, line 367: Are these `compareTo` used anywhere?

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePicker.kt
```diff
@@ -146,9 +146,16 @@ internal fun DateTimePicker(
     }
     
     // confirm button is only active when a date has been selected
```
kaushikrw, line 148:
```suggestion
    // confirm button is only active when a date and time have been selected and within given bounds, if provided
```

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePickerState.kt
```diff
@@ -243,12 +250,36 @@ private class DateTimePickerStateImpl(
     }
     
     override fun dateTimeValidator(timeStamp: Long): Boolean {
```
kaushikrw, line 252: I think the use of a `dateTimeValidator` and a `dateValidator` is little bit confusing and could be just one method. The difference I see is in addition to only using date, `dateValidator` is converting the `minDateTime` and `maxDateTime` to UTC but they're already in UTC.

The DatePicker's validation in all places should also consider the time. For ex, a field has min date of `7/1/1968 12:01:01 am PT`. In this case the time is the one being restricted and the date of `7/1/1968` should be valid but with the current `dateValidator` it won't be since it validates with the time of 12:00 am.

kaushikrw (changes requested):

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePicker.kt
```diff
@@ -166,13 +156,26 @@ internal fun DateTimePicker(
...
+            confirmEnabled = datePickerState.selectedDateMillis?.let {
+                state.dateTimeValidator(
+                    it + timePickerState.hour * 3_600_000 + timePickerState.minute * 60_000
+                )
+            } ?: false,
             pickerInput = pickerInput,
             onToday = {
-                state.today()
+                state.setDateTime(
+                    Instant.now().toEpochMilli().toDateMillis(),
```
kaushikrw, line 167: Doesn't this also need to converted into time in local zone similar to L291?

kaushikrw, line 167: Doesn't this also need to converted into time in local zone similar to L291?

Also, I see that the same code exists in `state.now()` and similarly in `state.today()`. Is there a difference? If not, I think we should just keep one.

kaushikrw (approved):
Looks good!

PR #117, opened 2023-09-07
consume core mutable elements

sorenoid:
pick up build 3989 of the Kotlin SDK. This has functioning FeatureForm collections of mutable elements and other enhancements.


sorenoid:
inadvertently merged in v.next which wasn't the point of this PR.

PR #123, opened 2023-09-12
DateTimePicker "now" freeze fix

kaushikrw:
## Summary of changes

The main fix to solve the "now" freezing and not snapping to the current month is to wrap the DatePicker in a `key(state.dateTime) { }`. 

### Findings

There seems to be unwanted behavior when `DatePicker` is recomposed with a new `DatePickerState` which freezes the current month. From my testing and debugging with the Layout Inspector I suspect it's due to a some internal state and remembered values inconsistency.

Wrapping the `DatePicker` with a key changes the scope of all internal remembered states and forces all recalculations when the key changes. This fixes the issue.


sorenoid (approved):
this works well. It also fixes the month paging issue. It makes the DatePicker work as I would expect it to -- updating when we update our state object's value. nice work.

PR #124, opened 2023-09-13
small today fix

sorenoid:
I didn't fully test the today function when fixing min/max validation...it was broken when the initial date value is not null (and also lots of other ways that don't manifest in our UI). I had only tested on a field that is initially empty.

This fix changes the "today" value to add the offset to arrive at a picker value, instead of subtracting it. It also gets the millis of the ensuing date after that addition so that the millis do not get counted when hours and minutes are derived from it, which happens when the initial date is not null.

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePicker.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePicker.kt
@@ -165,10 +165,11 @@ internal fun DateTimePicker(
             onToday = {
                 val now = Instant.now().toEpochMilli().toDateMillis()
                 state.setDateTime(
-                    now.plus(now.defaultTimeZoneOffset),
+                    now.minus(now.defaultTimeZoneOffset),
                     state.dateTime.value.hour,
                     state.dateTime.value.minute
                 )
+                datePickerState.setSelection(state.dateTime.value.dateForPicker)
             },
             onNow = {
                 val now = Instant.now().toEpochMilli().toZonedDateTime()
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePicker.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePicker.kt
@@ -165,9 +165,9 @@ internal fun DateTimePicker(
             onToday = {
                 val now = Instant.now().toEpochMilli().toDateMillis()
                 state.setDateTime(
-                    now.minus(now.defaultTimeZoneOffset),
-                    state.dateTime.value.hour,
-                    state.dateTime.value.minute
+                    now.plus(now.defaultTimeZoneOffset).toDateMillis(),
+                    timePickerState.hour,
+                    timePickerState.minute
                 )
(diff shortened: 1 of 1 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePicker.kt
```diff
@@ -165,10 +165,11 @@ internal fun DateTimePicker(
...
                 val now = Instant.now().toEpochMilli().toDateMillis()
                 state.setDateTime(
-                    now.plus(now.defaultTimeZoneOffset),
-                    state.dateTime.value.hour,
-                    state.dateTime.value.minute
+                    now.plus(now.defaultTimeZoneOffset).toDateMillis(),
+                    timePickerState.hour,
+                    timePickerState.minute
                 )
+                datePickerState.setSelection(state.dateTime.value.dateForPicker)
```
sorenoid, line 172: given the `DateTimePicker "now" freeze fix` I am not sure it will still be necessary to set the selection.

sorenoid: (I tested, it will no longer be necessary).

kaushikrw, line 172:
```suggestion
```
Yep, don't need to use setSelection anymore.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePicker.kt
```diff
@@ -166,9 +166,9 @@ internal fun DateTimePicker(
             onToday = {
                 val now = Instant.now().toEpochMilli().toDateMillis()
                 state.setDateTime(
-                    now.plus(now.defaultTimeZoneOffset),
-                    state.dateTime.value.hour,
-                    state.dateTime.value.minute
+                    now.plus(now.defaultTimeZoneOffset).toDateMillis(),
+                    timePickerState.hour,
+                    timePickerState.minute
```
kaushikrw, line 171: Seems like there is double counting happening here with the call to `toDateMillis()`

```
val now = Instant.now().toEpochMilli().toDateMillis()
state.setDateTime(
         now.plus(now.defaultTimeZoneOffset).toDateMillis(),
         timePickerState.hour,
         timePickerState.minute
)
```
This works.
```
val now = Instant.now().toEpochMilli()
state.setDateTime(
        now.plus(now.defaultTimeZoneOffset).toDateMillis(),
        timePickerState.hour,
        timePickerState.minute
)
```

kaushikrw (changes requested):
Another issue I found, I guess which is not due to this PR but is evident without incorporating the changes in the comment below, is related to how the `UtcDateTime.hour` and `UtcDateTime.minute` are set and used.

They are set using the constructor but are never used directly. Instead `dateForPicker` and `hourForPicker` are used which are calculate based on the `epochMillis` making the `hour` and `minute` properties redundant.

Hence this behavior can be seen:

If `state.setDateTime()` is called with a date which has a different (hour, minute) than the hour and minute parameters, that former takes precedence when the `TimePicker` is re-composed.

We should probably fix this in a separate issue, but wanted to bring it up.

kaushikrw (approved):

PR #125, opened 2023-09-20
required text

sorenoid:
Support for required text in FormTextField.

This PR appends an asterisk to the label, and shows "Required" in red in the helper text after focus is gained and the character count of a required field reaches zero. THe text remains after focus is lost.

This deviates from the spec in one minor way: the required text appears while focused if the char count reaches 0, while it shouldn't be shown until after focus is lost when the char count is 0.

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
@@ -85,6 +85,11 @@ internal interface FormTextFieldState {
     val isEditable: Boolean
     
     /**
+     * State that indicates if the field is required.
+     */
+    val isRequired: Boolean
+    
+    /**
      * Callback to update the current value of the FormTextFieldState to the given [input].
      */
     fun onValueChanged(input: String)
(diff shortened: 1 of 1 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
```diff
@@ -210,10 +222,15 @@ private class FormTextFieldStateImpl(
...
      */
     private fun validateLength() {
-        _hasError.value = (_value.value.length !in minLength..maxLength).also {
-            if (it) {
-                _errorMessage.value = helperText
-            }
+        _hasError.value = if (_value.value.length !in minLength..maxLength) {
+            _errorMessage.value = helperText
+            true
+
```
kaushikrw, line 228:
```suggestion
```

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
```diff
@@ -210,10 +222,15 @@ private class FormTextFieldStateImpl(
      * [hasError] and [errorMessage] if there was an error in validation.
```
kaushikrw, line 222: Might need to update the function doc comment too.

kaushikrw (approved):
LGTM!

sorenoid:
thanks @kaushikrw ! merging.

PR #126, opened 2023-09-20
consume event api

sorenoid:
This PR provides support for events and for requesting expression evaluation in the FeatureForm API.

* expressions are evaluated after each change to the value in an element.
* StateFlow properties from the FeatureFormElement are used by state objects to provide support for elements reacting to changes based on expression evaluation. 
* An initial expression evaluation is requested whenever a new FeatureForm object is passed into the FeatureForm composable function. Until this initial evaluation completes, the form simply displays a progress indicator. I haven't seen this, but it could happen with heavy load of expression evaluation defined in the form.
* currently the `FieldFormElement.value` StateFlow returns a formatted String, which is problematic for the DateTime composable, which needs a Long to pass to the pickers. To keep this formatting logic out of the composable, a coroutine scope is passed into the DateTimeFieldState object so that it can collect on the value flow, and map the value to a Long.

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
@@ -10,6 +10,7 @@ import androidx.compose.foundation.lazy.items
 import androidx.compose.material3.Divider
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Alignment
@@ -35,6 +36,9 @@ public fun FeatureForm(
 ) {
     val featureForm by featureFormState.featureForm.collectAsState()
     featureForm?.let {
+        LaunchedEffect(Unit) {
+            it.evaluateExpressions()
+        }
         FeatureFormContent(form = it, modifier = modifier)
     } ?: run {
         Column(
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
@@ -39,7 +39,7 @@ public fun FeatureForm(
 ) {
     val featureForm by featureFormState.featureForm.collectAsState()
     var initialEvaluation by remember { mutableStateOf(false) }
-    LaunchedEffect(Unit) {
+    LaunchedEffect(featureForm) {
         // ensure expressions are evaluated before state objects are created.
         featureForm?.evaluateExpressions()
         initialEvaluation = true
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
@@ -23,6 +23,7 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import com.arcgismaps.mapping.featureforms.FeatureForm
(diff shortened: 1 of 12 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
```diff
@@ -122,12 +127,13 @@ private class FormTextFieldStateImpl(
     private val featureForm: FeatureForm,
     private val context: Context
 ) : FormTextFieldState {
-    private val _value = mutableStateOf(formElement.value.ifEmpty {
+    private val _value = mutableStateOf(formElement.value.value.ifEmpty {
         // "prime" the value until expressions can be evaluated to populate the value.
         // TODO: remove this when the value is provided by expression evaluation.
         featureForm.getElementValue(formElement)?.toString() ?: ""
```
kaushikrw, line 133: Is this still required?

kaushikrw (changes requested):

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
```diff
@@ -34,14 +41,36 @@ public fun FeatureForm(
     modifier: Modifier = Modifier
 ) {
     val featureForm by featureFormState.featureForm.collectAsState()
+    var initialEvaluation by remember { mutableStateOf(false) }
```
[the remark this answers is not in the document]
kaushikrw: @sorenoid I think this still needs to be keyed on `featureForm` so that its reset to false every time a form is opened.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
```diff
@@ -50,7 +56,9 @@ internal interface FormTextFieldState {
     /**
      * Current value state for the [FormTextField].
      */
-    val value: State<String>
+    val value: StateFlow<String>
+    
+    //val valueChanged: StateFlow<String>
```
kaushikrw, line 61:
```suggestion
```

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextField.kt
```diff
@@ -55,7 +68,7 @@ internal fun FormTextField(
         .fillMaxSize()
         .onFocusChanged { state.onFocusChanged(it.hasFocus) }
         .pointerInput(Unit) {
-            // any tap on a blank space will also dismiss the keyboard and clear focus
+            // any tap on a blank space will also dismiss the keyboard and clear focusS
```
kaushikrw, line 71:
```suggestion
            // any tap on a blank space will also dismiss the keyboard and clear focus
```

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextField.kt
```diff
@@ -41,12 +42,24 @@ internal fun FormTextField(
...
     var clearFocus by remember { mutableStateOf(false) }
+    val isEditable by state.isEditable.collectAsState()
+    val isRequired by state.isRequired.collectAsState()
+    val label by remember(isRequired) {
+        if (isRequired) {
+            mutableStateOf("${state.label} *")
+        } else {
+            mutableStateOf(state.label)
+        }
+    }
```
kaushikrw, line 56:
```suggestion
    val label = remember(isRequired) {
        if (isRequired) {
            "${state.label} *"
        } else {
            state.label
        }
    }
```

I think the label need not be a `State` since it's backed by a `State`

kaushikrw (approved):
This looks good! Just some minor comments below and a note above.

I think it can merged post those changes so approving it to avoid any delays.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextField.kt
sorenoid: > I think the label need not be a `State` since it's backed by a `State`

do you mean that since the key's value is delegated to a state variable, there is no need for `label` to be a state variable?

sorenoid:
thanks for the help @kaushikrw !!

PR #127, opened 2023-09-25
Forms combo box implementation

kaushikrw:
### Summary of changes

- A new state object `BaseFieldState` is introduced which contains all the common properties of a field element. Any composite component which has a state class now inherits from `BaseFieldState`. Furthermore, these composite state classes are now a `plain old kotlin class` instead of an interface resulting in a simplified heirarchy.
- A new `BaseTextField` is introduced that acts as a standard text input field for the feature forms. The common visual design and behavior now reside here, making the UI more uniform across all the different composite fields.
- A new `ComboBoxField` is introduced which uses a `Dialog` to present the list of options from a coded value domain.

#### Important Notes

- There is a slight design deviation in the `ComboBoxField` where the dialog uses a `Done` button to dismiss instead of the `Cancel` as the design document specifies. I think `Done` is more apt, as after making a choice a `Cancel` operation is a bit confusing.
- `Required` or any additional validation for the `ComboBoxField` needs to be implemented pending additional data in the form of webmaps to test with.
- `DateTimeField` is not part of this refactoring and will be taken up separately.

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/combo/ComboBoxField.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/combo/ComboBoxField.kt
@@ -101,7 +101,7 @@ internal fun ComboBoxField(state: ComboBoxFieldState, modifier: Modifier = Modif
     if (showDialog) {
         ComboBoxDialog(
             initialValue = value,
-            values = state.codedValues.map { it.name },
+            values = state.codedValues.map { it.code.toString() },
             label = state.label,
             description = state.description,
             isRequired = isRequired,
@@ -173,7 +173,7 @@ internal fun ComboBoxDialog(
                             },
                             modifier = Modifier.weight(1f),
                             placeholder = {
-                                Text(text = stringResource(R.string.filter, label))
+                                Text(text = "${stringResource(R.string.filter)} $label")
                             },
                             trailingIcon = {
                                 if (searchText.isNotEmpty()) {
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/combo/ComboBoxField.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/combo/ComboBoxField.kt
@@ -18,11 +18,9 @@ import androidx.compose.foundation.lazy.items
 import androidx.compose.foundation.text.KeyboardActions
 import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.material.icons.Icons
-import androidx.compose.material.icons.materialIcon
 import androidx.compose.material.icons.outlined.Check
 import androidx.compose.material.icons.outlined.Close
 import androidx.compose.material.icons.outlined.List
-import androidx.compose.material.icons.outlined.Menu
 import androidx.compose.material3.Divider
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.Icon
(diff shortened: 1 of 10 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/combo/ComboBoxField.kt
```diff
@@ -0,0 +1,266 @@
...
+                        horizontalArrangement = Arrangement.SpaceBetween
+                    ) {
+                        TextField(
+                            value = searchText,
+                            onValueChange = {
+                                searchText = it
+                            },
+                            modifier = Modifier.weight(1f),
+                            placeholder = {
+                                Text(text = "Filter $label")
```
sorenoid, line 173: this string should be in resources as a formatted string.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/combo/ComboBoxField.kt
```diff
@@ -0,0 +1,266 @@
...
+                                imeAction = ImeAction.Done
+                            ),
+                            shape = RoundedCornerShape(15.dp),
+                            colors = TextFieldDefaults.colors(
+                                focusedIndicatorColor = Color.Transparent,
+                                unfocusedIndicatorColor = Color.Transparent
+                            ),
+                        )
+                        TextButton(onClick = onDismissRequest) {
+                            Text(text = "Done")
```
sorenoid, line 196: needs resource string.

sorenoid (changes requested):
Looks good. I know this will get updated to support events. There are a couple of strings that need to be moved into resources. Other than that very minor detail this is ready for approval, but I'll wait until it is updated.

kaushikrw:
Thanks @sorenoid, ready for another look!

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/combo/ComboBoxField.kt
```diff
@@ -0,0 +1,268 @@
...
+                        horizontalArrangement = Arrangement.SpaceBetween
+                    ) {
+                        TextField(
+                            value = searchText,
+                            onValueChange = {
+                                searchText = it
+                            },
+                            modifier = Modifier.weight(1f),
+                            placeholder = {
+                                Text(text = "${stringResource(R.string.filter)} $label")
```
sorenoid, line 175: the resource string itself ought to do the formatting, so that when it is translated the translator can change the placement of the `label` string within the resource string.

```suggestion
stringResource(R.string.filter, label)
```

toolkit/featureforms/src/main/res/values/strings.xml
```diff
@@ -20,10 +20,13 @@
...
+    <string name="no_value">No value</string>
+    <string name="enter_value">Enter Value</string>
     <string name="required">Required</string>
     <string name="today">Today</string>
     <string name="now">Now</string>
     <string name="cancel">Cancel</string>
-    <string name="ok">Ok</string>
+    <string name="ok">OK</string>
+    <string name="done">Done</string>
+    <string name="filter">Filter</string>
```
sorenoid, line 31: as above, to allow the translator to structure the string differently.

```suggestion
    <string name="filter">Filter %1$s</string>
```

sorenoid (changes requested):
Looks good, two quick things to address while I run it and test it. I probably misled you with my previous review..sorry

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/combo/ComboBoxField.kt
kaushikrw: Ah missed that, thanks.

sorenoid (approved):
Very nice work!

kaushikrw:
Thanks @sorenoid, merging this!

PR #128, opened 2023-09-26
GeoCompose module 

shubham7109:
PR to allow for any feedback and comments for GeoCompose. 

Reference issue: `runtime/kotlin/issues/3056`


microapps/MapComposableApp/app/src/test/java/com/arcgismaps/toolkit/templateapp/ExampleUnitTest.kt
```diff
@@ -0,0 +1,34 @@
+/*
```
gunt0001, line 1: I think we can remove this file.

settings.gradle.kts
```diff
@@ -85,3 +87,5 @@ project(":bom").projectDir = File(rootDir, "bom")
 project(":composable-map").projectDir = File(rootDir, "toolkit/composable-map")
 project(":indoors").projectDir = File(rootDir, "toolkit/indoors")
 project(":floor-filter-app").projectDir = File(rootDir, "microapps/FloorFilterApp/app")
+project(":geo-composables").projectDir = File(rootDir, "toolkit/geo-composables")
```
gunt0001, line 90: With regards to the module name "geo-composables" - I find the term "geo" too broad. One could argue that components such as FloorFilter or Compass are also "geo composables". How about "map-composables" with "map" also covering "scene", given a scene is basically a map as well.

Thoughts @shubham7109 @puneet-pdx @rchintapalli @sorenoid @kaushikrw ?

toolkit/geo-composables/src/main/java/com/arcgismaps/toolkit/geocomposables/SceneComposable.kt
```diff
@@ -0,0 +1,33 @@
...
+ *
+ */
+
+package com.arcgismaps.toolkit.geocomposables
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.tooling.preview.Preview
+
+@Composable
+public fun SceneComposable() {
```
gunt0001, line 25: As above, we could name this: 
```suggestion
public fun Scene() {
```

toolkit/geo-composables/src/main/java/com/arcgismaps/toolkit/geocomposables/MapComposable.kt
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package com.arcgismaps.toolkit.geocomposables
+
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.tooling.preview.Preview
+
+@Composable
+public fun MapComposable() {
```
gunt0001, line 26: @sorenoid brought up the idea of calling this composable simply `Map()` (and then the 3D equivalent would be `Scene()`). As opposed to `ArcGISMap`, which would have been conflicting with Kotlin's [Map interface](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-map/) if we wouldn't have prefixed it with "ArcGIS", we don't have this issue here since we define a function, not a class.  

Thoughts @shubham7109 @puneet-pdx @rchintapalli @sorenoid @kaushikrw ?

```suggestion
public fun Map() {
```

toolkit/geo-composables/src/test/java/com/arcgismaps/toolkit/geocomposables/ExampleUnitTest.kt
```diff
@@ -0,0 +1,34 @@
+/*
+ *
```
gunt0001, line 2: Can remove this file.

gunt0001 (commented):
@shubham7109 - looks good, see my comments on naming.

puneet-pdx (approved):

gunt0001 (approved):
👍

PR #130, opened 2023-09-28
AGP, gradle upgrade to 8.1.1, 8.1

sorenoid:
this upgrade matches the API's versions. These versions are necessary to include the SDK as a composite build. Also clears some false warnings in gradle files.

Android Studio giraffe is required to use AGP 8.1.1


shubham7109 (approved):
Looks good 👍

kaushikrw (approved):

PR #131, opened 2023-10-02
Form field states remember saveable

kaushikrw:
### Summary of changes

- Added custom Saver objects to implement `rememberSaveable()` for FormTextFieldState and `ComboBoxFieldState` by providing a `rememberFormTextFieldState`, `rememberComboBoxFieldState`.
- The field states must be saved outside the `LazyColumn` since the current `LazyColumn` implementation does not save any states within it that are not visible when a configuration change occurs.
- Additionally, wrapped the appropriate properties for a `BaseFieldState` into a `FieldProperties` class which essentially provides the properties of `FieldFormElement`. This is necessary because `BaseFieldState` needs be constructed without a dependency on `FormFieldElement` or `FeatureForm` for both `rememberSaveable()` and Compose preview purposes. It also helps to be passed as a single constructor parameter as opposed to having a long list of parameters. 
- Similarly, `FormTextFieldState` uses `TextFieldProperties` which inherits from `FieldProperties` while adding its own properties. `ComboBoxFieldState` uses `ComboBoxFieldProperties` in the same way.

#### Note
This PR does not address the `DateTimeFieldState` since it does not inherit from `BaseFieldState`.

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
@@ -39,6 +39,7 @@ import com.arcgismaps.toolkit.featureforms.components.base.BaseFieldState
 import com.arcgismaps.toolkit.featureforms.components.combo.rememberComboBoxFieldState
 import com.arcgismaps.toolkit.featureforms.components.text.rememberFormTextFieldState
 import kotlinx.coroutines.CoroutineScope
+import java.util.Objects
 
 /**
  * A composable Form toolkit component that enables users to edit field values of features in a
@@ -141,7 +142,7 @@ private fun rememberFieldStates(
     form: FeatureForm,
     context: Context,
     scope: CoroutineScope
-): Map<String, BaseFieldState?> {
+): Map<Int, BaseFieldState?> {
     return form.elements.filterIsInstance<FieldFormElement>().associateBy(
         { fieldElement ->
             fieldElement.id
@@ -201,7 +202,7 @@ private fun NoDataPreview() {
 /**
  * Unique id for each form element.
  */
-internal val FieldFormElement.id: String
+internal val FieldFormElement.id: Int
     get() {
-        return fieldName + label + description + hint
+        return Objects.hash(fieldName, label, description, hint)
     }
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
@@ -1,5 +1,7 @@
 package com.arcgismaps.toolkit.featureforms
 
+import android.content.Context
+import android.util.Log
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Spacer
@@ -9,6 +11,8 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
+import androidx.compose.foundation.lazy.itemsIndexed
+import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.Divider
 import androidx.compose.material3.Text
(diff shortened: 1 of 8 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/combo/ComboBoxFieldState.kt
```diff
@@ -25,44 +29,137 @@ import com.arcgismaps.mapping.featureforms.FormInputNoValueOption
...
+                        value = formElement.value,
+                        editable = formElement.isEditable,
+                        required = formElement.isRequired,
+                        codedValues = input.codedValues,
+                        showNoValueOption = input.noValueOption,
+                        noValueLabel = input.noValueLabel
+                    ),
+                    initialValue = list[0],
+                    context = context,
+                    scope = scope,
```
sorenoid, line 125: Wouldn't this CoroutineScope be cancelled here? It is passed in at object creation time, and then used at object re-creation time, but it most likely comes from a `rememberCoroutineScope()` call?

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/combo/ComboBoxFieldState.kt
```diff
@@ -25,44 +29,137 @@ import com.arcgismaps.mapping.featureforms.FormInputNoValueOption
...
  *
- * @param formElement The [FieldFormElement] to create the state from.
- * @param featureForm The [FeatureForm] that the [formElement] is a part of.
+ * @param properties the [ComboBoxFieldProperties] associated with this state.
+ * @param initialValue optional initial value to set for this field. It is set to the value of
+ * [TextFieldProperties.value] by default.
+ * @param scope a [CoroutineScope] to start [StateFlow] collectors on.
  * @param context a Context scoped to the lifetime of a call to the [FieldElement] composable function.
+ * @param onEditValue a callback to invoke when the user edits result in a change of value. This
+ * is called on [ComboBoxFieldState.onValueChanged].
```
sorenoid, line 61: This class can be marked as `@Stable` I think.

sorenoid (changes requested):
Looks good. One question about the lifetime of the CoroutineScope used to restore state objects...

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
```diff
@@ -127,3 +197,11 @@ private fun InitializingExpressionsPreview() {
...
     NoDataToDisplay()
 }
+
+/**
+ * Unique id for each form element.
+ */
+internal val FieldFormElement.id: String
+    get() {
+        return fieldName + label + description + hint
+    }
```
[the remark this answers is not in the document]
kaushikrw: For reasons I don't quite understand, I am seeing different values for `hashCode` when called in `rememberFieldStates()` compared to calling it in the `LazyColumn`. 

Is it safe to assume `fieldName + label + description + hint` is unique for every field?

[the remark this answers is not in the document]
sorenoid: really? Maybe we can provide hashCode and equals in a snippet in the SDK, or introduce it in the YAML API design. For this, you might want to try

```java
Objects.hash(fieldName, label, description, hint)
```

just in case one FieldFormElement fieldName is `Foo`, and label `Bar` and some some other FieldFormElement has fieldName `FooBar`, etc.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/combo/ComboBoxFieldState.kt
kaushikrw: I believe the `Saver( .., scope: CoroutineScope, ..)` is called each time a configuration change occurs, hence the `restore { .. }` block should always recompose with the `CoroutineScope` that is tied to the current composition.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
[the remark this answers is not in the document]
kaushikrw: I think that would be a good idea if we can get that unique identifier for each element from the api.

Thank you for the suggestion, updated!

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/combo/ComboBoxFieldState.kt
sorenoid: huh. that is not how I thought it worked.

sorenoid (approved):
very nice!

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/combo/ComboBoxFieldState.kt
kaushikrw: Testing for expression evaluation after configuration changes would be a good way to validate this and it works as expected.

kaushikrw:
Thanks @sorenoid, merging this!

PR #135, opened 2023-10-04
Forms: Microapp bottomsheet-FeatureForm lifecycle fix

kaushikrw:
### Summary of changes

- The `BottomSheetScaffold` is causing too many recompositions and keeps the `FeatureForm` alive in the compose hierarchy even though it is not on screen when dismissed. This is not the intended behavior. The fix is to scrap the `BottomSheetScaffold` and simply use a `StandardBottomSheet` with the `FeatureForm` as its content.
- The `StandardBottomSheet` is only drawn when the editing mode is set and hence a cleaner lifecycle for the `FeatureForm`.
- `ClearFocus` is also not needed anymore since the `FeatureForm` leaves the composition when dismissed along with the BottomSheet.


sorenoid (approved):
Looks good!

PR #138, opened 2023-10-05
Sor10874/numeric text field

sorenoid:
Provides support for numeric types in FormTextField. 

* adds a FieldType and a Domain? property to the BaseFieldState, and the FieldProperties. These are common to all elements, but only used in FormTextField and BaseTextField.
* adds Saver supporting logic for the new properties
* invokes the appropriate keyboard for integer, floating point, or non-numeric input on BaseTextField
* adds some validation strings for numeric inputs, and refactors validation a bit
* extensions on FieldType and platform types to make things a little simpler
* tests
* test refactoring to support FieldProperties (tests were broken).

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
@@ -30,6 +30,7 @@ import com.arcgismaps.data.FieldType
 import com.arcgismaps.data.RangeDomain
 import com.arcgismaps.mapping.featureforms.FeatureForm
 import com.arcgismaps.mapping.featureforms.FieldFormElement
+import com.arcgismaps.mapping.featureforms.TextAreaFormInput
 import com.arcgismaps.mapping.featureforms.TextBoxFormInput
 import com.arcgismaps.toolkit.featureforms.R
 import com.arcgismaps.toolkit.featureforms.components.FieldElement
@@ -108,7 +109,7 @@ internal class FormTextFieldState(
     // fetch the maxLength based on the featureFormElement.inputType
     val maxLength = properties.maxLength
     
-    private var wasFocused: Boolean = false
+    private var hasBeenFocused: Boolean = false
     
     // supporting text will depend on multiple other states. If there is an error, it will display
     // error message. Otherwise description is displayed, unless it is empty in which case
@@ -121,7 +122,6 @@ internal class FormTextFieldState(
 
     private var validationErrors: MutableList<ValidationErrorState> = mutableListOf()
     private val _hasError = mutableStateOf(false)
-    val hasError: State<Boolean> = _hasError
     private val _supportingTextIsErrorMessage = mutableStateOf(false)
     val supportingTextIsErrorMessage: State<Boolean> = _supportingTextIsErrorMessage
     
@@ -161,8 +161,6 @@ internal class FormTextFieldState(
         )
     }
     
-    private var _errorMessage: String = ""
-
     // build helper text
     private val helperText =
         if (fieldType.isNumeric) {
@@ -212,7 +210,7 @@ internal class FormTextFieldState(
         scope.launch {
             isFocused.drop(1).collect {
                 if (it) {
-                    wasFocused = true
+                    hasBeenFocused = true
                 }
                 updateValidation(value.value)
             }
@@ -314,7 +312,7 @@ internal class FormTextFieldState(
             } else {
                 validationErrors.first()
             }
-        } else if (wasFocused) {
+        } else if (hasBeenFocused) {
             validationErrors.firstOrNull { it == Required} ?: validationErrors.first()
         } else {
             // never been focused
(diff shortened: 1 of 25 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
```diff
@@ -102,52 +114,135 @@ internal class FormTextFieldState(
...
             }
         }
         scope.launch {
             isFocused.collect {
                 if (it) {
-                    validate(value.value)
+                    validate(value.value, true)
+                } else {
+                    validate(value.value, !isRequired.value)
+                }
```
kaushikrw, line 179:
```suggestion
                    validate(value.value, !isRequired.value)
                }
```
Is there a reason for this change? This would preemptively show all validation errors even for fields that cannot be edited.

kaushikrw (changes requested):
This looks good. The numeric entry is working really well. I had some comments regarding the 
 text validation.

Also, I was unable to get the `FormTextFieldTests` and the `FormTextFieldNumericTests` to run successfully. The former were failing with `Failed to assert the following: (Focused = 'true')` while the latter was failing with `expected:<[Enter value from 1 to 7]> but was:<[This Field Has a Range domain 1-7]>. `
Are you seeing the same behavior?

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
```diff
@@ -102,52 +114,135 @@ internal class FormTextFieldState(
...
-        // TODO: when consuming the core API throw here and remove the line above.
-        //throw IllegalStateException("invalid form data or attribute: text field must have a nonzero max length")
-    }
 
     init {
         scope.launch {
             value.drop(1).collect { newValue ->
                 if (isEditable.value) {
-                    validate(newValue)
+                    validate(newValue, true)
```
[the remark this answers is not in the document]
sorenoid: the `canBeEmpty` parameter is meant to prevent the "Required" text from showing when editing a required field. We do not run this `validate` unless the field is editable. If the user is editing, the field has focus. The design spec specifically shows that a required, focused, empty field should show the description rather than the "Required" error message.
Screen Shot 2023-10-11 at 10 24 22 AM



There is a corner case where a field becomes editable as the result of expression evaluation and should therefore be validated immediately. In that case we maybe ought to maintain a property like `wasEditable` and collect on `isEditable` and run validation, similar to what we do for focus changes. But I'd like to make an issue to consider that separately in case it isn't as straight forward as it seems, if that is ok?

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
sorenoid: fields that cannot be edited are disabled and cannot be focused. Their focus state will never change. If the field is focused, it should not show the "Required" error string according to the design spec.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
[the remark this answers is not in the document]
kaushikrw: I think that makes sense. 

There is another case where this would fail, when the value is cleared using the `clear` action button and the field is not focused, it does not trigger required validation.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
sorenoid: I have added a change to drop the first value collected from the isFocused StateFlow, so validation doesn't occur preemptively.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
```diff
@@ -85,84 +107,257 @@ internal class FormTextFieldState(
...
+            }
+        } else {
+            NoError
+        }
+    
+    private fun validateNumericRange(value: String): ValidationErrorState =
+        if (domain != null && domain is RangeDomain) {
+            if (fieldType.isIntegerType) {
+                val (min, max) = domain.asLongTuple
+                val numberVal = value.toLong()
```
kaushikrw, line 255:
```suggestion
                val numberVal = value.toLongOrNull()
```
This could throw an exception if the input is not a number.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
```diff
@@ -85,84 +107,257 @@ internal class FormTextFieldState(
...
+                        NoError
+                    } else {
+                        MaxNumericConstraint
+                    }
+                } else {
+                   NoError
+                }
+            } else {
+                val (min, max) = domain.asDoubleTuple
+                val numberVal = value.toDouble()
```
kaushikrw, line 280:
```suggestion
                val numberVal = value.toDoubleOrNull()
```
Similarly, here too.

kaushikrw (changes requested):
This looks good. Some small changes might be needed though!

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/text/FormTextFieldState.kt
```diff
@@ -85,84 +107,257 @@ internal class FormTextFieldState(
...
-    val supportingText = derivedStateOf {
-        if (_hasError.value) _errorMessage else {
-            description.ifEmpty {
-                if (_isFocused.value) helperText else ""
-            }
-        }
-    }
-
+    private val _supportingText: MutableState<String> = mutableStateOf(description)
+    val supportingText = _supportingText
```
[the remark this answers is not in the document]
sorenoid: I agree that the type should be specified, but I don't think this needs to be a computed property does it?

kaushikrw (approved):
LGTM!

sorenoid:
thanks @kaushikrw merging!

PR #140, opened 2023-10-07
filter out layers without forms in results of identify

sorenoid:


microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapViewModel.kt
```diff
@@ -48,29 +48,34 @@ class MapViewModel @Inject constructor(
...
+                        feature.load().onSuccess {
+                            try {
+                                (feature.featureTable?.layer as FeatureLayer).featureFormDefinition?.let { featureFormDefinition ->
+                                    val featureForm = FeatureForm(feature, featureFormDefinition)
+                                    // update the FeatureFormState's FeatureForm
+                                    setFeatureForm(featureForm)
+                                    // set the FeatureFormState to an editing state to bring up the
+                                    // FeatureForm UI
+                                    setTransactionState(EditingTransactionState.Editing)
+                                } ?: throw Exception("This feature does not have a FeatureFormDefinition")
```
kaushikrw, line 67:
```suggestion
                         val featureForm = FeatureForm(
                                feature,
                                (feature.featureTable?.layer as FeatureLayer).featureFormDefinition!!
                            )
                            // update the FeatureFormState's FeatureForm
                            setFeatureForm(featureForm)
                            // set the FeatureFormState to an editing state to bring up the
                            // FeatureForm UI
                            setTransactionState(EditingTransactionState.Editing)
```

Nitpick but this null check can be a `not-null assertion ` since we already know that from the `mapNotNull` transform.

kaushikrw (approved):
A small nitpick but good to merge after that.

PR #141, opened 2023-10-10
Forms: Add ComboBox tests

kaushikrw:
### Summary of changes

- Added tests for combo box based on the common test design


toolkit/featureforms/src/androidTest/java/com/arcgismaps/toolkit/featureforms/ComboBoxFieldTests.kt
```diff
@@ -0,0 +1,444 @@
...
+        val optionsIconNode =
+            comboBoxField.assertContentDescriptionContains(optionsIconSemanticLabel)
+        optionsIconNode.assertIsDisplayed()
+    }
+
+    /**
+     * Test case 3.3:
+     * Given a ComboBoxField with a pre-existing value, description and a no value label
+     * When the ComboBoxField is tapped
+     * Then the ComboBoxDialog is shown with all coded values AND
```
sorenoid, line 177: The test design doesn't specify that the dialog should show all coded values, and this test doesn't assert it either. I think this part of the comment can be removed. 

It seems like a good subject for a runtimecore unit test to test that all coded values are in the list, but not something that we need to assert in the toolkit.
```suggestion
     * Then a ComboBoxDialog is shown with AND
```

sorenoid (changes requested):
very nice. just a couple nitpicks.

kaushikrw:
@sorenoid resolved conflicts, ready for another look!

sorenoid (approved):

PR #142, opened 2023-10-10
Compose viewpoint operations

shubham7109:
### Description
PR to add viewpoint operations to the Composable Map toolkit project. This includes setting the viewpoint (animated/non-animated) and get viewpoint functions. 

### Implementation
I followed the API implementation pattern to define the viewpoint functions in the StateObjects. Where some functions are defined as part of `GeoComposeState` and some are `MapState`. The `GeoComposeState` holds the internal instance of the `viewpointChannel` which is also inherited in `MapState`, so the same channel is used across both state objects. 
  
Non-animated functions are managed by internal StateFlows in `GeoComposeState`: 
- getCurrentViewpoint()
- setViewpoint(...)

Animated functions are part of the `ViewpointOperation` class with internal subclasses used the in following state objects: 
- GeoComposeState
   - ViewpointAnimated
   - ~~ViewpointAnimatedWithDuration~~
- MapState
  - ViewpointAnimated
  - ~~ViewpointAnimatedWithDurationAndCurve~~
  - ViewpointCenter
  - ~~ViewpointCenterAndScale~~
  - ViewpointGeometry
  - ~~ViewpointGeometryAndPadding~~
  - ViewpointRotation
  - ViewpointScale

### Questions 🤔 

- [x]  Would the `UiThread` annotation be needed for the suspend function calls? The API doesn't seem to be using it, so I haven't used it in the PR
- [x]  Should `ViewpointOperation` be declared in a separate package? I assume there will be multiple "operation" files to provide a channel flow to the state objects
- [x]  Thoughts on `setViewpoint` and `getCurrentViewpoint`? As we set the properties as mutable state flows and need to account for changes using collect. I encountered a case where getCurrentViewpoint would notify the Map to set the mutable state flow with the current viewpoint, but something like `setViewpointScale` would not work as it is also needs to modify the same state flow. To avoid this, I treat both these functions to manage separate internal viewpoint state flows `currentViewpoint`(set by Map on viewpointChanged) or `viewpoint`(set by user on setViewpoint call) 

### Testing 
Tested the implementation through the micro app. This code snippet could help:
```kotlin
@Composable
fun MainScreen() {
    // set up a viewpoint
    val londonPoint = Point(-14093.0, 6711377.0, SpatialReference.webMercator())
    val londonViewpoint = Viewpoint(londonPoint, 10000000.0)

    // set the ArcGISMap to display the London viewpoint
    val arcGISMap = ArcGISMap(BasemapStyle.ArcGISStreets).apply {
        initialViewpoint = londonViewpoint
    }

    // create the MapState
    val mapState = remember { MapState(arcGISMap = arcGISMap) }

    // add the Compose Map using the MapState to show map with London viewpoint
    Map(
        mapState = mapState,
    )


    // Use this to test instantaneous setViewpoint call
    LaunchedEffect(mapState) {
        // non animated set viewpoint
        // mapState.setViewpoint(londonViewpoint)
    }


    val scope = rememberCoroutineScope()
    val newScale = 1000000.0
    // Use this to test asynchronous animated set Viewpoint calls:
    Button(onClick = {
        scope.launch {
            // animated set viewpoint functions
            mapState.setViewpointCenter(londonPoint, newScale).onSuccess {
                // Get the current viewpoint after center animation
                val viewpoint = mapState.getCurrentViewpoint(ViewpointType.CenterAndScale)
            }
        }
    }) {
        Text(text = "Set new viewpoint")
    }
}
```


gunt0001:
> Would the UiThread annotation be needed for the suspend function calls? The API doesn't seem to be using it, so I haven't used it in the PR

I don't think it's needed. Since we send all operations onto a channel, users should be able to make those calls concurrently, from any dispatcher they like.

> Should ViewpointOperation be declared in a separate package? I assume there will be multiple "operation" files to provide a channel flow to the state objects

I think it's fine to keep them in the same package, as long as they are defined as `internal`.

> Thoughts on setViewpoint and getCurrentViewpoint?

See our separate discussion on that...

shubham7109:
Thanks for the feedback @gunt0001. 
Changes made:
- Updated to use function overloading to combine, `setViewpointAnimated`, `setViewpointCenter`, and `setViewpointGeometry`
- Removed the previous implementation for the current/set viewpoint, and added `SetViewpointOperation` and `GetCurrentViewpointOperation` to handle operations using channel flows. 
- Updated Testing section in this PR with the code to set `initialViewpoint` before the map is rendered.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/Map.kt
```diff
@@ -52,6 +53,94 @@ public fun Map(modifier: Modifier = Modifier, mapState: MapState = MapState()) {
...
+        // Collect get current viewpoint operation and pass the current viewpoint back to the channel
+        launch {
+            mapState.getCurrentViewpointChannel.receiveAsFlow().collect {
+                when (it) {
+                    is GetCurrentViewpointOperation.GetCurrentViewpoint -> {
+                        val currentViewpoint = mapView.getCurrentViewpoint(it.viewpointType)
+                        if (currentViewpoint != null) {
+                            it.completeWith(Result.success(currentViewpoint))
+                        } else {
+                            it.completeWith(Result.failure(Throwable("Error retrieving the current viewpoint")))
```
gunt0001, line 77: I think returning a failure result in case of a null viewpoint makes sense, users can then call `result.getOrNull()`. I just think using a more specific exception type would make sense:
```suggestion
                            it.completeWith(Result.failure(IllegalStateException("Unable to retrieve the current viewpoint")))
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/GeoComposeState.kt
```diff
@@ -17,11 +17,69 @@
...
+            getCurrentViewpointChannel.send(it)
+            it.await()
+        }
+
+    /**
+     * Change the Map to the new [viewpoint]. The viewpoint is updated instantaneously.
+     *
+     * @since 200.3.0
+     */
+    public suspend fun setViewpoint(viewpoint: Viewpoint) {
```
gunt0001, line 80: Just realising that all our suspend functions need to return `Result<>` in order to propagate any errors. That means you need to change the function return type and wait on the operation, in case an error occurs while executing the function on the GeoView:
```suggestion
    public suspend fun setViewpoint(viewpoint: Viewpoint): Result<Unit> {
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/GeoComposeState.kt
```diff
@@ -17,11 +17,69 @@
...
+        SetViewpointOperation.SetViewpoint(viewpoint).let {
+            setViewpointChannel.send(it)
```
gunt0001, line 82:
```suggestion
            setViewpointChannel.send(it)
            it.await()
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/GeoComposeState.kt
```diff
@@ -17,11 +17,69 @@
...
+     * @since 200.3.0
+     */
+    public suspend fun setViewpointAnimated(
+        viewpoint: Viewpoint,
+        durationSeconds: Float
+    ): Result<Boolean> =
+        ViewpointOperation.ViewpointAnimated(viewpoint, durationSeconds).let {
+            viewpointChannel.send(it)
+            it.await()
+        }
```
gunt0001, line 62: This function still needs to be combined with the one above.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapState.kt
```diff
@@ -31,11 +35,96 @@ public class MapState(arcGISMap: ArcGISMap? = null) : GeoComposeState() {
...
+    /**
+     * Change the Map to the [center] point and [scale] asynchronously.
+     *
+     * @since 200.3.0
+     */
+    public suspend fun setViewpointCenter(center: Point, scale: Double): Result<Boolean> =
+        ViewpointOperation.ViewpointCenter(center, scale).let {
+            viewpointChannel.send(it)
+            it.await()
+        }
```
gunt0001, line 82: This function still needs to be combined with the one above.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapState.kt
```diff
@@ -31,11 +35,96 @@ public class MapState(arcGISMap: ArcGISMap? = null) : GeoComposeState() {
...
+     * @since 200.3.0
+     */
+    public suspend fun setViewpointGeometry(
+        boundingGeometry: Geometry,
+        paddingInDips: Double
+    ): Result<Boolean> =
+        ViewpointOperation.ViewpointGeometry(boundingGeometry, paddingInDips).let {
+            viewpointChannel.send(it)
+            it.await()
+        }
```
gunt0001, line 107: This function still needs to be combined with the one above.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/SetViewpointOperation.kt
```diff
@@ -0,0 +1,13 @@
+package com.arcgismaps.toolkit.geocompose
```
gunt0001, line 1: missing copyright

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/GetCurrentViewpointOperation.kt
```diff
@@ -0,0 +1,23 @@
+package com.arcgismaps.toolkit.geocompose
```
gunt0001, line 1: missing copyright

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/GeoComposeState.kt
shubham7109: Why should we combine the functions in the state object? I thought we were combining the `ViewpointOperation` nested classes to provide optional parameters.  However, the state object would need to define multiple set viewpoint suspend functions to match the API pattern.

shubham7109: Resolved 👍

shubham7109:
Thanks for the feedback @gunt0001. I've added in the changes and combined the state object's suspend functions to allow for optional params. Ready for review 👍

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapState.kt
```diff
@@ -31,11 +35,83 @@ public class MapState(arcGISMap: ArcGISMap? = null) : GeoComposeState() {
...
+        if (paddingInDips != null) ViewpointOperation.ViewpointGeometry(
+            boundingGeometry,
+            paddingInDips
+        ).let {
+            viewpointChannel.send(it)
+            it.await()
+        } else ViewpointOperation.ViewpointGeometry(boundingGeometry).let {
+            viewpointChannel.send(it)
+            it.await()
+        }
```
gunt0001, line 94: Can be simplified:
```suggestion
    public suspend fun setViewpointGeometry(
        boundingGeometry: Geometry,
        paddingInDips: Double? = null
    ): Result<Boolean> =
        ViewpointOperation.ViewpointGeometry(
            boundingGeometry,
            paddingInDips
        ).let {
            viewpointChannel.send(it)
            it.await()
        }
```

gunt0001 (approved):
@shubham7109 - a few final comments, otherwise looks good.

puneet-pdx (commented):
@shubham7109 looks good. Some doc related comments

PR #145, opened 2023-10-12
add viewpointChanged event

puneet-pdx:


shubham7109 (approved):
Tested with the viewpoint operations branch and works well!

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/GeoComposeState.kt
```diff
@@ -17,11 +17,42 @@
...
+     * @since 200.3.0
+     */
+    private val _viewpointChanged: MutableSharedFlow<Unit> =
+        MutableSharedFlow(
+            extraBufferCapacity = Int.MAX_VALUE,
+            onBufferOverflow = BufferOverflow.DROP_OLDEST
+        )
+
+    /**
+     * A [SharedFlow] which notifies when the viewpoint of GeoCompose has changed.
```
gunt0001, line 44: I think we might be better off to use [GeoView] instead of `GeoCompose`, which is not a real thing:
```suggestion
     * A [SharedFlow] which notifies when the viewpoint of the corresponding [GeoView] has changed.
```
Note sure, worth to discuss..

gunt0001 (approved):
@puneet-pdx - looks good. One comment re terminology in the doc.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/GeoComposeState.kt
puneet-pdx: Would it be better if we say `GeoComposeState` instead ?
I was thinking we don't want to refer to the underlying MapView/GeoView and restrict the conversation to the compose `Map`. So here I was wording it to state ... the viewpoint (property) of GeoComposeState interface has changed, but mistyped it as GeoCompose.

PR #146, opened 2023-10-12
Forms : Refactor `ComboBoxFieldState` into `CodedValueFieldState`

kaushikrw:
### Summary of changes

- Renamed `ComboBoxFieldState` to `CodedValueFieldState`.
- `CodedValueFieldState` class is now `open`.
- Renamed `ComboBoxFieldProperties` to `CodedValueFieldProperties`.
- `CodedValueFieldProperties` is now `open`.
- Renamed `combo` package to `codedvalue`.
- Moved placeholder calculation from `CodedValueFieldState` and into the `ComboBoxField`. This removes the `Context` dependency for the state object and a possible bug when the `isRequired` flow changes.
- Changed `List<CodedValue>` to a `List<String>` in `CodedValueFieldProperties`, since this mapping happens at the picker level anyway.
- Updated the `ComboBoxDialog` to use a `Map<Any?,String>` which represents `<CodedValue.code, CodedValue.name>` and to display the `name` instead of the `code`. `ComboBoxDialog.onValueChange` is called with the `code`.
- Added compose preview for a `ComboBox`.


sorenoid (approved):
The ComboBox's BaseTextField is still showing the code rather than the name. It is ok with me if you want to address that as part of Radio or a separate PR/issue. This PR was meant to be small and quick!

kaushikrw:
Thanks @sorenoid, will make a separate PR for that. Merging this.

PR #149, opened 2023-10-13
Forms : introduce `RadioButtonInput`

kaushikrw:
### Summary of changes

- Added new `RadioButtonField` and `RadioButtonFieldState` which derives from `CodedValueFieldState` with a custom saver.
- `RadioButtonFieldState.shouldFallback` indicates if the field should fallback to a `ComboBoxField`.
- Added custom typography and themed color properties using `RadioButtonFieldDefaults`.
- To support previews, the `RadioButtonField` has an overload that does not rely on non-constructible classes.

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/FormElements.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/FormElements.kt
@@ -40,15 +40,10 @@ internal fun FieldElement(field: FieldFormElement, state: BaseFieldState) {
             }
 
             is RadioButtonsFormInput -> {
-                val value by state.value.collectAsState()
-                val codedValues = (field.input as RadioButtonsFormInput).codedValues.map {
-                    it.code.toString()
-                }
-                val fallback = (value !in codedValues) && (field.input as RadioButtonsFormInput).noValueOption == FormInputNoValueOption.Show
-                if (fallback) {
+                if ((state as RadioButtonFieldState).shouldFallback()) {
                     ComboBoxField(state = state as CodedValueFieldState)
                 } else {
-                    RadioButtonField(state = state as RadioButtonFieldState)
+                    RadioButtonField(state = state)
                 }
             }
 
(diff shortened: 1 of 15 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/FormElements.kt
```diff
@@ -35,6 +39,14 @@ internal fun FieldElement(field: FieldFormElement, state: BaseFieldState) {
                 ComboBoxField(state = state as CodedValueFieldState)
             }
 
+            is RadioButtonsFormInput -> {
+                if ((state as RadioButtonFieldState).shouldFallback()) {
+                    ComboBoxField(state = state as CodedValueFieldState)
```
sorenoid, line 44:
```suggestion
                    ComboBoxField(state = state)
```

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/codedvalue/RadioButtonFieldState.kt
```diff
@@ -0,0 +1,126 @@
...
+    initialValue = initialValue,
+    scope = scope,
+    onEditValue = onEditValue
+) {
+
+    /**
+     * Returns true if the current value of [value] is not in the [codedValues]. This should
+     * trigger a fallback to a ComboBox. If the [value] is empty then this returns false.
+     */
+    fun shouldFallback(): Boolean {
```
sorenoid, line 48: My understanding was that there is no need for the radio to fall back to combo box -- just to show a radio with no selection and/or a no value option?

sorenoid (approved):
The fallback logic is kind of up in the air, but my understanding was that there was no case in which the radio needed to fall back to a combo box. I think it is ok to proceed and treat this corner case which isn't nailed down as an issue, should we gain clarity.

A couple of suggestions, but this looks good. the defaults class and the use of the composition local is very nice.

kaushikrw:
@sorenoid Thanks, that makes sense. We can update this if that design changes. Merging!

PR #152, opened 2023-10-17
Sor10874/disappearing edits

sorenoid:
For BaseFieldState `value` Flow, this fixes an issue where user edits to the field disappear when the `isEditable` flow emits `false`. The use of `Flow.combine` to combine `isEditable`, `properties.value`, and `_value` will toggle the displayed value between the `properties.value` and `_value` when `isEditable` emits. So user edits disappear, and reappear as `isEditable` goes from true to false to true. 

The solution is to use the latest value of either `properties.value` and `_value`, since values changing from editing and values changing from expression evaluation are mutually exclusive.


kaushikrw:
@sorenoid is this ready for review?

sorenoid:
@kaushikrw not yet, some issues with orientation changes

kaushikrw (approved):
LGTM!

PR #155, opened 2023-10-18
Forms: Micro-app crash guard

kaushikrw:
### Summary of changes

- Add an extra try-catch to prevent the app from crashing if accessing a `FeatureFormDefinition` (even though it shouldn't)


sorenoid (approved):
If you haven't can you please create an issue for the underlying crash and provide the details of the map that causes it?

kaushikrw:
Thanks @sorenoid. Created issue #238.

PR #158, opened 2023-10-19
Compose Map - graphics overlay

shubham7109:
Related issue: `#2948` 

### Description
PR to add the graphics overlay functionality to the compose map

### Changes made
- Added ResultExtensions.kt
- Added MapOperator.kt
- Updated compose Map 

### Testing
Here is a code snippet to test setting the viewpoint to the graphic geometry which is added to the MapView's graphics overlay.
```kotlin
private val borderPolyline by lazy {
    PolylineBuilder(SpatialReference.webMercator()) {
        addPoint(Point(-9981328.687124, 6111053.281447))
        addPoint(Point(-9946518.044066, 6102350.620682))
        addPoint(Point(-9872545.427566, 6152390.920079))
        addPoint(Point(-9838822.617103, 6157830.083057))
        addPoint(Point(-9446115.050097, 5927209.572793))
        addPoint(Point(-9430885.393759, 5876081.440801))
        addPoint(Point(-9415655.737420, 5860851.784463))
    }.toGeometry()
}

val polylineGraphic = Graphic(
    borderPolyline, SimpleLineSymbol(
        SimpleLineSymbolStyle.Dot,
        Color.red, 3F
    )
)

private val graphicsOverlay = GraphicsOverlay().apply {
    graphics.add(polylineGraphic)
}

@Composable
fun MainScreen() {
    val arcGISMap by remember { mutableStateOf(ArcGISMap(BasemapStyle.ArcGISStreets)) }

    MapOperator { mapOperator ->
        Map(
            arcGISMap = arcGISMap,
            onSingleTapConfirmed = { }
        )

        mapOperator.graphicsOverlays.add(graphicsOverlay)
        mapOperator.setViewpoint(Viewpoint(polylineGraphic.geometry!!))
    }
}

```


toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/Map.kt
```diff
@@ -20,34 +20,37 @@ package com.arcgismaps.toolkit.geocompose
...
- * @param overlay the composable overlays to display on top of the Map. Example, a compass, floorfilter etc.
+ * @param content the composable overlays to display on top of the Map. Example, a compass, floorfilter etc.
  * @since 200.3.0
  */
 @Composable
 public fun Map(
     modifier: Modifier = Modifier,
     arcGISMap: ArcGISMap? = null,
-    overlay: @Composable () -> Unit = {}
+    mapOperator: MapOperator,
```
puneet-pdx, line 48: We don't want to pass the mapOperator to the Map. Internally you should use `LocalMapView ` where ever you are using mapOperator in the Map()

shubham7109: @puneet-pdx Make sense, would it be okay if I were to create an instance of the local of the `LocalMapView` inside of the `Map()` or we should strictly only use the `LocalMapView` 
```kotlin
val mapView = (LocalMapView.current ?: return).apply {
    map = arcGISMap
}
// OR
LocalMapView.current?.map = arcGISMap
```

puneet-pdx: I think the idea is that we create the MapView in the Map if `LocalMapView.current` is null, for the use case when the user does not want to use MapOperator

shubham7109: Updated  👍

gunt0001 (changes requested):
As discussed, this implementation won't work.

PR #159, opened 2023-10-20
add viewpointChangedLambda

puneet-pdx:


toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/Map.kt
```diff
@@ -30,19 +31,22 @@ import androidx.compose.ui.tooling.preview.Preview
...
  * @param arcGISMap the [ArcGISMap] to be rendered by this composable
+ * @param onViewPointChanged lambda invoked when the viewpoint of the Map has changed
  * @param overlay the composable overlays to display on top of the Map. Example, a compass, floorfilter etc.
  * @since 200.3.0
  */
 @Composable
 public fun Map(
     modifier: Modifier = Modifier,
     arcGISMap: ArcGISMap? = null,
+    onViewPointChanged: (() -> Unit)? = null,
```
shubham7109, line 49: Not sure if there is a reason this is initialized to be null, but you could initialize it to a default lambda. 
```suggestion
    onViewPointChanged: () -> Unit = {},
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/Map.kt
```diff
@@ -69,6 +73,16 @@ public fun Map(
             mapView.onDestroy(lifecycleOwner)
         }
     }
+
+    LaunchedEffect(Unit) {
+        onViewPointChanged?.let { onViewPointChangedLambda ->
+            launch {
```
shubham7109, line 79: `LaunchedEffect` should create a coroutine scope, so a `launch` is not needed here

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/Map.kt
gunt0001: Defaulting the parameter to null is the correct thing to do. It expresses to the user that the parameter is "absent". See [Compose API guidelines](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-component-api-guidelines.md#nullable-parameter).

gunt0001: casing: It''s a "Viewpoint" that is changing, not a "ViewPoint", so the parameter name should be `onViewpointChanged `.

gunt0001: We need to consider passing the changed viewpoint to the event callback, so users can immediately process the changed viewpoint. Since there are two types of viewpoints the user might be interested in, we might have to support two different "viewpoint changed events", one passing a viewpoint of type `CenterAndScale` and another passing a viewpoint of type `BoundingGeometry`. We need to think about the naming, but something like this may work:
```kotlin
onViewPointChangedForScale: ((Viewpoint) -> Unit)? = null,
onViewPointChangedForVisibleArea: ((Viewpoint) -> Unit)? = null,
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/Map.kt
```diff
@@ -69,6 +73,16 @@ public fun Map(
...
+                mapView.viewpointChanged.collect {
+                    onViewPointChangedLambda()
+                }
+            }
+        }
+    }
```
gunt0001, line 85: We should always collect the viewpoint changes on the MapView, regardless of whether a callback parameter was passed or not. We also need to prevent the LaunchedEffect from recomposing when the parameter changes. You can achieve this with [rememberUpdatedState](https://developer.android.com/jetpack/compose/side-effects#rememberupdatedstate).

In principle the implementation would look like this:
```kotlin
val currentViewPointChanged = rememberUpdatedState(onViewPointChanged)
LaunchedEffect(Unit) {
   launch {
         currentViewPointChanged?.let {
            it()
         }
      }
   }
}
```

puneet-pdx: updated

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/Map.kt
puneet-pdx: I have updated it as per @gunt0001 's [comment](https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/pull/159/files#r1368802730).

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/Map.kt
puneet-pdx: As per our discussion, I have updated the parameter name

shubham7109 (approved):
Changes here look good to me 👍

gunt0001 (approved):
@puneet-pdx - looks good.

PR #160, opened 2023-10-22
Support LocationDisplay

gunt0001:
## Summary:
Adds support for LocationDisplay to the composable `Map()` function.

## Description:
- Adds a non-nullable `locationDisplay` parameter to the `Map()` function. The parameter is defaulted to a non-started LocationDisplay cached in a remember.
- A public helper function `rememberLocationDisplay` has been exposed to conveniently create and remember a LocationDisplay. This function also checks `ArcGISEnvironment.applicationContext`. If it isn't set, a context will be automatically set from the composition. This function is also used to initialize the defaulted `locationDisplay` parameter.
- A micro app `LocationDisplayApp` has been added to test & demo Map usage with a location display.


toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/Map.kt
```diff
@@ -28,27 +28,32 @@ import androidx.compose.ui.semantics.contentDescription
...
     arcGISMap: ArcGISMap? = null,
+    locationDisplay: LocationDisplay = rememberLocationDisplay(),
     overlay: @Composable () -> Unit = {}
 ) {
     val lifecycleOwner = LocalLifecycleOwner.current
     val context = LocalContext.current
     val mapView = remember { MapView(context) }.apply {
-        map = arcGISMap
+        this.map = arcGISMap
+        this.locationDisplay = locationDisplay
```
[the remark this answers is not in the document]
gunt0001: Yes, good catch! Updated.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/Map.kt
```diff
@@ -71,6 +76,29 @@ public fun Map(
...
+ * @param init called when the [LocationDisplay] is created to configure its initial state
+ * @since 200.3.0
+ */
+@Composable
+public inline fun rememberLocationDisplay(
+    key: Any? = null,
+    crossinline init: LocationDisplay.() -> Unit = {}
+): LocationDisplay {
+    if (ArcGISEnvironment.applicationContext == null) {
+        ArcGISEnvironment.applicationContext = LocalContext.current
```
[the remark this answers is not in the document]
gunt0001: It should always work. Since `rememberLocationDisplay` is a composable function we can get a `Context` from the composition via `LocalContext.current`. The setter of `ArcGISEnvironment.applicationContext` will retrieve the application context internally via `Context.applicationContext`. In theory we could write here:
```
if (ArcGISEnvironment.applicationContext == null) {
        ArcGISEnvironment.applicationContext = LocalContext.current.applicationContext
```
But that's redundant, since the API code will take care of it.

microapps/LocationDisplayApp/app/src/main/java/com/arcgismaps/toolkit/locationdisplayapp/screens/MainScreen.kt
```diff
@@ -0,0 +1,212 @@
...
+                        Switch(
+                            checked.value,
+                            onCheckedChange = {
+                                checked.value = it
+                                if (locationDisplay.isStarted) {
+                                    locationDisplay.stop(scope)
+                                    checked.value = false
+                                } else {
+                                    locationDisplay.start(scope) { result ->
+                                        checked.value = result.isSuccess
```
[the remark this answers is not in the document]
gunt0001: Good catch, update.

gunt0001 (commented):
@hud10837 - thanks, I addressed your comments, please take another look.

microapps/LocationDisplayApp/app/src/main/java/com/arcgismaps/toolkit/locationdisplayapp/screens/MainScreen.kt
```diff
@@ -0,0 +1,214 @@
...
+ * If the location display fails to start, an error message is displayed below the switch.
+ * An action button provides a choice of [LocationDisplayAutoPanMode] options. Upon selecting an
+ * option, the auto pan mode of the location display is set.
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun MainScreen() {
+    val scope = rememberCoroutineScope()
+    val arcGISMap = remember { ArcGISMap(BasemapStyle.ArcGISImagery) }
+    var checked by remember { mutableStateOf(false) }
```
hud10837, line 66: I think this needs `rememberSaveable`. If I turn off the switch then rotate the device, it turns back on

hud10837 (approved):
LGTM, one change 👍

microapps/LocationDisplayApp/app/src/main/java/com/arcgismaps/toolkit/locationdisplayapp/screens/MainScreen.kt
gunt0001: Good point. While it's easy enough to use `rememberSaveable` for the checked property, so I can easily keep the on/off state of the LDS preserved between configuration changes, preserving the auto pan mode state would be more involved. The way to achieve that in a real app would be to use a view model. But this micro app should remain super simple. by introducing a view model we would loose the simplicity. Given that, I would argue we don't try to handle config changes here.

hud10837: I agree, handling the auto pan mode would be too complicated for the simple app. Happy to have no config change handling here

microapps/LocationDisplayApp/app/src/main/java/com/arcgismaps/toolkit/locationdisplayapp/screens/MainScreen.kt
```diff
@@ -0,0 +1,214 @@
...
+import com.arcgismaps.mapping.ArcGISMap
+import com.arcgismaps.mapping.BasemapStyle
+import com.arcgismaps.mapping.view.LocationDisplay
+import com.arcgismaps.toolkit.geocompose.Map
+import com.arcgismaps.toolkit.geocompose.rememberLocationDisplay
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+
+/**
+ * Displays a [Map] with a [LocationDisplay] obtained with [rememberLocationDisplay].
```
puneet-pdx, line 55: [Map] here points to Kotlin.collections.Map interface

puneet-pdx: could we merge this and make sure it gets updated when we do the name change refactor ??

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/Map.kt
```diff
@@ -28,36 +28,43 @@ import androidx.compose.ui.semantics.contentDescription
...
 import com.arcgismaps.mapping.ArcGISMap
+import com.arcgismaps.mapping.view.LocationDisplay
 import com.arcgismaps.mapping.view.MapView
 
 /**
  * A compose equivalent of the [MapView].
  *
  * @param modifier Modifier to be applied to the Map
  * @param arcGISMap the [ArcGISMap] to be rendered by this composable
+ * @param locationDisplay the [LocationDisplay] used by the [Map]
```
puneet-pdx, line 41: [Map] here points to Kotlin.collections.Map interface

puneet-pdx (approved):
@gunt0001 looks good. Just a comment about the doc.

gunt0001:
Thanks @puneet-pdx , I addressed your comments.

PR #161, opened 2023-10-23
Forms: `DateTimeField` changes

kaushikrw:
### Summary of changes

- Updated `DateTimeField` to use `BaseTextField`.
- Updated `Required` validation to occur only after gaining focus.
- Added compose preview.

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/DateTimeField.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/DateTimeField.kt
@@ -28,10 +28,15 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.runtime.saveable.rememberSaveable
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.tooling.preview.Preview
 import com.arcgismaps.toolkit.featureforms.R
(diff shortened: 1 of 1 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/DateTimeField.kt
```diff
@@ -18,216 +18,109 @@
...
+            // if the field was focused and is required, validate the current value
+            if (wasFocused && isRequired && epochMillis == null) {
+                Text(
+                    text = stringResource(id = R.string.required),
+                    color = MaterialTheme.colorScheme.error
+                )
+            } else {
+                Text(
+                    text = state.description,
+                    modifier = Modifier.semantics { contentDescription = "description" },
```
sorenoid, line 86:
```suggestion
                    text = state.description,
                    modifier = Modifier.semantics { contentDescription = "description" },
                    color = MaterialTheme.colorScheme.onBackground
```

sorenoid: Without this an empty required field has a red description after being edited to be non-empty.

sorenoid (changes requested):
one small change.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/DateTimeField.kt
kaushikrw: I don't think that should be needed since the `required` and the `description` composable `Text` are different and by default the `description` color would be `LocalContentColor`

sorenoid: I tested it, but now on re-test I don't see it. I must have been incorrect. Never mind!

sorenoid (approved):
LGTM

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/DateTimeField.kt
kaushikrw: That is odd, will keep an eye out if something like this shows up.

kaushikrw:
Thanks @sorenoid, merging!

PR #162, opened 2023-10-23
consume update value

sorenoid:
This updates the SDK to 4036 and uses the `formattedValue` when the `valueChanged` event is emitted. `updateValue` is used to update attribute values. Its usage is wrapped in a try-catch until the API is changed so it no longer throws.

There are some known issues with this version of core but I think we can work with it for a weekly build.

issue 1: [TreeTrimmingHealthCheck](https://devtopia.esri.com/runtime/apollo/issues/254) map
issue 2: [AdhocUItest_Switch](https://devtopia.esri.com/runtime/apollo/issues/181) map

### DIFF
--- a/microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapScreen.kt
+++ b/microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapScreen.kt
@@ -1,5 +1,6 @@
 package com.arcgismaps.toolkit.featureformsapp.screens.map
 
+import android.util.Log
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material.icons.Icons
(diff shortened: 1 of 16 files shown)


microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapScreen.kt
```diff
@@ -57,7 +58,12 @@ fun MapScreen(mapViewModel: MapViewModel = hiltViewModel(), onBackPressed: () ->
                     scope.launch { mapViewModel.rollbackEdits(EditingTransactionState.NotEditing) }
                 },
                 onSave = {
-                    scope.launch { mapViewModel.commitEdits(EditingTransactionState.NotEditing) }
+                    scope.launch {
+                        mapViewModel.commitEdits(EditingTransactionState.NotEditing)
+                            .onFailure {
+                                Log.w("Forms","applying edits from feature form failed with ${it.message}")
```
kaushikrw, line 64: Can we also add a toast here?

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/DateTimeFieldState.kt
```diff
@@ -112,7 +113,7 @@ internal class DateTimeFieldState(
                         label = field.label,
                         placeholder = field.hint,
                         description = field.description,
-                        value = field.value,
+                        value = field.formattedFlow(scope),
```
kaushikrw, line 116: It seems like the dates after saving and re-opening are off due to time zone mismatch. Are we planning to fix it separately? I may have forgot what we talked about..

kaushikrw (changes requested):
This looks good! The change to use the `codedValue.name` reduced quite a bit of code.

Just some minor nitpicks and a question regarding `DateTimeField` time zone conversion.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/DateTimeFieldState.kt
sorenoid: it is being fixed in core.

kaushikrw (approved):
@sorenoid LGTM!

PR #169, opened 2023-10-25
`Forms`: Optimized initial expression evaluation

kaushikrw:
### Summary of changes

- Updated the `initialEvaluation` state to use `rememberSaveable` to not re-trigger initial eval again on orientation change.
- The `InitializingExpressions` now appears on top of the `FeatureFormContent` and disappears when `initialEvaluation` turns true. This is because we don't want the `FeatureFormContent` to recompose since this is an expensive operation and the `evaluateExpressions` will trigger updates on individual elements anyway.
- This also adds a 300ms artificial delay after the expressions have been evaluated to avoid flickering which would happen in most cases when the `evaluateExpressions` runs fast. Reasoning behind this is explained [here](https://uxmag.com/articles/let-your-users-wait).


sorenoid (approved):
👍

kaushikrw:
thanks @sorenoid, merging!

PR #172, opened 2023-10-27
add implementation for gesture events

puneet-pdx:
## Summary:
Adds support for Gesture events to the composable `MapView`

## Description:
- Adds a nullable property for the following gesture events to the `MapView` function,
    - onUp
    - onDown
    - onSingleTapConfirmed
    - onDoubleTap
    - onLongPress
    - onTwoPointerTap
    - onPan

### DIFF
--- a/toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
+++ b/toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
@@ -125,20 +125,9 @@ public fun MapView(
         }
     }
 
-    val currentViewPointChanged by rememberUpdatedState(onViewpointChanged)
-
-    LaunchedEffect(Unit) {
-        launch {
-            mapView.viewpointChanged.collect {
-                currentViewPointChanged?.let {
-                    it()
-                }
-            }
-        }
-    }
-
     MapViewEventHandler(
         mapView,
+        onViewpointChanged,
         onInteractingChanged,
         onRotate,
         onScale,
@@ -153,11 +142,12 @@ public fun MapView(
 }
 
 /**
- * Sets up the callbacks for all the gesture events.
+ * Sets up the callbacks for all the MapView events.
  */
 @Composable
 private fun MapViewEventHandler(
     mapView: MapView,
+    onViewpointChanged: (() -> Unit)?,
     onInteractingChanged: ((isInteracting: Boolean) -> Unit)?,
     onRotate: ((RotationChangeEvent) -> Unit)?,
     onScale: ((ScaleChangeEvent) -> Unit)?,
@@ -169,6 +159,7 @@ private fun MapViewEventHandler(
     onTwoPointerTap: ((TwoPointerTapEvent) -> Unit)?,
     onPan: ((PanChangeEvent) -> Unit)?
 ) {
+    val currentViewPointChanged by rememberUpdatedState(onViewpointChanged)
     val currentOnInteractingChanged by rememberUpdatedState(onInteractingChanged)
     val currentOnRotate by rememberUpdatedState(onRotate)
     val currentOnScale by rememberUpdatedState(onScale)
@@ -181,6 +172,13 @@ private fun MapViewEventHandler(
     val currentOnPan by rememberUpdatedState(onPan)
 
     LaunchedEffect(Unit) {
+        launch {
+            mapView.viewpointChanged.collect {
+                currentViewPointChanged?.let {
+                    it()
+                }
+            }
+        }
         launch(Dispatchers.Main.immediate) {
             mapView.isInteracting.collect { isInteracting ->
                 currentOnInteractingChanged?.let {
--- a/toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
+++ b/toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
@@ -33,9 +33,18 @@ import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.viewinterop.AndroidView
 import com.arcgismaps.ArcGISEnvironment
 import com.arcgismaps.mapping.ArcGISMap
+import com.arcgismaps.mapping.view.DoubleTapEvent
+import com.arcgismaps.mapping.view.DownEvent
 import com.arcgismaps.mapping.view.LocationDisplay
+import com.arcgismaps.mapping.view.LongPressEvent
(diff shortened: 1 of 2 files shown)


toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -101,6 +124,17 @@ public fun MapView(
                 }
             }
         }
+        setupCallbacksForGestureEvents(
```
gunt0001, line 127: Each of the callback parameters needs to be remembered in a `rememberUpdatedState`, otherwise the LaunchedEffect may get recomposed when one of these parameters changes. I suggest you remember them all just below `currentViewPointChanged` (line 118), i.e.:
```kotlin
val currentViewPointChanged by rememberUpdatedState(onViewpointChanged)
val currentOnUp by rememberUpdatedState(onUp)
...
```
Then pass the `currentXYZ` values to `setupCallbacksForGestureEvents`. But you need to test what happens when callback parameters change, i.e. set to null/non-null subsequently. It might actually not work by passing them to this helper function since the function gets only called once and any parameter updates might not get applied. I'm not fully sure, needs testing...

puneet-pdx: I tested and it does not work by passing values to the helper function. I have made the changes

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -61,6 +82,16 @@ public fun MapView(
     locationDisplay: LocationDisplay = rememberLocationDisplay(),
     mapViewInteractionOptions: MapViewInteractionOptions = MapViewInteractionOptionDefaults,
     onViewpointChanged: (() -> Unit)? = null,
+    isInteracting: ((Boolean) -> Unit)? = null,
```
gunt0001, line 85: I suggest we name the parameter `onInteractionChanged` to be consistent with the other MapView related events, where each is prefixed with "on". Also suffixing it with "Changed" implies that there is a start and an end to map biew interaction, i.e. at start it passes `true` and at the end it passes `false`.
```suggestion
    onInteractionChanged: ((Boolean) -> Unit)? = null,
```

gunt0001 (approved):
@puneet-pdx - apart from a naming suggestion this looks good.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
puneet-pdx: >at start it passes true and at the end it passes false

I don't think that `onInteractionChanged` depicts accurately what it happening.  The api doc states 
`Indicates whether the user is currently interacting with the GeoView` 

With that in mind, in my opinion `onInteractionStarted` makes more sense. So then when the interaction starts it returns `true` and when it ends it returns `false`

and the doc should be changed from 
`@param isInteracting lambda invoked when the user is currently interacting with the composable MapView` 

to 
`@param onInteractionStarted lambda invoked when the user starts and ends the interaction with the composable MapView`

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -51,6 +62,16 @@ public val MapViewInteractionOptionDefaults: MapViewInteractionOptions = MapView
  * @param locationDisplay the [LocationDisplay] used by the composable [com.arcgismaps.toolkit.geocompose.MapView]
  * @param mapViewInteractionOptions the [MapViewInteractionOptions] used by this composable [com.arcgismaps.toolkit.geocompose.MapView]
  * @param onViewpointChanged lambda invoked when the viewpoint of the composable MapView has changed
+ * @param isInteracting lambda invoked when the user is currently interacting with the composable MapView
```
shubham7109, line 65:
```suggestion
 * @param isInteracting lambda invoked when the user is currently interacting with the composable [MapView]
```

At first, I was considering to suggest using `[com.arcgismaps.toolkit.geocompose.MapView]` for the doc in this PR but I feel like it would be somewhat overwhelming especially when the doc is previewed in Android Studio: 

image

Maybe (in this / a future issue), entirely remove the import `import com.arcgismaps.mapping.view.MapView` and instead create the MapView using:
```kotlin
val mapView = remember { MapView(context) } // Before
val mapView = remember { com.arcgismaps.mapping.view.MapView(context) } // After
```

This way we can use `[MapView]` in the docs to freely reference the composable MapView rather than the SDK MapView. Thoughts? @puneet-pdx @gunt0001

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -33,9 +33,20 @@ import androidx.compose.ui.tooling.preview.Preview
...
+import com.arcgismaps.mapping.view.LongPressEvent
 import com.arcgismaps.mapping.view.MapView
 import com.arcgismaps.mapping.view.MapViewInteractionOptions
+import com.arcgismaps.mapping.view.PanChangeEvent
+import com.arcgismaps.mapping.view.RotationChangeEvent
+import com.arcgismaps.mapping.view.ScaleChangeEvent
+import com.arcgismaps.mapping.view.SingleTapConfirmedEvent
+import com.arcgismaps.mapping.view.TwoPointerTapEvent
+import com.arcgismaps.mapping.view.UpEvent
+import kotlinx.coroutines.CoroutineScope
```
shubham7109, line 48: Unused import: 
```suggestion
```

shubham7109 (approved):
Gesture events implementation looks good, so approving this for now. I do have one concern with the documentation structure which I have addressed in the comment below.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -93,6 +123,7 @@ public fun MapView(
     }
 
     val currentViewPointChanged by rememberUpdatedState(onViewpointChanged)
+
     LaunchedEffect(Unit) {
         launch {
             mapView.viewpointChanged.collect {
```
gunt0001, line 129: I think this event can also be handled in the `MapViewEventHandler()` function.

gunt0001 (approved):
@puneet-pdx - looks good, one comment.

PR #181, opened 2023-10-31
`Forms`: Optimized field state creation

kaushikrw:
### Summary of changes

- Optimized how the field state `Map<Int, BaseFieldState>` is created in `rememberFieldStates()` to return non-nullable states.


sorenoid (approved):
LGTM

kaushikrw:
thanks @sorenoid merging!

PR #184, opened 2023-11-02
use datetime raw value

sorenoid:
The formattedValue for DateTime elements is no longer a format string, it is truly a formatted string. This isn't what we need for interacting with the picker, we need a Long.

`BaseFieldState` is changed to be generic in the value type. This is a String type for all but the DateTime, for which it is a nullable `Instant?`.

`DateTimeFieldState` now uses `Instant?` rather than a Long? for the value, max, and min properties.


kaushikrw (changes requested):
This looks good! Some minor comments below.

kaushikrw (approved):
LGTM!

PR #186, opened 2023-11-02
Draw Status

shubham7109:
PR to implement `DrawStatus` to the Composable `MapView`

### DIFF
--- a/toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
+++ b/toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
@@ -76,6 +76,7 @@ import kotlinx.coroutines.launch
  * @param onLongPress lambda invoked when a user holds a pointer on the composable MapView
  * @param onTwoPointerTap lambda invoked when a user taps two pointers on the composable MapView
  * @param onPan lambda invoked when a user drags a pointer or pointers across composable MapView
+ * @param drawStatus lambda invoked when the draw status of the composable MapView is changed
  * @param overlay the composable overlays to display on top of the composable MapView. Example, a compass, floorfilter etc.
  * @since 200.3.0
  */
--- a/toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
+++ b/toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
@@ -76,7 +76,7 @@ import kotlinx.coroutines.launch
  * @param onLongPress lambda invoked when a user holds a pointer on the composable MapView
  * @param onTwoPointerTap lambda invoked when a user taps two pointers on the composable MapView
  * @param onPan lambda invoked when a user drags a pointer or pointers across composable MapView
- * @param onDrawStatusChanged lambda invoked when the draw status of the composable MapView is changed
+ * @param onDrawStatusChanged lambda invoked when the draw status of the composable MapView is changes
  * @param overlay the composable overlays to display on top of the composable MapView. Example, a compass, floorfilter etc.
  * @since 200.3.0
  */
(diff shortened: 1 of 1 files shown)


toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -100,6 +102,7 @@ public fun MapView(
     onLongPress: ((LongPressEvent) -> Unit)? = null,
     onTwoPointerTap: ((TwoPointerTapEvent) -> Unit)? = null,
     onPan: ((PanChangeEvent) -> Unit)? = null,
+    drawStatus: ((DrawStatus) -> Unit)? = null,
```
hud10837, line 105: I think this should be `onDrawStatusChanged`. Otherwise, it looks like the user should be passing in a draw status (although the lambda wouldn't allow that). In the view-based api, it's okay because it's a state flow

hud10837 (changes requested):

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
shubham7109: Makes sense, I've updated the parameter to `onDrawStatusChanged`.

shubham7109:
@hud10837, thanks for the feedback. I also would like to slip in a change for `viewpointChanged` to use `Dispatchers.Main.immediate` here. Ready for your review.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -183,10 +183,10 @@ private fun MapViewEventHandler(
     val currentOnLongPress by rememberUpdatedState(onLongPress)
     val currentOnTwoPointerTap by rememberUpdatedState(onTwoPointerTap)
     val currentOnPan by rememberUpdatedState(onPan)
-    val currentDrawStatus by rememberUpdatedState(drawStatus)
+    val currentOnDrawStatusChanged by rememberUpdatedState(onDrawStatusChanged)
 
     LaunchedEffect(Unit) {
-        launch {
+        launch(Dispatchers.Main.immediate) {
```
hud10837, line 189: Thanks for adding this

hud10837 (approved):

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -257,6 +263,13 @@ private fun MapViewEventHandler(
                 }
             }
         }
+        launch(Dispatchers.Main.immediate) {
```
gunt0001, line 266: `Dispatchers.Main.immediate` is only required for the gesture events. Events raised by Core will always be raised on a Core thread, so we cannot avoid a re-dispatch when collecting on the composition's coroutine context.
```suggestion
        launch() {
```

Same above fore `viewpointChanged`...

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -75,6 +76,7 @@ import kotlinx.coroutines.launch
  * @param onLongPress lambda invoked when a user holds a pointer on the composable MapView
  * @param onTwoPointerTap lambda invoked when a user taps two pointers on the composable MapView
  * @param onPan lambda invoked when a user drags a pointer or pointers across composable MapView
+ * @param onDrawStatusChanged lambda invoked when the draw status of the composable MapView is changed
```
gunt0001, line 79:
```suggestion
 * @param onDrawStatusChanged lambda invoked when the draw status of the composable MapView changes
```

shubham7109:
Thanks for the feedback. I've reverted the dispatchers change with `viewpointChanged` and `drawStatus` and applied the feedback. Ready for your review. @gunt0001

gunt0001 (approved):
@shubham7109 - looks good 👍

PR #187, opened 2023-11-03
add support for attribution text

puneet-pdx:


toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -85,11 +87,13 @@ public fun MapView(
     graphicsOverlays: GraphicsOverlayCollection = rememberGraphicsOverlayCollection(),
     locationDisplay: LocationDisplay = rememberLocationDisplay(),
     wrapAroundMode: WrapAroundMode = WrapAroundMode.EnabledWhenSupported,
+    attributionBarVisible: Boolean = true,
     geometryEditor: GeometryEditor? = null,
     mapViewInteractionOptions: MapViewInteractionOptions = MapViewInteractionOptions(),
     viewLabelProperties: ViewLabelProperties = ViewLabelProperties(),
     selectionProperties: SelectionProperties = SelectionProperties(),
     onViewpointChanged: (() -> Unit)? = null,
+    onAttributionTextChanged: ((String) -> Unit)? = null,
```
hud10837, line 96: Just as a general question, what was the reasoning behind using null rather than `{}` for the default values of event params?

hud10837 (approved):

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
shubham7109: This comment could be helpful: https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/pull/159#discussion_r1368759924
> Defaulting the parameter to null is the correct thing to do. It expresses to the user that the parameter is "absent". See [Compose API guidelines](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-component-api-guidelines.md#nullable-parameter).

hud10837: Okay, I suppose the difference between these lambdas and `Button`'s `onClick` is that a Button must have an onClick event, whereas we don't expect users to always do something when these events are received.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -85,11 +87,13 @@ public fun MapView(
     graphicsOverlays: GraphicsOverlayCollection = rememberGraphicsOverlayCollection(),
     locationDisplay: LocationDisplay = rememberLocationDisplay(),
     wrapAroundMode: WrapAroundMode = WrapAroundMode.EnabledWhenSupported,
+    attributionBarVisible: Boolean = true,
```
gunt0001, line 90: Should the parameter be suffixed with "is"?
```suggestion
    isAttributionBarVisible: Boolean = true,
```

PR #189, opened 2023-11-03
searchable items map list screen

sorenoid:
makes the list of maps searchable by title or item id. the title search is case insensitive.

### DIFF
--- a/microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/browse/MapListScreen.kt
+++ b/microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/browse/MapListScreen.kt
@@ -80,9 +80,8 @@ fun MapListScreen(
 ) {
     val uiState by mapListViewModel.uiState.collectAsState()
     val lazyListState = rememberLazyListState()
-    var searchText by remember { mutableStateOf("") }
     Scaffold(topBar = {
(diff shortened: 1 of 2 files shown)


microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/browse/MapListScreen.kt
```diff
@@ -107,7 +113,13 @@ fun MapListScreen(
                         state = lazyListState,
                         horizontalAlignment = Alignment.CenterHorizontally
                     ) {
-                        items(uiState.data) { item ->
+                        items(
+                            uiState.data.filter {
```
kaushikrw, line 117: I think the `MapListViewModel` is the perfect place to hoist the search and filtering instead of here. When the search text changes, a new `MapListUIState` would be emitted with the filtered list.

kaushikrw (changes requested):
Thanks for adding this!

The layout of the search text field looks a little off. It's size and padding do not seem uniform in the top bar. There is a search bar material 3 [design](https://m3.material.io/components/search/overview) that perhaps would be appropriate here, although it might be a bit more work..

sorenoid:
@kaushikrw when we support both agol and enterprise portals I think there will be cause for a navigation icon, and the full search bar and search view design.  I am not sure exactly what you thought was off with the search field. Can you take another look? I've made the layout consistent between landscape and portrait, and made the text box more rounded. I have also hoisted the search text filtering to the view model as suggested.

kaushikrw (approved):
LGTM!

PR #191, opened 2023-11-07
Update MapView doc pattern

shubham7109:
PR to address the pattern used for the compose MapView doc. 
 
Related issue: `#3164`

### DIFF
--- a/toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
+++ b/toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
@@ -62,7 +62,7 @@ import kotlinx.coroutines.launch
  * @param arcGISMap the [ArcGISMap] to be rendered by this composable MapView
  * @param graphicsOverlays the [GraphicsOverlayCollection] used by this composable MapView
  * @param locationDisplay the [LocationDisplay] used by the composable MapView
- * @param geometryEditor the [GeometryEditor] used by the composable MapView to create and edit geometries by user interaction.
+ * @param geometryEditor the [GeometryEditor] used by the composable MapView to create and edit geometries by user interaction
  * @param mapViewInteractionOptions the [MapViewInteractionOptions] used by this composable MapView
  * @param viewLabelProperties the [ViewLabelProperties] used by the composable MapView
  * @param selectionProperties the [SelectionProperties] used by the composable MapView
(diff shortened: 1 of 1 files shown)


hud10837 (commented):
This is a good change, but I think we can probably also remove the references to `composable [MapView]` from all of the parameters here. These are parameters on the composable MapView, so there's no need to link to it here. I'm not sure what we can replace it with though, perhaps just:

```
@param onScale invoked when a user performs a pinch gesture on the MapView
```

shubham7109:
@hud10837 Sure, I think it would be fine to remove the in-line references (`[]`) but I would still think there is value in using the prefix (view-based/composable) when `MapView` is referenced. I believe it will help clarify which type of MapView the doc is referring to. 

Adding on your suggestion would you say it would be acceptable to use:
```
* @param onScale invoked when a user performs a pinch gesture on the composable MapView
```

hud10837:
I think this would be acceptable 👍

shubham7109:
Updated PR, ready for your review @hud10837

hud10837 (approved):
Thanks for those changes 👍

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -56,17 +56,17 @@ import com.arcgismaps.mapping.view.geometryeditor.GeometryEditor
...
- * @param locationDisplay the [LocationDisplay] used by the composable [com.arcgismaps.toolkit.geocompose.MapView]
- * @param geometryEditor the [GeometryEditor] used by the composable [com.arcgismaps.toolkit.geocompose.MapView] to create and edit geometries by user interaction.
- * @param mapViewInteractionOptions the [MapViewInteractionOptions] used by this composable [com.arcgismaps.toolkit.geocompose.MapView]
- * @param viewLabelProperties the [ViewLabelProperties] used by the composable [com.arcgismaps.toolkit.geocompose.MapView]
- * @param selectionProperties the [SelectionProperties] used by the composable [com.arcgismaps.toolkit.geocompose.MapView]
- * @param grid represents the display of a coordinate system [Grid] on the composable [com.arcgismaps.toolkit.geocompose.MapView]
+ * @param arcGISMap the [ArcGISMap] to be rendered by this composable MapView
+ * @param graphicsOverlays the [GraphicsOverlayCollection] used by this composable MapView
+ * @param locationDisplay the [LocationDisplay] used by the composable MapView
+ * @param geometryEditor the [GeometryEditor] used by the composable MapView to create and edit geometries by user interaction.
```
gunt0001, line 65: nitpick - remove full stop:
```suggestion
 * @param geometryEditor the [GeometryEditor] used by the composable MapView to create and edit geometries by user interaction
```

gunt0001 (approved):
@shubham7109 - looks good, one minor comment

PR #194, opened 2023-11-07
MapView TimeExtent

shubham7109:
PR to add an `onTimeExtentChanged` and `timeExtent` properties to the MapView.


toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -100,6 +104,7 @@ public fun MapView(
     onLongPress: ((LongPressEvent) -> Unit)? = null,
     onTwoPointerTap: ((TwoPointerTapEvent) -> Unit)? = null,
     onPan: ((PanChangeEvent) -> Unit)? = null,
+    onTimeExtentChanged: ((TimeExtent?) -> Unit)? = null,
```
gunt0001, line 107: I think we should move this parameter next to the `timeExtent` parameter. Alternatively, lets consider if a single `TimeExtentState` or `TimeAwareState` parameter would make sense here:
```kotlin
@Stable
data class TimeAwareState(
   val timeExtent: TimeExtent,
   val onTimeExtentChanged: ((TimeExtent?) -> Unit)? = null
)
```

shubham7109: Added the feedback from today's discussion to add the `timeExtent` next to the `onTimeExtentChanged` parameter.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -178,9 +185,11 @@ private fun MapViewEventHandler(
     onLongPress: ((LongPressEvent) -> Unit)?,
     onTwoPointerTap: ((TwoPointerTapEvent) -> Unit)?,
     onPan: ((PanChangeEvent) -> Unit)?,
+    onTimeExtentChanged: ((TimeExtent?) -> Unit)?,
```
gunt0001, line 188: Can you move this up just above the `onViewpointChanged` parameter so it matches the parameter order in the `MapView()` function.

gunt0001 (approved):
@shubham7109 - looks good, one comment.

shubham7109:
Thanks for the review @puneet-pdx, I've added the changes and updated the branch with the base feature branch. Ready for your re-review.

PR #195, opened 2023-11-07
MapView Rotation & Scale

shubham7109:
PR to add `onMapRotationChanged` & `onMapScaleChanged` lambda parameters to the composable MapView.


toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
hud10837, line 1: I'm not sure these should be placed between the gesture events. Map scale and rotation changes with set viewpoint, so I would put them above or below `onViewpointChanged`, and keep all of the gesture events together as a block

shubham7109: Good point, moved the two lamdas below the `onViewpointChanged`

hud10837 (approved):
LGTM 👍

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -70,6 +70,8 @@ import kotlinx.coroutines.launch
  * @param backgroundGrid the default color and context grid behind the map surface
  * @param wrapAroundMode the [WrapAroundMode] to specify whether continuous panning across the international date line is enabled
  * @param onViewpointChanged lambda invoked when the viewpoint of the composable MapView has changed
+ * @param onMapRotationChanged lambda invoked with the composable MapView's current rotation
+ * @param onMapScaleChanged lambda invoked with the composable MapView's current scale
```
gunt0001, line 74:
```suggestion
 * @param onMapScaleChanged lambda invoked when the scale of this composable MapView has changed
```

gunt0001 (approved):
@shubham7109 - some doc comments, otherwise looks good.

shubham7109:
Thanks for the reviews @hud10837 & @gunt0001. Updated this branch with the target branch, now merging it in.

PR #198, opened 2023-11-08
Forms: DatePickerDialog scaling

kaushikrw:
### Summary of changes

Fixes the improper scaling of `DateTimePickerDialog` on devices that are narrower than the minimum width needed to fit the `DatePicker`. This is 360dp as mentioned [here](https://issuetracker.google.com/issues/297249202#comment3).

The fix is to scale down the Dialog using the extension `Modifier.scaleIfNarrow(minWidth: Dp)` if the available screen width is less than required. Hence it will not effect devices that have sufficient width. 

Since this effect occurs only in portrait mode, the extension `Modifier.widthWithOrientation(width: Dp)` is provided that sets the width constraint accordingly.

| Before  | After |
| ------------- | ------------- |
|    |  |

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePicker.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePicker.kt
@@ -18,6 +18,8 @@
 
 package com.arcgismaps.toolkit.featureforms.components.datetime.picker
 
+import android.content.res.Configuration
+import android.util.Log
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.ColumnScope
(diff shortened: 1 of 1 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/datetime/picker/DateTimePicker.kt
```diff
@@ -349,6 +355,33 @@ private fun DateTimePickerDialog(
...
+        this.requiredWidth(width)
+    }
+}
+
+/**
+ * Scales the content appropriately if the current screen width is less than the [minWidth].
+ */
+internal fun Modifier.scaleIfNarrow(minWidth: Dp): Modifier = composed {
+    val screenWidth = LocalConfiguration.current.screenWidthDp.dp
+    val scale = if (screenWidth <= minWidth)
```
sorenoid, line 381: slight nitpick. equality of floats is flaky anyway and we can avoid a floating point division this way.

```suggestion
    val scale = if (screenWidth < minWidth)
```

sorenoid (approved):
LGTM. I didn't verify but it looks reasonable. one nitpick.

kaushikrw:
Thanks @sorenoid, merging.

PR #201, opened 2023-11-10
fix ff tests

sorenoid:
some slight modifications to get the tests compiling and running successfully again

* add new parameters from type safe state objects
* remove unnecessary modifier from BaseTextField


kaushikrw (approved):
LGTM!

PR #207, opened 2023-11-16
`Forms` : Microapp SideSheet support

kaushikrw:
### Summary of changes

- Introduced a new `SideSheetLayout` which uses a custom layout to either display the bottomsheet in the typical place when in portrait or as a side sheet when in landscape based on the screen size.
- This calculation of screen size is done using Jetpack WindowManager. More info [here](https://developer.android.com/guide/topics/large-screens/support-different-screen-sizes).
- When an expanded size class is available, the SideSheet occupies 40% of the available width.
- When presented as a SideSheet, the underlying container is still a bottomsheet and hence the swipe behavior will remain the same.
- Added animations to the sheet enter and exit.

With this change, in the SideSheet mode, the controls to close/submit the feature form are presented in the top bar which might not be ideal. If this turns out to be confusing, these could be embedded into the bottomsheet itself.

#### Screenshots

| Orientation | Phone (Pixel 6A)  | Tablet |
| ------- | ------- | ------- |
| Portrait |    |    |
| Landscape |   |   |


microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/bottomsheet/SideSheetLayout.kt
```diff
@@ -0,0 +1,92 @@
...
+ * @param windowSizeClass The current [WindowSizeClass].
+ * @param sheetOffsetY An offset in pixels for the [sheetContent] in the Y axis.
+ * @param modifier The [Modifier]
+ * @param maxWidth A maximum width if specified will be enforced only when the orientation is portrait
+ * and the [WindowSizeClass.windowWidthSizeClass] is not [WindowWidthSizeClass.EXPANDED]. Otherwise
+ * this is set to [Infinity] which indicates to the maximum width available.
+ * @param sheetContent The sheet content lambda which is passed the width and height of the layout in pixels.
+ */
+@Composable
+fun SideSheetLayout(
```
sorenoid, line 47: This name is a bit misleading. It is the layout for the sheet content when it is centered or when it is on the side.

sorenoid (changes requested):
Looks great! Just some naming feedback. Not sure what I would suggest though

sorenoid (approved):
preapproved!

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/bottomsheet/SideSheetLayout.kt
kaushikrw: Renamed it to `SheetLayout`.

kaushikrw:
Thanks @sorenoid, merging.

PR #211, opened 2023-11-21
Update SDK dependency to use latest

shubham7109:
Note: Remember to remove the build tag before publishing.


eri9000 (approved):
LGTM 
Would suggest creating an issue to remove the build tag so that we don't forget.

gunt0001 (approved):

shubham7109:
@eri9000 The step to remove the build tag is part of the Toolkit release tasks.

PR #213, opened 2023-11-29
Improve WebView Display

eri9000:
Refactors the Authenticator composable. 

Changes: 
Introduces a new composable `AuthenticatorDelegate` that has all the logic previously held by `Authenticator` with a new parameter `container`. This allows for the calling composable to pass in the desired container to display the `Server trust` and `username/password` prompt. 
This is used by `DialogAuthenticator.`

### Testing
---
The current workflows should work the same way as before. In the micro app, when using a `WebView,` it should use the full app screen and the `server trust /username password` prompts should be displayed in an Alert Dialog

### DIFF
--- a/toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
+++ b/toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
@@ -60,6 +60,7 @@ public fun Authenticator(
 ) {
     AuthenticatorDelegate(
         authenticatorState = authenticatorState,
+        // `fillMaxSize()` is needed, otherwise the prompts are displayed at the top of the screen.
         modifier = modifier.fillMaxSize(),
         onPendingOAuthUserSignIn = onPendingOAuthUserSignIn
     )
--- a/toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
+++ b/toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
@@ -24,7 +24,6 @@ import android.content.ContextWrapper
 import android.security.KeyChain
 import androidx.activity.compose.BackHandler
 import androidx.compose.foundation.layout.fillMaxSize
-import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.material3.AlertDialog
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.MaterialTheme
@@ -57,6 +56,26 @@ public fun Authenticator(
     Shareable(authenticatorState = authenticatorState, modifier = modifier, onPendingOAuthUserSignIn = onPendingOAuthUserSignIn)
 }
 
+
+/**
+ * Displays the [Authenticator] in an [AlertDialog]. See the Authenticator component for more details.
+ *
+ * @param authenticatorState an [AuthenticatorState]. See [AuthenticatorState.Companion.Factory].
+ * @param onPendingOAuthUserSignIn if not null, this will be called when an OAuth challenge is pending
+ * and the browser should be launched. Use this if you wish to handle OAuth challenges from your own
+ * activity rather than using the [OAuthUserSignInActivity].
+ * @since 200.2.0
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+public fun DialogAuthenticator(authenticatorState: AuthenticatorState, onPendingOAuthUserSignIn: ((OAuthUserSignIn) -> Unit)? = null) {
+    val showDialog = authenticatorState.isDisplayed.collectAsStateWithLifecycle(initialValue = false).value
+    if (showDialog) {
+        val modifier =  Modifier.clip(MaterialTheme.shapes.extraLarge)
+        Shareable(displayOnDialog = true, authenticatorState = authenticatorState, modifier = modifier, onPendingOAuthUserSignIn)
+    }
+}
+
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 private fun Shareable(
@@ -82,7 +101,7 @@ private fun Shareable(
     pendingServerTrustChallenge?.let {
         if (displayOnDialog) {
             AlertDialog(onDismissRequest = { /*TODO*/ }) {
-                ServerTrustAuthenticator(it, modifier)        
+                ServerTrustAuthenticator(it, modifier)
             }
         } else {
             ServerTrustAuthenticator(it, modifier)
@@ -126,23 +145,4 @@ private fun Context.findActivity(): Activity {
         context = context.baseContext
     }
     throw IllegalStateException("Could not find an activity from which to launch client certificate picker.")
-}
-
-/**
- * Displays the [Authenticator] in an [AlertDialog]. See the Authenticator component for more details.
- *
- * @param authenticatorState an [AuthenticatorState]. See [AuthenticatorState.Companion.Factory].
- * @param onPendingOAuthUserSignIn if not null, this will be called when an OAuth challenge is pending
- * and the browser should be launched. Use this if you wish to handle OAuth challenges from your own
- * activity rather than using the [OAuthUserSignInActivity].
- * @since 200.2.0
- */
-@OptIn(ExperimentalMaterial3Api::class)
-@Composable
-public fun DialogAuthenticator(authenticatorState: AuthenticatorState, onPendingOAuthUserSignIn: ((OAuthUserSignIn) -> Unit)? = null) {
-    val showDialog = authenticatorState.isDisplayed.collectAsStateWithLifecycle(initialValue = false).value
-    if (showDialog) {
-        val modifier =  Modifier.clip(MaterialTheme.shapes.extraLarge)
-        Shareable(displayOnDialog = true, authenticatorState = authenticatorState, modifier = modifier, onPendingOAuthUserSignIn)
-    }
 }
--- a/toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
+++ b/toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
@@ -24,6 +24,7 @@ import android.content.ContextWrapper
 import android.security.KeyChain
 import androidx.activity.compose.BackHandler
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.material3.AlertDialog
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.MaterialTheme
@@ -35,6 +36,7 @@ import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import com.arcgismaps.httpcore.authentication.ArcGISAuthenticationChallenge
 import com.arcgismaps.httpcore.authentication.OAuthUserConfiguration
 import com.arcgismaps.httpcore.authentication.OAuthUserSignIn
+
 /**
  * Displays appropriate Authentication UI when issued a challenge. For example, if an [ArcGISAuthenticationChallenge]
  * is issued and the [AuthenticatorState] has a corresponding [OAuthUserConfiguration],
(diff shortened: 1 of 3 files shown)


toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
```diff
@@ -51,8 +51,65 @@ import com.arcgismaps.httpcore.authentication.OAuthUserSignIn
 @Composable
 public fun Authenticator(
     authenticatorState: AuthenticatorState,
-    modifier: Modifier = Modifier.fillMaxSize(),
+    modifier: Modifier = Modifier,
     onPendingOAuthUserSignIn: ((OAuthUserSignIn) -> Unit)? = null
+) {
+    AuthenticatorImpl(
+        authenticatorState = authenticatorState,
+        modifier = modifier,
```
hud10837, line 59: We should do this to get fullscreen behavior
```suggestion
        modifier = modifier.fillMaxSize(),
```

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
```diff
@@ -51,8 +51,65 @@ import com.arcgismaps.httpcore.authentication.OAuthUserSignIn
 @Composable
 public fun Authenticator(
```
hud10837, line 52: This might be a good time to enhance the doc to specify that this displays full screen prompts and @see DialogAuthenticator

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
```diff
@@ -51,8 +51,65 @@ import com.arcgismaps.httpcore.authentication.OAuthUserSignIn
...
+ * @param onPendingOAuthUserSignIn if not null, this will be called when an OAuth challenge is pending
+ * and the browser should be launched. Use this if you wish to handle OAuth challenges from your own
+ * activity rather than using the [OAuthUserSignInActivity].
+ * @since 200.2.0
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+public fun DialogAuthenticator(
+    authenticatorState: AuthenticatorState,
+    onPendingOAuthUserSignIn: ((OAuthUserSignIn) -> Unit)? = null,
```
hud10837, line 77: I think as part of this PR we should add a modifier argument in the middle which should be there anyway, I discussed with @gunt0001 and this is not a breaking change because the user either has a trailing lambda for `onPendingOAuthUserSignIn` or a named parameter.

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
```diff
@@ -51,8 +51,65 @@ import com.arcgismaps.httpcore.authentication.OAuthUserSignIn
...
+}
+
+/**
+ * Listens for state changes and displays the appropriate authentication composable.
+ *
+ * @param authenticatorState an [AuthenticatorState]. See [AuthenticatorState.Companion.Factory].
+ * @param onPendingOAuthUserSignIn if not null, this will be called when an OAuth challenge is pending
+ * and the browser should be launched. Use this if you wish to handle OAuth challenges from your own
+ * activity rather than using the [OAuthUserSignInActivity].
+ * @param container if not null, the passed composable will be used as a container for [ServerTrustAuthenticator] and
```
hud10837, line 103: I know this is a private function, but I think we should be very thorough with the naming and doc here. Perhaps `promptContainer` to be explicit that this container will be used to contain authentication prompts. For the doc, I think we should explicitly mention that the caller must use the argument passed to the lambda in order for authentication prompts to display, otherwise it won't work.

We could also consider naming this something like `onShouldDisplayComposableAuthenticationPrompt`. This way it's explicit that this function is used to display prompts and it will be a bit more obvious what the parameter being passed to it is (ie. an authentication prompt).

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
```diff
@@ -51,8 +51,65 @@ import com.arcgismaps.httpcore.authentication.OAuthUserSignIn
...
+        AuthenticatorImpl(
+            authenticatorState = authenticatorState,
+            modifier = Modifier.fillMaxWidth(),
+            onPendingOAuthUserSignIn,
+        ) {
+            AlertDialog(
+                onDismissRequest = authenticatorState::dismissAll,
+                modifier = Modifier.clip(MaterialTheme.shapes.extraLarge),
+            ) {
+                it()
```
hud10837, line 90: I think we should use a named argument here, just to make it clear what's happening

hud10837 (changes requested):
I like this pattern, some initial thoughts

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
```diff
@@ -51,8 +51,65 @@ import com.arcgismaps.httpcore.authentication.OAuthUserSignIn
...
+ * @param authenticatorState an [AuthenticatorState]. See [AuthenticatorState.Companion.Factory].
+ * @param onPendingOAuthUserSignIn if not null, this will be called when an OAuth challenge is pending
+ * and the browser should be launched. Use this if you wish to handle OAuth challenges from your own
+ * activity rather than using the [OAuthUserSignInActivity].
+ * @param container if not null, the passed composable will be used as a container for [ServerTrustAuthenticator] and
+ * [UsernamePasswordAuthenticator].
+ * @since 200.4.0
+ */
+@Composable
+private fun AuthenticatorImpl(
```
eri9000, line 108: Alternative name: 
- AuthenticatorWithContainer
- ContainerAuthenticator

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
```diff
@@ -105,26 +193,3 @@ private fun Context.findActivity(): Activity {
...
-@OptIn(ExperimentalMaterial3Api::class)
-@Composable
-public fun DialogAuthenticator(authenticatorState: AuthenticatorState, onPendingOAuthUserSignIn: ((OAuthUserSignIn) -> Unit)? = null) {
-    val showDialog = authenticatorState.isDisplayed.collectAsStateWithLifecycle(initialValue = false).value
-    if (showDialog) {
-        AlertDialog(
-            onDismissRequest = { authenticatorState.dismissAll() },
-            modifier = Modifier.clip(MaterialTheme.shapes.extraLarge),
-        ) {
-            Authenticator(authenticatorState = authenticatorState, modifier = Modifier.fillMaxWidth(), onPendingOAuthUserSignIn)
```
eri9000, line 127: I removed `fillMaxWidth()` as it doesn't seem to change how the dialogs are displayed.

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
eri9000: I discussed this with Gunther and we both agreed that `container` describes the behavior. 
I renamed `AuthenticaorImpl` -> `AuthenticatorDelegate`. I am open to suggestions though.

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
```diff
@@ -38,21 +37,98 @@ import com.arcgismaps.httpcore.authentication.OAuthUserConfiguration
...
 @Composable
 public fun Authenticator(
     authenticatorState: AuthenticatorState,
-    modifier: Modifier = Modifier.fillMaxSize(),
+    modifier: Modifier = Modifier,
     onPendingOAuthUserSignIn: ((OAuthUserSignIn) -> Unit)? = null
+) {
+    AuthenticatorDelegate(
+        authenticatorState = authenticatorState,
+        modifier = modifier.fillMaxSize(),
```
eri9000, line 63: `fillMaxSize()` is needed here otherwise the prompts are displayed at the top of the screen

hud10837 (approved):
@eri9000 Works as expected on my device. I have some nitpick suggestions, but approving as they don't need re-review

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
```diff
@@ -38,21 +37,98 @@ import com.arcgismaps.httpcore.authentication.OAuthUserConfiguration
...
+        onPendingOAuthUserSignIn = onPendingOAuthUserSignIn
+    )
+}
+
+/**
+ * Displays appropriate Authentication UI when a challenge is issued.
```
gunt0001, line 69: nitpick
```suggestion
 * Displays appropriate Authentication UI when an authentication challenge is issued.
```

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
```diff
@@ -38,21 +37,98 @@ import com.arcgismaps.httpcore.authentication.OAuthUserConfiguration
...
+ *
+ * For example, when an [ArcGISAuthenticationChallenge] is issued and the [AuthenticatorState] has a corresponding
+ * [OAuthUserConfiguration], then a Custom Tab will be launched to complete the OAuth sign in.
+ *
+ * The [ServerTrustAuthenticator] and [UsernamePasswordAuthenticator] components will be displayed in an [AlertDialog].
```
gunt0001, line 74: `ServerTrustAuthenticator` & `UsernamePasswordAuthenticator` are internal types, so we cannot link to them in public doc. Instead I suggest:
```suggestion
 * Server trust prompts and username/password prompts will be displayed in an [AlertDialog].
```

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
```diff
@@ -38,21 +37,98 @@ import com.arcgismaps.httpcore.authentication.OAuthUserConfiguration
...
+ * All other components are displayed in full screen.
```
gunt0001, line 75:
```suggestion
 * All other prompts are displayed in full screen.
```

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
```diff
@@ -38,21 +37,98 @@ import com.arcgismaps.httpcore.authentication.OAuthUserConfiguration
...
+ *
+ * For alternate behavior, see the [Authenticator] component.
+ *
+ * @param authenticatorState the application viewModel's [AuthenticatorState].
```
gunt0001, line 79:
```suggestion
 * @param authenticatorState the object that holds the state to handle authentication challenges.
```

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
```diff
@@ -38,21 +37,98 @@ import com.arcgismaps.httpcore.authentication.OAuthUserConfiguration
...
+ * @param modifier the [Modifier] to be applied to this DialogAuthenticator.
+ * @param onPendingOAuthUserSignIn if not null, this will be called when an OAuth challenge is pending
+ * and the browser should be launched. Use this if you wish to handle OAuth challenges from your own
+ * activity rather than using the [OAuthUserSignInActivity].
```
gunt0001, line 83:
```suggestion
 * activity rather than using the [OAuthUserSignInActivity].
 * @see Authenticator
```

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/Authenticator.kt
```diff
@@ -38,21 +37,98 @@ import com.arcgismaps.httpcore.authentication.OAuthUserConfiguration
...
+
+/**
+ * Listens for [AuthenticatorState] changes and displays the corresponding authentication component on the screen.
+ *
+ * If a different container is desired for [ServerTrustAuthenticator] and [UsernamePasswordAuthenticator], then it
+ * should be defined in the [container] lambda. Additionally, the argument should be invoked inside this container
+ * otherwise this will not work.
+ *
+ * @sample [DialogAuthenticator]
+ * @param authenticatorState the application viewModel's [AuthenticatorState].
```
gunt0001, line 118:
```suggestion
 * @param authenticatorState the object that holds the state to handle authentication challenges.
```

gunt0001 (commented):
@eri9000 - great work. Some doc related comments.

gunt0001 (approved):
@eri9000 - looks good 👍

PR #225, opened 2023-12-06
`Forms`: Combo Box Dialog sizing

kaushikrw:


sorenoid (approved):
LGTM. just one suggestion that is more of an open question. approving.

sorenoid (approved):

PR #226, opened 2023-12-06
Rename MapView-related microapps to have MapView prefix

hud10837:
- Deletes the unnecessary `map-compose-app`
- Renames the location display app and geometry editor apps to have the MapView prefix, including all package names, strings etc.


puneet-pdx (approved):

shubham7109 (approved):
I see a warning _"Package directive does not match the file location"_ on the package declarations for the `ExampleInstrumentedTest.kt` & `ExampleUnitTest.kt` for both the microapps.  

Should we remove these files altogether? They don't really serve any purpose since we don't need testing for microapps. 

The rest of the PR changes looks good 👍

hud10837:
Thanks @shubham7109 @puneet-pdx I have renamed those files. I think we can consider deleting them as a second PR

PR #229, opened 2023-12-08
add visible area

puneet-pdx:
changes for supporting multiple listerners

### DIFF
--- a/toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
+++ b/toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
@@ -35,6 +35,8 @@ import com.arcgismaps.ArcGISEnvironment
 import com.arcgismaps.geometry.SpatialReference
 import com.arcgismaps.mapping.ArcGISMap
 import com.arcgismaps.mapping.TimeExtent
+import com.arcgismaps.mapping.Viewpoint
+import com.arcgismaps.mapping.ViewpointType
 import com.arcgismaps.mapping.view.BackgroundGrid
 import com.arcgismaps.mapping.view.DoubleTapEvent
 import com.arcgismaps.mapping.view.DownEvent
@@ -102,6 +104,7 @@ public fun MapView(
     arcGISMap: ArcGISMap? = null,
     viewpointOperation: MapViewpointOperation? = null,
     onViewpointChanged: (() -> Unit)? = null,
+    viewpointChangedEvents: Map<ViewpointType, ((Viewpoint) -> Unit)>? = null,
     graphicsOverlays: GraphicsOverlayCollection = rememberGraphicsOverlayCollection(),
     locationDisplay: LocationDisplay = rememberLocationDisplay(),
     geometryEditor: GeometryEditor? = null,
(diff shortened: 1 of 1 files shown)


toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -223,6 +227,28 @@ private fun ViewpointUpdater(
...
+                    it.value.invoke(
+                        mapView.getCurrentViewpoint(
+                            it.key
+                        ) ?: throw IllegalStateException("Viewpoint should not be null")
+                    )
+                }
+            }
+        }
+    }
+}
```
gunt0001, line 250: You should do a null check on `viewpointChangedEvents` before launching a coroutine. If `viewpointChangedEvents` there is no point in launching a coroutine. Also, I don't think there is a need for launchin a separate coroutine here, since you only want to collect a single event in this LaunchedEffect:
```suggestion
private fun ViewpointChangedEventHandler(mapView: MapView, viewpointChangedEvents: Map<ViewpointType, ((Viewpoint) -> Unit)>?) {
   viewpointChangedEvents?.let { lambdas ->
      mapView.viewpointChanged.collect {
         lambdas.foreach {
            it.value.invoke(
                        mapView.getCurrentViewpoint(
                            it.key
                        ) ?: throw IllegalStateException("Viewpoint should not be null")
                    )
         }
      }
   }
}
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -223,6 +227,28 @@ private fun ViewpointUpdater(
     }
 }
 
+/**
+ * Sets up the ViewpointChangedState's property and event.
```
gunt0001, line 231:
```suggestion
 * Sets up the ViewPointChanged events.
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -450,3 +476,13 @@ public inline fun rememberGraphicsOverlayCollection(
...
 }
+
+@Composable
+public fun rememberViewpointChangedEvents(
+    type: ViewpointType = ViewpointType.BoundingGeometry,
+    onViewpointChanged: ((Viewpoint) -> Unit)
+): Map<ViewpointType, ((Viewpoint) -> Unit)> {
+    return hashMapOf(
+        type to onViewpointChanged
+    )
```
gunt0001, line 487: You will need to wrap this into a `remember`. Also good to pass a `key` parameter. See [rememberLocationDisplay](https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/blob/feature-branches/geo-compose/toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt#L425-L435) for an example.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -102,6 +104,7 @@ public fun MapView(
     arcGISMap: ArcGISMap? = null,
     viewpointOperation: MapViewpointOperation? = null,
     onViewpointChanged: (() -> Unit)? = null,
+    viewpointChangedEvents: Map<ViewpointType, ((Viewpoint) -> Unit)>? = null,
```
gunt0001, line 107: need to add an `@param` in the doc

PR #231, opened 2023-12-11
add isWrapAroundEnabled property to GeoViewProxy

puneet-pdx:


gunt0001 (approved):

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/GeoViewProxy.kt
```diff
@@ -54,6 +54,14 @@ public sealed class GeoViewProxy(className: String) {
     private val nullGeoViewErrorMessage: String =
         "$className must be part of the composition when this member is called."
 
+    /**
+     * True if continuous panning across the international date line is enabled in the GeoView, false otherwise.
+     *
```
hud10837, line 59: Could you add an `@return` here to explain why it may be null?

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/GeoViewProxy.kt
```diff
@@ -54,6 +54,14 @@ public sealed class GeoViewProxy(className: String) {
...
+     * @since 200.4.0
+     */
+    public val isWrapAroundEnabled: Boolean?
+        get() = geoView?.isWrapAroundEnabled
```
gunt0001, line 63: The return value should not be nullable. Since properties should not throw, it makes sense to return `false` in the case of `geoView` being null: 
```suggestion
    public val isWrapAroundEnabled: Boolean
        get() = geoView?.isWrapAroundEnabled ?: false
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/GeoViewProxy.kt
puneet-pdx: @hud10837 `@return` tag applies to functions only. I have added the explanation in the description.

gunt0001 (approved):
LGTM 👍

hud10837 (approved):
LGTM 👍

PR #234, opened 2023-12-12
add ViewPointState

puneet-pdx:


toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -102,6 +103,7 @@ public fun MapView(
     arcGISMap: ArcGISMap? = null,
     viewpointOperation: MapViewpointOperation? = null,
     onViewpointChanged: (() -> Unit)? = null,
```
gunt0001, line 105: This parameter (and corresponding doc) will need to be removed, since `viewpointChangedState` replaces it.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -223,6 +226,23 @@ private fun ViewpointUpdater(
     }
 }
 
+/**
+ * Sets up the ViewpointChangedState's property and event.
+ *
+ * @since 200.4.0
+ */
+@Composable
+private fun ViewpointChangedStateHandler(
```
gunt0001, line 235: Instead of a separate composable function, could the existing `MapViewEventHandler()` function handle this?

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/ViewpointChangeState.kt
```diff
@@ -0,0 +1,105 @@
+/*
```
gunt0001, line 1: The file name doesn't quite match - needs to be `ViewpointChangedState.kt` instead of `ViewpointChangeState.kt` (missing "d")

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/ViewpointChangeState.kt
```diff
@@ -0,0 +1,105 @@
...
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.Stable
+import com.arcgismaps.mapping.Viewpoint
+import com.arcgismaps.mapping.ViewpointType
+import com.arcgismaps.mapping.view.MapView
+import androidx.compose.runtime.remember
+
+/**
+ * State holder for a lambdas invoked when the viewpoint of a composable [MapView] has changed.
```
gunt0001, line 27:
```suggestion
 * State holder for lambdas invoked when the viewpoint of a composable [MapView] has changed.
 * Depending on the [ViewpointType] you need to be notified about, you can create a ViewpointChangedState instance to be notified
 * about Viewpoint changes for viewpoints of type [ViewpointType.CenterAndScale] or [ViewpointType.BoundingGeometry] or both. Use the provided factory functions [rememberViewpointChangedStateForCenterAndScale], [rememberViewpointChangedStateForBoundingGeometry] or [rememberViewpointChangedState] to create the appropriate ViewpointChangedState.
 
@see rememberViewpointChangedStateForCenterAndScale
@see rememberViewpointChangedStateForBoundingGeometry
@see rememberViewpointChangedState
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/ViewpointChangeState.kt
```diff
@@ -0,0 +1,105 @@
...
+    onViewpointChanged: (Viewpoint) -> Unit
+): ViewpointChangedState = remember(key) {
+    ViewpointChangedState(
+        null,
+        onViewpointChanged
+    )
+}
+
+/**
+ * Create and [remember] a [ViewpointChangedState] with the lambdas that returns the viewpoint of the
```
gunt0001, line 75:
```suggestion
 * Create and [remember] a [ViewpointChangedState] with the lambdas that return the viewpoint of the
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/ViewpointChangeState.kt
```diff
@@ -0,0 +1,105 @@
...
+    key: Any? = null,
+    onViewpointChangedForCenterAndScale: (Viewpoint) -> Unit,
+    onViewpointChangedForBoundingGeometry: (Viewpoint) -> Unit): ViewpointChangedState = remember(key) {
+    ViewpointChangedState(
+        onViewpointChangedForCenterAndScale,
+        onViewpointChangedForBoundingGeometry
+    )
+}
+
+internal fun ViewpointChangedState.invoke(mapView: MapView) {
```
gunt0001, line 94: Could you move this function into `MapView.kt` and make it private. It's only used there.

Missing doc.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -223,6 +226,23 @@ private fun ViewpointUpdater(
...
+    mapView: MapView,
+    viewpointChangedState: ViewpointChangedState?
+) {
+    LaunchedEffect(viewpointChangedState) {
+        mapView.viewpointChanged.collect {
```
gunt0001, line 240: As mentioned previously, there is no need to collect `mapView.viewpointChanged` events if `viewpointChangedState` is null.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
puneet-pdx: I had it in a separate composable function as I thought it would make sense in how we are handling the state objects. We have the attributionState in a separate composable function too.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/ViewpointChangeState.kt
puneet-pdx: updated

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
puneet-pdx: updated

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
puneet-pdx: I have made the change. But we don't do this for any of the other callback methods in `MapViewEventHandler()` or in `AttributionStateHandler()`. Should we be doing this everywhere else too ?

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
gunt0001: Ok, I think one could argue for either way. One argument could be that viewpoint changed events are MapView events, thus should be handled by `MapViewEventHandler()`. On the other hand, it's a state object we are dealing with (`ViewpointChangedState`), and as such we should handle it in a similar way as we do for `AttributionState`. I'm fine leaving it as is, i.e. a separate `ViewpointChangedStateHandler`.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -296,7 +320,10 @@ private fun MapViewEventHandler(
         }
         launch {
             mapView.viewpointChanged.collect {
-                currentViewPointChanged?.invoke()
+                currentVisibleAreaChanged?.invoke(
+                    mapView.visibleArea
+                        ?: throw IllegalStateException("Mapview visible area should not be null")
```
gunt0001, line 325:
```suggestion
                        ?: throw IllegalStateException("MapView visible area should not be null")
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -223,6 +228,25 @@ private fun ViewpointUpdater(
     }
 }
 
+/**
+ * Sets up the ViewpointChangedState's property and event.
```
gunt0001, line 232:
```suggestion
 * Invokes the ViewpointChangedState's lambdas when a [MapView.viewpointChanged] event is collected.
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/ViewpointChangedState.kt
```diff
@@ -0,0 +1,100 @@
...
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.Stable
+import com.arcgismaps.mapping.Viewpoint
+import com.arcgismaps.mapping.ViewpointType
+import com.arcgismaps.mapping.view.MapView
+import androidx.compose.runtime.remember
+
+/**
+ * State holder for lambdas invoked when the viewpoint of a composable [MapView] has changed.
```
gunt0001, line 27: I think this [MapView] link points to the view-based MapView instead of the composable MapView.

gunt0001 (approved):
@puneet-pdx - looks good, some minor comments. 

> I have made the change. But we don't do this for any of the other callback methods in MapViewEventHandler() or in AttributionStateHandler(). Should we be doing this everywhere else too ?

It's different for `MapViewEventHandler()` where the `LaunchedEffect` is never re-composed (due to taking a key of `Unit`). So we have to launch each coroutine right away, independent of whether the callback is null or not. However, maybe we should review this and see if there is a more efficient way to do this.

For `AttributionStateHandler`, yes, we should revisit this and make it consistent with `ViewpointChangedStateHandler`. Could you create a separate issue for it.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -296,7 +320,10 @@ private fun MapViewEventHandler(
         }
         launch {
             mapView.viewpointChanged.collect {
-                currentViewPointChanged?.invoke()
+                currentVisibleAreaChanged?.invoke(
+                    mapView.visibleArea
+                        ?: throw IllegalStateException("MapView visible area should not be null")
```
hud10837, line 325: What is the reason for this not to be nullable anymore?

gunt0001: @hud10837 - when we collect a `mapView.viewpointChanged` event from the view-based MapView, `mapView.visibleArea` should never be null. So there is no point in exposing a nullable parameter on the `currentVisibleAreaChanged` callback.

hud10837 (approved):
LGTM 👍

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
hud10837: Okay, I won't hold this up.

PR #239, opened 2023-12-13
`Forms`: Microapp optimizations

kaushikrw:
### Summary of changes

- Portal items and their thumbnail are now loaded in parallel reducing overall loading times on average from ~40s to ~10s (first load). 
- Added a new composable `AsyncImage` which uses a model class `ImageLoader` to load images asynchronously.
    - Thumbnails are now loaded directly using `PortalItem.thumbnail` and `AsyncImage` instead of using Coil to load images stored on disk.
    - Coil cannot be used to directly load the thumbnails from its urls when the PortalItem is private because it does have the authentication credentials to access those resources.
    - This is similar to the `AsyncImageView` of the swift toolkit [component](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/main/Sources/ArcGISToolkit/Utility/AsyncImageView.swift).
    - One minor downside to this is the thumbnails are not loaded when the device is offline. But this could be fixed by adding a local cache layer to store thumbnails.
- Removed the option of a "clear cache". By default the local cache will be deleted when a "refresh" is performed.
- Cleaned up and optimized some code with added comments.
- Removed Coil dependency.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -14,7 +14,6 @@ androidxMaterialIcons = "1.4.3"
 androidxTestExt = "1.1.2"
 androidxViewmodelCompose = "2.6.1"
 androidxWindow = "1.2.0"
-coil = "2.4.0"
 compileSdk = "34"
 compose-material3 = "1.1.0"
 compose-navigation = "2.7.5"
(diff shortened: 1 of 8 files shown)


sorenoid (approved):
looks great and works for me!

kaushikrw:
Thanks @sorenoid, merging!

PR #244, opened 2023-12-14
settings.gradle refactor

sorenoid:
remove list of projects, just individually include each project since most projects were being included that way.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -16,14 +16,6 @@
  *
  */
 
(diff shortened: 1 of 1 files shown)


kaushikrw (commented):
I think the `new-component-starter.sh` script needs to be updated as well to include the project into this settings.gradle file.

sorenoid:
well, I don't have a solution to make the new project starter work if this is merged, so I am closing for now, and will add some comments to make things more clear what is going on.

PR #247, opened 2023-12-18
SceneView: attribution

hud10837:
Adds `AttributionState` parameter to the SceneView:

- Moves `AttributionStateHandler` to `AttributionState.kt`
- Adds the param to the SceneView


gunt0001 (approved):
👍

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/AttributionState.kt
```diff
@@ -32,3 +36,26 @@ public data class AttributionState(
     val onAttributionBarLayoutChanged: ((AttributionBarLayoutChangeEvent) -> Unit)? = null
 )
 
+/**
+ * Sets up the attribution bar's property and events.
+ */
```
changanxian, line 41: missing @since tag

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/AttributionState.kt
```diff
@@ -32,3 +36,26 @@ public data class AttributionState(
...
+@Composable
+internal fun AttributionStateHandler(geoView: GeoView, attributionState: AttributionState) {
+    LaunchedEffect(attributionState) {
+        // isAttributionBarVisible does not take effect if applied in the AndroidView update callback
+        geoView.isAttributionBarVisible = attributionState.isAttributionBarVisible
```
changanxian, line 46: should this line be outside the `LaunchedEffect` function?

gunt0001: It probably could be outside of the LaunchedEffect, but I don't think it matters in this case. The LaunchedEffect will be recomposed with every change of `attributionState`, and then we simply apply all the AttributionState proeprties to the GeoView "in one go".

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/AttributionState.kt
gunt0001: updated

gunt0001 (commented):
@changanxian - please take another look

changanxian (approved):
👍

PR #249, opened 2023-12-18
SceneView: add ViewLabelProperties

hud10837:
- Adds `viewLabelProperties` to the composable SceneView.

This is easier test with the following scene, as disabling labels only applies to layers that are added to the scene (so the basemap layer won't be affected unless it is added separately)

```kotlin
val arcGISScene = remember { 
            ArcGISScene().apply {
            val layer = ArcGISTiledLayer("https://www.arcgis.com/home/item.html?id=10df2279f9684e4a9f6a7f08febac2a9")
            operationalLayers.add(layer)
        } 
    }
```


gunt0001 (approved):

changanxian (approved):
LGTM

PR #250, opened 2023-12-19
`Forms`: Remove FeatureFormState

kaushikrw:
### Summary of changes

- Removed `FeatureFormState` and changed the `@Composable FeatureForm` to accept the `FeatureForm` type directly.
- The logic of managing the editing state is moved to the caller by introducing the `UIState` type in the `MapViewModel`.
- `UIState.Loading` can be used for showing progress if the `commitEdits()` operation is taking too long.


microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapScreen.kt
```diff
@@ -147,10 +120,12 @@ fun MapScreen(mapViewModel: MapViewModel = hiltViewModel(), onBackPressed: () ->
                     sheetWidth = with(LocalDensity.current) { layoutWidth.toDp() }
                 ) {
                     // set bottom sheet content to the FeatureForm
-                    FeatureForm(
-                        featureFormState = mapViewModel,
-                        modifier = Modifier.fillMaxSize()
-                    )
+                    if (uiState is UIState.Editing) {
+                        FeatureForm(
+                            featureForm = (uiState as UIState.Editing).featureForm,
```
sorenoid, line 125: is this cast necessary?

sorenoid (approved):
this looks good. Just one small question, approving.

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapScreen.kt
kaushikrw: Smart casting doesn't work if `uiState` is read using the `by` delegate. 

Otherwise, does this read better?

```kotlin
val uiState = mapViewModel.uiState.value
...
featureForm = uiState.featureForm
```

sorenoid: probably not worth changing, just caught my eye. thanks for the explanation.

kaushikrw:
@sorenoid thanks! merging now.

PR #252, opened 2023-12-20
`Forms`: Common base state

kaushikrw:
### Summary of changes

Introduced a new `FormElementState` which is an abstract class that forms the state object for a `FormElement` type. Since in the SDK all the element types inherit from a `FormElement` this state class acts as a base class for any other derived `FormElement` state objects. This then creates the following inheritance heirarchy
   - FormElementState 
       - BaseFieldState
           - FormTextField  
           - .. other field element type states
       - BaseGroupState
       - RelationshipElement state (future)
       - AttachmentElement state (future)


sorenoid (approved):
looks great, just like we discussed. thanks!

kaushikrw:
thanks @sorenoid, merging!

PR #254, opened 2023-12-21
`Forms`: Refactored Dialog Mechanism

kaushikrw:
### Summary of changes

Depends on #252 to be merged.
- Uses a `DialogRequester` class that handles dialog requests. An instance of `DialogRequester` is injected into the hierarchy using `CompositionLocal` as `LocalDialogRequester` which is internal.
- Any FormElements that need to request for a dialog accesses the `LocalDialogRequester` and call *requestDialog()*. 
- At the root level a `FeatureFormDialog` can then listen to dialog requests via the `LocalDialogRequester` and display the appropriate one.
- Thus this mechanism abstracts the inner workings of requesting and displaying dialogs from the end user.


sorenoid (approved):
LGTM and works as expected.

PR #264, opened 2024-01-02
The mapView property needs to be assigned specifically in MapViewProxy class

changanxian:
The mapView property needs to be assigned in the mapViewProxy class
The sceneView property needs to be assigned in the sceneViewProxy class


hud10837 (approved):
Thanks for fixing this!

puneet-pdx (approved):

PR #265, opened 2024-01-03
SceneView: add CameraController param

hud10837:
Adds the `cameraController` parameter.

Note that this interacts with the viewpoint set on the SceneView via `SceneViewpointOperation`, but it works the same as `setViewpoint` methods on the view-based SceneView. I considered adding extra doc to explain what will happen, but I think the doc on `OrbitLocationCameraController` and `OrbitGeoElementCameraController` do a good enough job of explaining that setting the viewpoint when they are active won't work, and I confirmed that that's the case here with ad hoc tests.


puneet-pdx (approved):
@hud10837 LGTM 👍🏽

changanxian (approved):
👍

PR #267, opened 2024-01-03
validation via the SDK

sorenoid:
This PR pulls in and uses validation errors from the SDK. It does not implement the full error propagation design, just a minimal usage of the SDK to provide the same validation feedback in the UI.

* add a `validate` lambda property to BaseFieldState, passed as a parameter with a default no-op value. 
* the `validate` lambda returns a list of Throwables, and simply calls `FieldFormElement.validate()`.
    * these throwables are mapped to the existing error toolkit hierarchy (`ErrorValidationState`), which provides a bit more granularity for error states.


kaushikrw (approved):
LGTM! Just a minor comment.

PR #268, opened 2024-01-03
Update OAuthUserSignInActivity launch mode to singleTask. 

eri9000:
Fixes issues with FireFox not redirecting back to `OAuthUserSignInActivity` by changing the launch mode from `singleTop` to `singleTask`

Here's the doc on `singleTask`:
> The system creates the activity at the root of a new task or locates the activity on an existing task with the same affinity. If an instance of the activity already exists, the system routes the intent to the existing instance through a call to its [onNewIntent()](https://developer.android.com/reference/android/app/Activity#onNewIntent(android.content.Intent)) method, rather than creating a new instance. Meanwhile all of the other activities on top of it are destroyed.

Since Activities launched in the same app have affinity to each other, the `OAuthUserSignInActivity` will not create a new task and instead use the same task where it is launched.


hud10837 (approved):
Thanks for investigating this issue. I couldn't find any problems with the recent task list when using `singleTask`.

PR #276, opened 2024-01-08
`Forms`: Refactor Validation

kaushikrw:
### Summary of changes

The changes are based on the new validation design.

- BaseFieldState introduces two new methods to facilitate the design.
  - open fun validate() : List< ValidationErrorState>
      - This method uses the default validation logic `field::getValidationErrors` to generate a list of `ValidationErrorState` for the compose UI to consume. Child classes can override this to provide a more concise list of errors according to their data type.
  - private fun filter(errors : List< ValidationErrorState>) : ValidationErrorState : 
      - This method takes all the errors for a field value and only produces one validation error based on the "Field Validation UI Messaging Algorithm". Since this algorithm is the single source of truth, this method is final.

- Additionally, the value flows are broken down into `_value` (provides user input), `_mergedValue`(combines _value and onValueChanged events) and finally the actual `value` as a `Value(val data: T, val error: ValidationErrorState)` object. This separation of data is needed to accurately model the data and the error into a single object. `_validationError` provides the errors in this case.

- To preserve validation errors when the device orientation changes, the `wasFocused` property is included in the Savers of all the state classes. Then when the state classes are restored the focus flag is also restored and hence any validation errors will be once again visible.

- Only the `FormTextField`, `DateTimeField` and `ComboBoxField` are effected in terms of validation since the other input types do not show any validation errors.

- The `FeatureFormDialog` now clears focus from the tapped field since losing focus is needed to display the "required" errors.

- The `FormTextFieldState` does not need a reference to a `Context` anymore since fetching resource strings are now part of the `ValidationErrorState` class which provides a compose accessible method `getString()`.

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseFieldState.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseFieldState.kt
@@ -16,7 +16,6 @@
 
 package com.arcgismaps.toolkit.featureforms.components.base
 
-import android.util.Log
 import com.arcgismaps.exceptions.FeatureFormValidationException
 import com.arcgismaps.mapping.featureforms.FieldFormElement
 import kotlinx.coroutines.CoroutineScope
@@ -42,6 +41,9 @@ internal open class FieldProperties<T>(
     val visible: StateFlow<Boolean>
 )
 
+/**
+ * A class that provides a validation error [error] for the value of [data].
+ */
 internal data class Value<T>(
     val data: T,
     val error: ValidationErrorState = ValidationErrorState.NoError
@@ -74,12 +76,16 @@ internal open class BaseFieldState<T>(
      */
     open val placeholder: String = properties.placeholder
 
-    // a state flow to handle user input changes
+    /**
+     * A mutable state flow to handle user input changes.
+     */
     protected val _value = MutableStateFlow(initialValue)
 
-
+    /**
+     * A state flow that combines the user input [_value] and calculated property callbacks.
+     */
     @OptIn(ExperimentalCoroutinesApi::class)
-    private val _mergedValue: StateFlow<T> = flowOf(_value, properties.value.drop(1))
+    protected val _mergedValue: StateFlow<T> = flowOf(_value, properties.value.drop(1))
         .flattenMerge()
         .stateIn(
             scope = scope,
@@ -87,13 +93,13 @@ internal open class BaseFieldState<T>(
             initialValue = initialValue
         )
 
+    // a state flow for sending validation errors
     private val _validationError: MutableStateFlow<ValidationErrorState> =
         MutableStateFlow(ValidationErrorState.NoError)
 
     /**
-     * Current value state for the field.
-     *
-     * ---validation behavior---
+     * Current value for this field state. The actual data of this type is wrapped in a [Value]
+     * object. The [Value.error] provides the current validation error for the [Value.data].
      */
     val value: StateFlow<Value<T>> = combine(_mergedValue, _validationError) { newValue, error ->
         Value(newValue, error)
@@ -113,7 +119,13 @@ internal open class BaseFieldState<T>(
      */
     val isRequired: StateFlow<Boolean> = properties.required
 
+
     private val _isFocused: MutableStateFlow<Boolean> = MutableStateFlow(false)
+
+    /**
+     * Property that indicates if the field for this state currently has focus. Use [onFocusChanged]
+     * to set this value.
+     */
     val isFocused: StateFlow<Boolean> = _isFocused.asStateFlow()
 
     protected var wasFocused = false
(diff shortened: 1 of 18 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseFieldState.kt
```diff
@@ -64,16 +78,38 @@ internal open class BaseFieldState<T>(
...
+
+    // a state flow for sending validation errors
+    private val _validationError: MutableStateFlow<ValidationErrorState> =
+        MutableStateFlow(ValidationErrorState.NoError)
+
+    /**
+     * Current value for this field state. The actual data of this type is wrapped in a [Value]
+     * object. The [Value.error] provides the current validation error for the [Value.data].
+     */
+    val value: StateFlow<Value<T>> = combine(_mergedValue, _validationError) { newValue, error ->
```
sorenoid, line 106: Based on my understanding of `combine` I would expect to get 2 emissions of this flow every time the `_mergedValue` changes, once when the value actually changes, and again when the validation error changes after calling `filter(validate(newValue))`. The first emission would have an out of date error, and would be incorrect. But that doesn't seem to be happening. I suspect though that the reason it isn't happening is because of the conflated nature of StateFlows, which makes it possible for a collector to miss values if the value changes before the collector is invoked. Placing a 1 second delay above the call to `updateValidation` at line 161 bears this out. So this is only working correctly because of timing.

I think what you want is a flow which emits a `Value`  after the filter method is complete. I am not sure what the best way to do this is. For starters this could become 

```kotlin
val value: StateFlow<Value<T>> = _mergedValue.drop(1).map {
    Value(it, filter(updateValidation(_mergedValue.value))
}.stateIn(
        scope = scope,
        started = SharingStarted.Eagerly,
        initialValue = Value(_mergedValue.value)
    )
```

but this would cut off focus, isEditable, isRequired from updating the UI. One way to solve that is to have all of them emit into the value flow, using concurrent merging of a flow for each. This seems to work but seems overly complicated.

```kotlin
    val _valueFromFocusChange: Flow<Value<T>> = isFocused.drop(1).map {
        Value(_mergedValue.value, filter(validate()))
    }
    
    val _valueFromIsEditableChange: Flow<Value<T>> = isEditable.drop(1).map {
        Value(_mergedValue.value, filter(validate()))
    }
    
    val _valueFromIsRequiredChange: Flow<Value<T>> = isRequired.drop(1).map {
        Value(_mergedValue.value, filter(validate()))
    }
    
    val _valueFromValueChange: Flow<Value<T>> = _mergedValue.drop(1).map {
        Value(it, filter(validate()))
    }
    
    val value: StateFlow<Value<T>> =
        flowOf(
            _valueFromFocusChange,
            _valueFromIsEditableChange,
            _valueFromIsRequiredChange,
            _valueFromValueChange
        ).flattenMerge(concurrency = 4)
            .stateIn(
                scope = scope,
                started = SharingStarted.Eagerly,
                initialValue = Value(initialValue)
            )
  ```

File: toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseFieldState.kt [hunk 2]
```diff
@@ -84,12 +120,129 @@ internal open class BaseFieldState<T>(
...
+                updateValidation()
+            }
+        }
+        scope.launch {
+            // validate when the value changes
+            _mergedValue.drop(1).collect {
+                updateValidation()
+            }
+        }
+    }
```
sorenoid, line 164: this would all go away if the concurrent merging suggested above is used.

sorenoid (changes requested):
This looks great but there is a small timing issue between the emission of values and emission of errors.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseFieldState.kt
```diff
@@ -64,16 +78,38 @@ internal open class BaseFieldState<T>(
...
         .flattenMerge()
-        .stateIn(scope, SharingStarted.Eagerly, initialValue)
+        .stateIn(
+            scope = scope,
+            started = SharingStarted.Eagerly,
+            initialValue = initialValue
+        )
+
+    // a state flow for sending validation errors
+    private val _validationError: MutableStateFlow<ValidationErrorState> =
```
sorenoid, line 99: `_validationError` seems a bit misleading since it provides a singular error to display in the UI, same with the property in the `Value` class. Can you please add some doc to this effect, and if you can think of an appropriate name that describes its role please consider changing the name here and in the `Value` class. I can't think of a great name, but something like `_errorToDisplay`

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseFieldState.kt
```diff
@@ -64,16 +78,38 @@ internal open class BaseFieldState<T>(
...
+     */
     protected val _value = MutableStateFlow(initialValue)
 
     /**
-     * Current value state for the field.
+     * A state flow that combines the user input [_value] and calculated property callbacks.
      */
     @OptIn(ExperimentalCoroutinesApi::class)
-    val value: StateFlow<T> = flowOf(_value, properties.value.drop(1))
+    protected val _mergedValue: StateFlow<T> = flowOf(_value, properties.value)
```
sorenoid, line 90:
```suggestion
    @Suppress("PropertyName")
    protected val _mergedValue: StateFlow<T> = flowOf(_value, properties.value)
```

sorenoid: there are a few more protected backing properties that could use this annotation, but they aren't part of this diff, such as `_value`.

sorenoid (approved):
after discussion I resolved my suggestion based on the synchronous and asynchronous nature of input and error validation.

I have some minor cleanup suggestions. happy to approve!

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseFieldState.kt
kaushikrw: as discussed this will be going away in #287

PR #277, opened 2024-01-08
`Forms:` date time UI tests

sorenoid:
* test cases 2.1, 2.2, 2.3, 2.6 from the toolkit common design
* 2.4, 2.5 require semantic matching of the calendar dates within the date picker, which may not be possible.

data seems to have changed, I will change the common toolkit test design to match what I found writing these tests. 

* new helper methods as needed.
* new screenshot capture method which is  not used but are helpful while developing a tests and it fails.


toolkit/featureforms/src/androidTest/java/com/arcgismaps/toolkit/featureforms/DateTimeFieldTests.kt
```diff
@@ -0,0 +1,260 @@
...
+        val dateTimeField = composeTestRule.onNodeWithText("${formElement.label} *")
+        val iconMatcher = hasContentDescription("field icon")
+        assert(iconMatcher.matches(dateTimeField.fetchSemanticsNode()))
+        dateTimeField.assertHasClickAction()
+        dateTimeField.performClick()
+    
+        val dialogSurface =
+            composeTestRule.onNodeWithContentDescription("DateTimePickerDialogSurface", useUnmergedTree = true)
+        dialogSurface.assertIsDisplayed()
+        saveScreenshot("xx",  dialogSurface.captureToImage().asAndroidBitmap())
```
kaushikrw, line 149:
```suggestion
        saveScreenshot("testRequiredFocusedValue",  dialogSurface.captureToImage().asAndroidBitmap())
```
It might be easy look up the test results if saved with the test name.

kaushikrw (commented):
Looks good!

Just a small question - was the plan to add saving screenshots to all of the DateTime tests?

PR #287, opened 2024-01-11
`Forms`: Refactored state flow based Impl into compose state based

kaushikrw:
### Summary of changes

- Updated the way values are passed to compose which now uses a compose `State` instead of a `StateFlow`. This is so that the transformation of input data via unidirectional flow from the UI -> State Object -> UI (recomposition) happens synchronously. 
- Calculated value changes as a result of `onValueChanged` callbacks are still routed through a `StateFlow` into the `State`.
- Observing the `required`, `editable` and `focus` changed flows are now part of the `observeProperties` method and hence needs an explicit invoke. This is so that any open functions like `validate()` can be invoked after the proper base-derived classes initialization order.
- `BaseFieldState` is made abstract since the behavior is defined by any derived classes and a `BaseFieldState` should never be instantiated by itself.
- Fixed the failing `FormTextFieldTexts` where the state was not being remembered through recompositions.

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseFieldState.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseFieldState.kt
@@ -128,6 +128,17 @@ internal abstract class BaseFieldState<T>(
      */
     private var isObserving = AtomicBoolean(false)
 
+    init {
+        // start listening to calculated value updates immediately
+        scope.launch {
+            // update the current value when the calculated value changes
+            // calculated properties do not get validated
+            _calculatedValue.collect {
+                _value.value = Value(it)
+            }
+        }
+    }
+
     /**
      * Changes the current focus state for the field. Use [isFocused] to read the value.
      */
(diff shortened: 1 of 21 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/codedvalue/CodedValueFieldState.kt
```diff
@@ -99,6 +100,11 @@ internal open class CodedValueFieldState(
      */
     val fieldType: FieldType = properties.fieldType
 
+    init {
+        // start observing the properties
+        observeProperties()
+    }
+
```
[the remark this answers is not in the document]
kaushikrw: Since it should _not_ be called from the base class init block, I've created a new `ComboBoxFieldState` that provides this concrete implementation while `CodedValueFieldState` is now abstract.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseFieldState.kt
```diff
@@ -148,7 +147,8 @@ internal abstract class BaseFieldState<T>(
     }
 
     /**
-     * Runs and updates the validation using [validate] and [filterErrors].
+     * Runs and updates the validation using [validate] and [filterErrors]. Avoid calling this
+     * method in any open/abstract class constructors since it directly invokes open members.
```
sorenoid, line 151: 👍

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseFieldState.kt
```diff
@@ -159,7 +159,9 @@ internal abstract class BaseFieldState<T>(
 
     /**
      * Start observing the [isEditable], [isRequired] and [isFocused] flows and update
-     * the validation to generate any validation errors.
+     * the validation to generate any validation errors. This method must NOT be invoked from
```
sorenoid, line 162: 👍

sorenoid (approved):
LGTM.

kaushikrw:
thanks @sorenoid merging!

PR #289, opened 2024-01-12
Feature Editor: Update geometry editor symbology

Luiy0:
**Issue(s)**
[2062](https://devtopia.esri.com/runtime/bistromath/issues/2062)

**Description of change**
Create helper method that applies feature symbology to the geometry editor when the feature is being edited. 

To test with data that doesn't have a feature form definition, remove the feature form non-null requirement in the `FeatureEditorAppState.onSingleTapConfirmed()` and `FeatureEditor.start()`. If the feature geometry is not allowed to be edited, comment the appropriate line when the feature editor is initialised, see [issue](https://devtopia.esri.com/runtime/bistromath/issues/2191). 

**Testing**
Manual testing with a variety of web maps linked in the issue.

### DIFF
--- a/toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
+++ b/toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
@@ -192,7 +192,7 @@ private fun updateGeometryEditorStyle(geometryEditor: GeometryEditor, feature: A
     else feature.featureTable?.geometryType == GeometryType.Polygon
 
     val renderer = (feature.featureTable?.layer as? FeatureLayer)?.renderer
-    val featureSymbol = renderer?.getSymbol(feature, true)
+    val featureSymbol = renderer?.getSymbol(feature, true) ?: return
 
     if (isGeometryPoint) {
         geometryEditor.tool.style.apply {
--- a/toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
+++ b/toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
@@ -207,8 +207,8 @@ private fun updateGeometryEditorStyle(geometryEditor: GeometryEditor, feature: A
             val vertexOutlineSymbol = SimpleLineSymbol(SimpleLineSymbolStyle.Solid, color = Color.black)
             when (featureSymbol) {
                 is SimpleLineSymbol -> {
-                    val vertexSize = max(featureSymbol.width, 3f) * 3
-                    val midVertexSize = max(featureSymbol.width, 3f) * 2
+                    val vertexSize = max(featureSymbol.width, 1f) * 5
+                    val midVertexSize = max(featureSymbol.width, 1f) * 4
 
                     val vertex = SimpleMarkerSymbol(
                         color = featureSymbol.color.opaque,
@@ -241,8 +241,8 @@ private fun updateGeometryEditorStyle(geometryEditor: GeometryEditor, feature: A
                     val strokeSymbolLayer = featureSymbol.symbolLayers.filterIsInstance<StrokeSymbolLayer>().firstOrNull()
                     if (strokeSymbolLayer != null) {
                         val symbolStrokeWidth = strokeSymbolLayer.width.toFloat()
-                        val vertexSize = max(symbolStrokeWidth, 3f) * 3
-                        val midVertexSize = max(symbolStrokeWidth, 3f) * 2
+                        val vertexSize = max(symbolStrokeWidth, 1f) * 5
+                        val midVertexSize = max(symbolStrokeWidth, 1f) * 4
 
                         val vertex = SimpleMarkerSymbol(
                             color = featureSymbol.color.opaque,
@@ -288,8 +288,8 @@ private fun updateGeometryEditorStyle(geometryEditor: GeometryEditor, feature: A
                     featureSymbol.outline?.let {
                         outlineColor = it.color
                         outlineWidth = it.width
-                        vertexSize = max(outlineWidth, 3f) * 3
-                        midVertexSize = max(outlineWidth, 3f) * 2
+                        vertexSize = max(outlineWidth, 1f) * 5
+                        midVertexSize = max(outlineWidth, 1f) * 4
                     }
 
                     // Color can be null (no fill)
@@ -325,8 +325,8 @@ private fun updateGeometryEditorStyle(geometryEditor: GeometryEditor, feature: A
                     val strokeSymbolLayer = featureSymbol.symbolLayers.filterIsInstance<StrokeSymbolLayer>().firstOrNull()
                     if (strokeSymbolLayer != null) {
                         val symbolStrokeWidth = strokeSymbolLayer.width.toFloat()
-                        val vertexSize = max(symbolStrokeWidth, 3f) * 3
-                        val midVertexSize = max(symbolStrokeWidth, 3f) * 2
+                        val vertexSize = max(symbolStrokeWidth, 1f) * 5
+                        val midVertexSize = max(symbolStrokeWidth, 1f) * 4
 
                         val vertex = SimpleMarkerSymbol(
                             color = featureSymbol.color.opaque ,
--- a/toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
+++ b/toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
@@ -47,6 +47,7 @@ import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asSharedFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.launch
+import kotlin.math.max
 
 public sealed class FinishState(public open val feature: ArcGISFeature) {
     public data class Discarded(override val feature: ArcGISFeature) : FinishState(feature)
@@ -206,12 +207,20 @@ private fun updateGeometryEditorStyle(geometryEditor: GeometryEditor, feature: A
             val vertexOutlineSymbol = SimpleLineSymbol(SimpleLineSymbolStyle.Solid, color = Color.black)
             when (featureSymbol) {
                 is SimpleLineSymbol -> {
-                    val vertexWidth = if (featureSymbol.width < 2) featureSymbol.width * 10 else featureSymbol.width * 3
-                    val midVertexWidth = if (featureSymbol.width < 2) featureSymbol.width * 7 else featureSymbol.width * 2
+                    val vertexSize = max(featureSymbol.width, 3f) * 3
+                    val midVertexSize = max(featureSymbol.width, 3f) * 2
 
-                    val vertex = SimpleMarkerSymbol(color = featureSymbol.color, style = SimpleMarkerSymbolStyle.Circle, size = vertexWidth)
+                    val vertex = SimpleMarkerSymbol(
+                        color = featureSymbol.color.opaque,
+                        style = SimpleMarkerSymbolStyle.Circle,
+                        size = vertexSize
+                    )
                     vertex.outline = vertexOutlineSymbol
-                    val midVertex = SimpleMarkerSymbol(color = Color.white, style = SimpleMarkerSymbolStyle.Circle, size = midVertexWidth)
+                    val midVertex = SimpleMarkerSymbol(
+                        color = Color.white,
+                        style = SimpleMarkerSymbolStyle.Circle,
+                        size = midVertexSize
+                    )
                     midVertex.outline = vertexOutlineSymbol
 
                     vertexSymbol = vertex
@@ -219,29 +228,48 @@ private fun updateGeometryEditorStyle(geometryEditor: GeometryEditor, feature: A
                     feedbackVertexSymbol = vertex
                     midVertexSymbol = midVertex
                     selectedMidVertexSymbol = midVertex
-                    feedbackLineSymbol = SimpleLineSymbol(SimpleLineSymbolStyle.Dash, color = featureSymbol.color, width = featureSymbol.width)
+                    feedbackLineSymbol = SimpleLineSymbol(
+                        SimpleLineSymbolStyle.Dash,
+                        color = featureSymbol.color.opaque,
+                        width = featureSymbol.width
+                    )
                     lineSymbol = featureSymbol
                     vertexTextSymbol = null
                 }
                 is MultilayerPolylineSymbol -> {
-                    // Can potentially use featureSymbol.width instead
-                    val symbolStrokeWidth = (featureSymbol.symbolLayers[0] as StrokeSymbolLayer).width.toFloat()
-                    val vertexSize = if (symbolStrokeWidth < 2) symbolStrokeWidth * 10 else symbolStrokeWidth * 3
-                    val midVertexSize = if (symbolStrokeWidth < 2) symbolStrokeWidth * 7 else symbolStrokeWidth * 2
+                    // Find the first stroke symbol layer in the list of layers, if any.
+                    val strokeSymbolLayer = featureSymbol.symbolLayers.filterIsInstance<StrokeSymbolLayer>().firstOrNull()
+                    if (strokeSymbolLayer != null) {
+                        val symbolStrokeWidth = strokeSymbolLayer.width.toFloat()
+                        val vertexSize = max(symbolStrokeWidth, 3f) * 3
+                        val midVertexSize = max(symbolStrokeWidth, 3f) * 2
 
-                    val vertex = SimpleMarkerSymbol(color = featureSymbol.color, style = SimpleMarkerSymbolStyle.Circle, size = vertexSize)
-                    vertex.outline = vertexOutlineSymbol
-                    val midVertex = SimpleMarkerSymbol(color = Color.white, style = SimpleMarkerSymbolStyle.Circle, size = midVertexSize)
-                    midVertex.outline = vertexOutlineSymbol
+                        val vertex = SimpleMarkerSymbol(
+                            color = featureSymbol.color.opaque,
+                            style = SimpleMarkerSymbolStyle.Circle,
+                            size = vertexSize
+                        )
+                        vertex.outline = vertexOutlineSymbol
+                        val midVertex = SimpleMarkerSymbol(
+                            color = Color.white,
+                            style = SimpleMarkerSymbolStyle.Circle,
+                            size = midVertexSize
+                        )
+                        midVertex.outline = vertexOutlineSymbol
 
-                    vertexSymbol = vertex
-                    selectedVertexSymbol = vertex
-                    feedbackVertexSymbol = vertex
-                    midVertexSymbol = midVertex
-                    selectedMidVertexSymbol = midVertex
-                    feedbackLineSymbol = SimpleLineSymbol(SimpleLineSymbolStyle.Dash, color = featureSymbol.color, width = symbolStrokeWidth)
-                    lineSymbol = featureSymbol
-                    vertexTextSymbol = null
+                        vertexSymbol = vertex
+                        selectedVertexSymbol = vertex
+                        feedbackVertexSymbol = vertex
+                        midVertexSymbol = midVertex
+                        selectedMidVertexSymbol = midVertex
+                        feedbackLineSymbol = SimpleLineSymbol(
+                            SimpleLineSymbolStyle.Dash,
+                            color = featureSymbol.color.opaque,
+                            width = symbolStrokeWidth
+                        )
+                        lineSymbol = featureSymbol
+                        vertexTextSymbol = null
+                    }
                 }
                 else -> {} // Use default symbology
             }
@@ -252,22 +280,30 @@ private fun updateGeometryEditorStyle(geometryEditor: GeometryEditor, feature: A
             when (featureSymbol) {
                 is SimpleFillSymbol -> {
                     // Defaults when outline is null
-                    var vertexSize = (vertexSymbol as? SimpleMarkerSymbol)?.size ?: 2f
+                    var vertexSize = (vertexSymbol as? SimpleMarkerSymbol)?.size ?: 3f
                     var midVertexSize = vertexSize / 2
                     var outlineColor = Color.white
-                    var outlineWidth = (feedbackLineSymbol as? SimpleLineSymbol)?.width ?: 2f
+                    var outlineWidth = (feedbackLineSymbol as? SimpleLineSymbol)?.width ?: 3f
 
                     featureSymbol.outline?.let {
                         outlineColor = it.color
                         outlineWidth = it.width
-                        vertexSize = if (outlineWidth < 2) outlineWidth * 10 else outlineWidth * 3
-                        midVertexSize = if (outlineWidth < 2) outlineWidth * 7 else outlineWidth * 2
+                        vertexSize = max(outlineWidth, 3f) * 3
+                        midVertexSize = max(outlineWidth, 3f) * 2
                     }
 
                     // Color can be null (no fill)
-                    val vertex = SimpleMarkerSymbol(color = featureSymbol.color, style = SimpleMarkerSymbolStyle.Circle, size = vertexSize)
+                    val vertex = SimpleMarkerSymbol(
+                        color = featureSymbol.color.opaque,
+                        style = SimpleMarkerSymbolStyle.Circle,
+                        size = vertexSize
+                    )
                     vertex.outline = vertexOutlineSymbol
-                    val midVertex = SimpleMarkerSymbol(color = Color.white, style = SimpleMarkerSymbolStyle.Circle, size = midVertexSize)
+                    val midVertex = SimpleMarkerSymbol(
+                        color = Color.white,
+                        style = SimpleMarkerSymbolStyle.Circle,
+                        size = midVertexSize
+                    )
                     midVertex.outline = vertexOutlineSymbol
 
                     vertexSymbol = vertex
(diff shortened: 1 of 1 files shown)


toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+                    (selectedVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (selectedMidVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (feedbackVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (midVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    vertexTextSymbol = null
+                }
+                is MultilayerPolygonSymbol -> {
+                    // Check nullability of symbol layers
+                    val symbolStrokeWidth = (featureSymbol.symbolLayers[1] as StrokeSymbolLayer).width.toFloat()
+                    val opaqueFillColor = Color.fromRgba(featureSymbol.color.component2(), featureSymbol.color.component3(), featureSymbol.color.component4())
```
fpearson1111, line 313: It would be better to use the `red`, `green` and `blue` properties on `featureSymbol.color` instead of these component functions -- they are more for enabling destructuring such as
```kotlin
val (a, r, g, b) = myColor
```

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+                    (vertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (selectedVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (selectedMidVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (feedbackVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (midVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    vertexTextSymbol = null
+                }
+                is MultilayerPolygonSymbol -> {
+                    // Check nullability of symbol layers
+                    val symbolStrokeWidth = (featureSymbol.symbolLayers[1] as StrokeSymbolLayer).width.toFloat()
```
fpearson1111, line 312: Will this data always be at index 1 in the symbol layers?

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+                    fillSymbol = featureSymbol
+
+                    (vertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (selectedVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (selectedMidVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (feedbackVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (midVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    vertexTextSymbol = null
+                }
+                is PictureMarkerSymbol -> {
```
fpearson1111, line 336: Hmm I am a bit confused about this case as I thought PictureMarkerSymbol only applied to points and multipoints. What does it mean for a PictureMarkerSymbol to be used with a polygon? Also should there be a PictureMarkerSymbol case for when the geometry is a point/mulitpoint?

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+                    val selectedVertex = SimpleMarkerSymbol(color = Color.white , style = SimpleMarkerSymbolStyle.Circle, size = featureSymbol.toMultilayerSymbol().size)
+                    val midVertex = SimpleMarkerSymbol(color = Color.white , style = SimpleMarkerSymbolStyle.Circle, size = featureSymbol.toMultilayerSymbol().size/3)
+
+                    vertexSymbol = featureSymbol
+                    selectedVertexSymbol = selectedVertex
+                    feedbackVertexSymbol= selectedVertex
+                    midVertexSymbol = midVertex
+                    selectedMidVertexSymbol = midVertex
+                    feedbackLineSymbol = SimpleLineSymbol(SimpleLineSymbolStyle.Dash, featureSymbol.toMultilayerSymbol().color ,width = featureSymbol.toMultilayerSymbol().size/5)
```
fpearson1111, line 345:
```suggestion
                    feedbackLineSymbol = SimpleLineSymbol(SimpleLineSymbolStyle.Dash, featureSymbol.toMultilayerSymbol().color, width = featureSymbol.toMultilayerSymbol().size/5)
```

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+                    (vertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (selectedVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (selectedMidVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (feedbackVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (midVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    vertexTextSymbol = null
+                }
+                is PictureMarkerSymbol -> {
+                    val selectedVertex = SimpleMarkerSymbol(color = Color.white , style = SimpleMarkerSymbolStyle.Circle, size = featureSymbol.toMultilayerSymbol().size)
+                    val midVertex = SimpleMarkerSymbol(color = Color.white , style = SimpleMarkerSymbolStyle.Circle, size = featureSymbol.toMultilayerSymbol().size/3)
```
fpearson1111, line 338:
```suggestion
                    val selectedVertex = SimpleMarkerSymbol(color = Color.white, style = SimpleMarkerSymbolStyle.Circle, size = featureSymbol.toMultilayerSymbol().size)
                    val midVertex = SimpleMarkerSymbol(color = Color.white, style = SimpleMarkerSymbolStyle.Circle, size = featureSymbol.toMultilayerSymbol().size/3)
```

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
fpearson1111, line 345: Not sure about using the size of the PictureMarkerSymbol to work out a size for the feedback line symbol as they don't really seem like related things. I would be more in favour of using the same size in all cases, maybe taken from the default line symbol that's given by the geometry editor?

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+                    vertexSymbol = featureSymbol
+                    selectedVertexSymbol = featureSymbol
+                    feedbackVertexSymbol = featureSymbol
+                    vertexTextSymbol = null
+                }
+                else -> { }
+            }
+        }
+    } else if (isGeometryLine) {
+        geometryEditor.tool.style.apply {
```
fpearson1111, line 224: Just noting that this geometry editor is passed in by the user so in principle they may have already done some stuff to the style, which could change the effect of this method. I wouldn't really say that's a problem because we can just document it, and probably almost all users will start with a fresh geometry editor.

We could consider whether we want the feature editor to create a new geometry editor when it is created, and then get the user to set that one onto the MapView they are using, rather than what we are currently doing. But that would be part of a separate issue.

fpearson1111 (commented):
Nice work, I have a few comments but the overall approach looks good to me

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
fpearson1111: Also does this thing about needing to make the colour opaque apply to lines and points too?

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+    val renderer = (feature?.featureTable?.layer as? FeatureLayer)?.renderer
+    val featureSymbol = renderer?.getSymbol(feature, true)
+
+    if (isGeometryPoint) {
+        geometryEditor.tool.style.apply {
+            when (featureSymbol) {
+                is SimpleMarkerSymbol -> {
+                    vertexSymbol = featureSymbol
+                    selectedVertexSymbol = featureSymbol
+                    feedbackVertexSymbol = featureSymbol
```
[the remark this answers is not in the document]
shellygill: I agree this could be a potential enhancement, but low priority.

In some Esri apps, they dont actually have the equivalent of a `selectedVertexSymbol`, as there's no real notion of a selected symbol - the UI just responds to whatever you interact with. In our `GeometryEditor` the option is there for users to use though if that's the behaviour they want, and the selected symbol is differentiated by default. The selected vertex is differentiated visually already by the selection halo, so the user doesnt actually miss any crutial information by not seeing a different `selectedVertexSymbol` hence low priority. (It might be a matter of opinion/taste as to whether you prefer differentiating all the symbols or not). I *think* that the default symbols might acutally be the same for `vertexSymbol` and `selectedVertexSymbol` and also `feedbackVertexSymbol` but would need to confirm that for sure....).

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+                    vertexTextSymbol = null
+                }
+                is MultilayerPointSymbol -> {
+                    vertexSymbol = featureSymbol
+                    selectedVertexSymbol = featureSymbol
+                    feedbackVertexSymbol = featureSymbol
+                    vertexTextSymbol = null
+                }
+                is PictureMarkerSymbol -> {
```
shellygill, line 214: PictureMarkerSymbol and SimpleMarkerSymbol are both instances of MarkerSymbol, so it seems to me that these cases could be combined. In face, given that there is no differentiation between how these different cases are treated (`featureSymbol` is always assigned to two properties and the other set to null), I'm not sure of the benefit of checking the symbol type at all. If the symbol is currently being used to symbolise a `Point` feature, then I'd think that same symbol could safely be used to symbolise the `Point`s that comprise our geometry editor style properties that also draw a point. That would then remove the issue of the default symbology below? Worth testing with your known cases, but I'm not sure conceptually I know any reason why this should not work.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+                    (vertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (selectedVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (selectedMidVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (feedbackVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    (midVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    vertexTextSymbol = null
+                }
+                is MultilayerPolylineSymbol -> {
+                    // Can potentially use featureSymbol.width instead
+                    val symbolStrokeWidth = (featureSymbol.symbolLayers[0] as StrokeSymbolLayer).width.toFloat()
```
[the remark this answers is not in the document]
shellygill: I would have thought that won't always be the case - I think multilayer symbols might be a collection in any order, based on how the user wanted to create things? [This web app](https://sagewall.github.io/symbol-builder/cim-symbol/) can help you experiement to understand things a bit (I dont find it great for usability but good for a bit of random symbol hacking....).

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+                    val midVertexWidth = if (featureSymbol.width < 2) featureSymbol.width * 7 else featureSymbol.width * 2
+                    val selectedVertex = SimpleMarkerSymbol(color = Color.white, style = SimpleMarkerSymbolStyle.Circle, size = vertexWidth)
+                    val midVertex = SimpleMarkerSymbol(color = Color.white, style = SimpleMarkerSymbolStyle.Circle, size = midVertexWidth)
+
+                    vertexSymbol = SimpleMarkerSymbol(color = featureSymbol.color, style = SimpleMarkerSymbolStyle.Circle, size = vertexWidth)
+                    selectedVertexSymbol = selectedVertex
+                    feedbackVertexSymbol = selectedVertex
+                    midVertexSymbol = midVertex
+                    selectedMidVertexSymbol = midVertex
+                    feedbackLineSymbol = SimpleLineSymbol(SimpleLineSymbolStyle.Dash, color = featureSymbol.color, width = featureSymbol.width)
```
shellygill, line 238: 👍 I think this is a really good addition - taking colout and width from the existing symbol but making it the familiar dash line style for feedback.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+        geometryEditor.tool.style.apply {
+            val vertexOutlineSymbol = SimpleLineSymbol(SimpleLineSymbolStyle.Solid, color = Color.black)
+            when (featureSymbol) {
+                is SimpleLineSymbol -> {
+                    val vertexWidth = if (featureSymbol.width < 2) featureSymbol.width * 10 else featureSymbol.width * 3
+                    val midVertexWidth = if (featureSymbol.width < 2) featureSymbol.width * 7 else featureSymbol.width * 2
+                    val selectedVertex = SimpleMarkerSymbol(color = Color.white, style = SimpleMarkerSymbolStyle.Circle, size = vertexWidth)
+                    val midVertex = SimpleMarkerSymbol(color = Color.white, style = SimpleMarkerSymbolStyle.Circle, size = midVertexWidth)
+
+                    vertexSymbol = SimpleMarkerSymbol(color = featureSymbol.color, style = SimpleMarkerSymbolStyle.Circle, size = vertexWidth)
```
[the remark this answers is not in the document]
shellygill: > but adds a halo around selected vertices

The halo (AFAIK) is just built-in, based on the internal graphics layer graphic members being `selected` state.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+                    (midVertexSymbol as SimpleMarkerSymbol).outline = vertexOutlineSymbol
+                    vertexTextSymbol = null
+                }
+                is MultilayerPolylineSymbol -> {
+                    // Can potentially use featureSymbol.width instead
+                    val symbolStrokeWidth = (featureSymbol.symbolLayers[0] as StrokeSymbolLayer).width.toFloat()
+                    val vertexSize = if (symbolStrokeWidth < 2) symbolStrokeWidth * 10 else symbolStrokeWidth * 3
+                    val midVertexSize = if (symbolStrokeWidth < 2) symbolStrokeWidth * 7 else symbolStrokeWidth * 2
+                    val selectedVertex = SimpleMarkerSymbol(color = Color.white, style = SimpleMarkerSymbolStyle.Circle, size = vertexSize)
+                    val midVertex = SimpleMarkerSymbol(color = Color.white, style = SimpleMarkerSymbolStyle.Circle, size = midVertexSize)
```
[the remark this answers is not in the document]
shellygill: With CIM symbols, I'm not sure how you'd be able to tell if a marker is used for vertices, vs being used as a repeating symbol along the line - it might be there in the API somewhere but I've not used it enough to know. Worth adding to a 'potential enhancements' followup but probably not worth holding this first PR for is my take.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+
+                    vertexSymbol = SimpleMarkerSymbol(color = featureSymbol.color, style = SimpleMarkerSymbolStyle.Circle, size = vertexSize)
```
shellygill, line 256: Potentially could refactor out these repeated code chunks setting vertexSymbol, etc (might be a kotlin toolkit code style question though...).

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+                    val selectedVertex = SimpleMarkerSymbol(color = Color.white , style = SimpleMarkerSymbolStyle.Circle, size = vertexSize)
+                    val midVertex = SimpleMarkerSymbol(color = Color.white, style = SimpleMarkerSymbolStyle.Circle, size = midVertexSize)
+
+                    vertexSymbol = SimpleMarkerSymbol(color = featureSymbol.color, style = SimpleMarkerSymbolStyle.Circle, size = vertexSize)
+                    selectedVertexSymbol = selectedVertex
+                    feedbackVertexSymbol = selectedVertex
+                    midVertexSymbol = midVertex
+                    selectedMidVertexSymbol = midVertex
+                    feedbackLineSymbol = SimpleLineSymbol(SimpleLineSymbolStyle.Dash, color = outlineColor, width = outlineWidth)
+                    lineSymbol = featureSymbol // Could be null (no outline)
```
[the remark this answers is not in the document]
shellygill: It it that the internals of the geometry editor code do something special with the symbols here, and just take the outline if it's a fill symbol? (Not sure, I think that someone mentioned that - possibly yourself @Luiy0 ?)

shellygill (changes requested):
I've added a few comments to be going on with. I think my own lack of knowledge about the details of CIM symbology may be limiting me, and perhaps your own testing @Luiy0 can explain some of FInlay's questions about the types of symbol used for different geometries.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+    else feature?.featureTable?.geometryType == GeometryType.Point || feature?.featureTable?.geometryType == GeometryType.Multipoint
+
+    val isGeometryLine = if (geometry?.let { !it.isEmpty } == true) geometry is Polyline
+    else feature?.featureTable?.geometryType == GeometryType.Polyline
+
+    val isGeometryPolygon = if (geometry?.let { !it.isEmpty } == true) geometry is Polygon
+    else feature?.featureTable?.geometryType == GeometryType.Polygon
+
+    val renderer = (feature?.featureTable?.layer as? FeatureLayer)?.renderer
+    val featureSymbol = renderer?.getSymbol(feature, true)
```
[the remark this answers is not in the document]
Luiy0: The method currently relies on `featureSymbol` not being null but if it is then I agree that we should exit early. We would then use default symbology from the geometry editor. Should the feature editor have different default symbology to that of the geometry editor?

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+        }
+    } else if (isGeometryLine) {
+        geometryEditor.tool.style.apply {
+            val vertexOutlineSymbol = SimpleLineSymbol(SimpleLineSymbolStyle.Solid, color = Color.black)
+            when (featureSymbol) {
+                is SimpleLineSymbol -> {
+                    val vertexWidth = if (featureSymbol.width < 2) featureSymbol.width * 10 else featureSymbol.width * 3
+                    val midVertexWidth = if (featureSymbol.width < 2) featureSymbol.width * 7 else featureSymbol.width * 2
+                    val selectedVertex = SimpleMarkerSymbol(color = Color.white, style = SimpleMarkerSymbolStyle.Circle, size = vertexWidth)
+                    val midVertex = SimpleMarkerSymbol(color = Color.white, style = SimpleMarkerSymbolStyle.Circle, size = midVertexWidth)
```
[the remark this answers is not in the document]
Luiy0: It was mostly personal preference since I find the squares to be too bulky and noticeable. AGOL uses circles while Pro uses hollow squares. This brings me back to the question above about the feature editor having its own default symbology.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -181,3 +182,178 @@ public class FeatureEditor(
...
+                else -> { }
+            }
+        }
+    } else if (isGeometryLine) {
+        geometryEditor.tool.style.apply {
+            val vertexOutlineSymbol = SimpleLineSymbol(SimpleLineSymbolStyle.Solid, color = Color.black)
+            when (featureSymbol) {
+                is SimpleLineSymbol -> {
+                    val vertexWidth = if (featureSymbol.width < 2) featureSymbol.width * 10 else featureSymbol.width * 3
+                    val midVertexWidth = if (featureSymbol.width < 2) featureSymbol.width * 7 else featureSymbol.width * 2
```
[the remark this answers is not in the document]
Luiy0: Originally, line width comparisons were made against a width of one as you suggest but I recall that for width values of around 1.3, the vertices would still appear too small hence why I increased the comparison value. I had to use two different multiplayers depending on the line width to maintain proportionality between line and vertex/midVertex. 

As an example, if a line width is 0.86, the above suggestion would make the vertex appear too small. We could change the multiplier to 10 but that would not work for a line feature with a width of 2.66. I'm not sure if comparing with a value of two this way is the better option but it got me closer to the results I was expecting.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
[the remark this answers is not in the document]
Luiy0: I would think that because these are multilayer symbols, they must have at least one symbol layer which most likely corresponds to the geometry they are designed to render. The line features I've been working with don't contain vertices which agrees with the symbol not having a markerLayer and only a strokeLayer at `symbolLayers[0]`. For other complicated multilayerSymbols this order wouldn't necessarily work since it depends on the available layers which I assume follow the order in [this](https://developers.arcgis.com/net/api-reference/api/netwin/Esri.ArcGISRuntime/Esri.ArcGISRuntime.Symbology.SymbolLayer.html#:~:text=The%20following%20table%20lists%20symbol%20types%20and%20symbol%20layers%3A) table or perhaps the order in which they are applied. However, the Qt documentation also states [here](https://developers.arcgis.com/qt/cpp/api-reference/esri-arcgisruntime-multilayersymbol.html#:~:text=Note%3A%20Symbol%20layers%20in%20a%20multilayer%20symbol%20are%20in%20reverse%20order%20of%20how%20they%20appear%20in%20the%20JSON%20representation.):

> Symbol layers in a multilayer symbol are in reverse order of how they appear in the JSON representation.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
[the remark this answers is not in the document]
shellygill: > Following from what was done in AGOL, a vertex changes to white when selected and back to the original colour when unselected

It sounds like the white color change is their equivalent of our selection halo indicating the "current selection", so I don't see a specific need to emulate the color change too.

A lot of these decisions are very subjective though, so I wouldnt spend too much time worrying about the finer detail personally. If such logic were to be pushed down into core for example, I'd imagine that we would want to validate / review most of the choices made, and also really make sure we understand what we're most likely to encounter in terms of CIM symbology (perhaps by discussing with whatever feature team knows that area).

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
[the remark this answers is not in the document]
Luiy0: It's likely possible to get vertex symbology from a markerLayer and it's probably straightforward if the only symbology contained in this layer is reserved for vertices. If not, then the symbology could refer to other things at arbitrary locations along lines, outlines, or within areas.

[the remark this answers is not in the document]
shellygill: > It's likely possible to get vertex symbology from a markerLayer and it's probably straightforward if the only symbology contained in this layer is reserved for vertices.

The symbol must contain the information required to decide whether to apply the marker symbol to vertices, or to locations along the line - how is that expressed in the API @Luiy0 ?

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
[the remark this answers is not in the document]
Luiy0: I'm not sure how this internally works but it has worked for all the features that I've tested with and it's also simpler than creating a SimpleLineSymbol. I also ended up preferring it as to not enforce an outline on a feature that doesn't have one because it can appear strange. I became more aware of this when trying to decide on the size and colour of the outline.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
Luiy0: I think it's unusual for point or line features to have a translucent colour. This colour breakup was only necessary because of polygon fills usually being translucent and we needed to use this colour for the vertices since it didn't have any. I suppose it applies but I'm unsure if the unlikeliness of it happening means we could ignore it?

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
Luiy0: I can't say for certain.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
Luiy0: Yeah, this is an odd one to edit which I tested with [this](https://runtimecoretest.maps.arcgis.com/home/item.html?id=9452f271998742d6b0b6a8a4cffb234c) webmap. The feature geometry comes back as a polygon yet it's represented by a picutreMarkerSymbol which I had to convert to a multilayerPointSymbol because getting information from the image was a bit complicated. There is a case for point/multipoint.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
Luiy0: The size in this case comes from a multilayer point symbol. The other only size I can get is from the width/height of the image. I would still have to resize the default line you suggest depending on the point symbol for proportion.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
Luiy0: Agreed. Works fine, thanks!

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
Luiy0: I could create one SimpleMarkerSymbol but it'll most likely have to be resized after creation as the midVertex is just a bit smaller.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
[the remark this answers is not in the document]
Luiy0: From Pro documentation [here](https://pro.arcgis.com/en/pro-app/3.1/help/mapping/layer-properties/position-and-place-marker-symbol-layers.htm#:~:text=When%20a%20marker,geometry%20type%20only.):

> When a marker symbol layer is included in a line or polygon symbol, it includes a marker placement, which determines how and where the symbol positions the markers relative to the feature geometry. Marker symbol layers in a point symbol do not have a marker placement; they are always placed relative to the point geometry. There are a number of marker placements to choose from. Most are valid in both line and polygon symbols, but some are valid in one geometry type only.

We should be able to access the `SymbolAnchor`and `SymbolAnchorPlacementMode` through the `MarkerSymbolLayer`.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
fpearson1111: >  I would still have to resize the default line you suggest depending on the point symbol for proportion.

Ah okay, I think I understand the reason for this a bit better now. Happy to keep it as it is :)

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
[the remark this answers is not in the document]
Luiy0: I've updated this to use max and so far it has worked well. Thanks.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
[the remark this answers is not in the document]
Luiy0: Updated for feature to remain with the same colour when selected (midVertices are still white).

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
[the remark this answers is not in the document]
Luiy0: Filtering works well, thanks.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
Luiy0: Using opaque colours all around now! Thanks.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -206,46 +210,69 @@ private fun updateGeometryEditorStyle(geometryEditor: GeometryEditor, feature: A
...
-                    selectedVertexSymbol = selectedVertex
-                    feedbackVertexSymbol = selectedVertex
-                    midVertexSymbol = midVertex
-                    selectedMidVertexSymbol = midVertex
-                    feedbackLineSymbol = SimpleLineSymbol(SimpleLineSymbolStyle.Dash, color = featureSymbol.color, width = symbolStrokeWidth)
-                    lineSymbol = featureSymbol
-                    vertexTextSymbol = null
+                    // Find the first stroke symbol layer in the list of layers, if any.
+                    val strokeSymbolLayer = featureSymbol.symbolLayers.filterIsInstance<StrokeSymbolLayer>().firstOrNull()
+                    if (strokeSymbolLayer != null) {
```
fpearson1111, line 245: This approach looks good but I think I was being a bit pessimistic in my comment by saying that if this is null then we need to fall back to default symbology. Since this layer is only being used to work out the sizes of various lines and vertices, I think it would make sense to use default values for those sizes if this symbol is null, for example by having a default value for `symbolStrokeWidth`. Same applies to the polygon case.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -256,72 +283,105 @@ private fun updateGeometryEditorStyle(geometryEditor: GeometryEditor, feature: A
...
                     // Defaults when outline is null
-                    var vertexSize = (vertexSymbol as? SimpleMarkerSymbol)?.size ?: 2f
+                    var vertexSize = (vertexSymbol as? SimpleMarkerSymbol)?.size ?: 5f
                     var midVertexSize = vertexSize / 2
                     var outlineColor = Color.white
-                    var outlineWidth = (feedbackLineSymbol as? SimpleLineSymbol)?.width ?: 2f
+                    var outlineWidth = (feedbackLineSymbol as? SimpleLineSymbol)?.width ?: 3f
 
                     featureSymbol.outline?.let {
                         outlineColor = it.color
```
fpearson1111, line 292: Could also be made opaque?

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
```diff
@@ -256,72 +283,105 @@ private fun updateGeometryEditorStyle(geometryEditor: GeometryEditor, feature: A
...
-                    vertexTextSymbol = null
+                    // Find the first stroke symbol layer in the list of layers, if any.
+                    val strokeSymbolLayer = featureSymbol.symbolLayers.filterIsInstance<StrokeSymbolLayer>().firstOrNull()
+                    if (strokeSymbolLayer != null) {
+                        val symbolStrokeWidth = strokeSymbolLayer.width.toFloat()
+                        val vertexSize = max(symbolStrokeWidth, 1f) * 5
+                        val midVertexSize = max(symbolStrokeWidth, 1f) * 4
+
+                        val vertex = SimpleMarkerSymbol(
+                            color = featureSymbol.color.opaque ,
```
fpearson1111, line 335:
```suggestion
                            color = featureSymbol.color.opaque,
```

fpearson1111 (commented):
Few more small comments ☺️

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
Luiy0: Updated.

toolkit/featureeditor/src/main/java/com/arcgismaps/toolkit/featureeditor/FeatureEditor.kt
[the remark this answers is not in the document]
shellygill: I don't think we should come up with a separate default set of symbology for the feature editor at this point, no. If we were to come up with a new set of default GeometryEditor geometry then I think we'd need to be clear on what would be informing the decisions about that default set of symbology, and I don't feel like we know a default set of feature symbology that would be a better set of symbols to follow there, as symbology is so varied and so specific to the data and map. Let me know if my thinking is off here and you did have a specific set of symbols in mind though. To my thinking, possibly there would never be a better set of default symbology. But equally possibly it does become obvious that we should switch to a different specific set of default symbols - and if so then we can come back to this idea when that becomes clear. Until then, there's a default set of symbology for the geometry editor that, while it might not look great for all cases of existing feature geometry, does the job.

fpearson1111 (approved):
LGTM

PR #290, opened 2024-01-12
Update the camera controller be nullable

changanxian:
- Make the CameraController be nullable


hud10837 (commented):
@changanxian can you explain why this should be nullable? It is not nullable in the SDK

changanxian:
@hud10837 After thinking again, and review the [API guideline](After reviewing the [API guideline](https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-component-api-guidelines.md), I feel we don't align with Swift for this case.), I feel we don't need to make a change to align with Swift in this case. Thanks

PR #292, opened 2024-01-17
`Forms`: Submit Form Validation

kaushikrw:
### Summary of changes

- Added the usage of `FeatureForm::getValidationErrors()` during form submission in the microapp to show any validation errors (for editable fields). If any errors are present, an `AlertDialog` displays the errors preventing the user from submitting the form. If no errors are present a progress dialog is displayed until saving the edits is complete.
- `FeatureForm` composable now accepts a new parameter of `ValidationErrorVisibility` which can be used to set the behavior for when the validation errors become visible on each field.

### DIFF
--- a/microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapScreen.kt
+++ b/microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapScreen.kt
@@ -164,7 +164,7 @@ fun MapScreen(mapViewModel: MapViewModel = hiltViewModel(), onBackPressed: () ->
         }
     }
     if (uiState is UIState.Committing) {
-        SubmitForm(uiState = (uiState as UIState.Committing)) {
+        SubmitForm(errors = (uiState as UIState.Committing).errors) {
             mapViewModel.cancelCommit()
         }
     }
@@ -214,8 +214,7 @@ fun TopFormBar(
 
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
-private fun SubmitForm(uiState: UIState.Committing, onDismissRequest: () -> Unit) {
-    val errors = uiState.errors
+private fun SubmitForm(errors : List<ErrorInfo>, onDismissRequest: () -> Unit) {
     if (errors.isEmpty()) {
         // show a progress dialog if no errors are present
         AlertDialog(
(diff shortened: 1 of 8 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseFieldState.kt
```diff
@@ -146,6 +146,16 @@ internal abstract class BaseFieldState<T>(
         _isFocused.value = focus
     }
 
+    /**
+     * Forces the validation of this field irrespective of the current focus state [isFocused] and
+     * generates any validation errors via the [value] property. Avoid calling this method in any
+     * open/abstract class constructors since it indirectly invokes open members.
+     */
+    fun forceValidation() {
+        wasFocused = true
```
sorenoid, line 155: I think this name should change to something like `alwaysValidate` or `alwaysShowErrors`

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapScreen.kt
```diff
@@ -174,16 +212,140 @@ fun TopFormBar(
...
+        // show all the validation errors in a dialog
+        AlertDialog(
+            onDismissRequest = onDismissRequest,
+            modifier = Modifier.heightIn(max = 600.dp),
+            confirmButton = {
+                Row(
+                    modifier = Modifier.fillMaxWidth(),
+                    horizontalArrangement = Arrangement.Center
+                ) {
+                    Button(onClick = onDismissRequest) {
```
sorenoid, line 245: I think this could use one more button to cancel the submit, and return to the `Automatic` error validation state. I think this would require splitting the logic in `wasFocused` into two variables. If this makes sense, and I am not mistaken, we can set that aside for now as an enhancement.

sorenoid (approved):
LGTM. approving. I have minor naming feedback and maybe a follow on enhancement, if I understand things correctly. The name change is related to the enhancement and can be deferred.

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapScreen.kt
kaushikrw: The `MapViewModel.cancelCommit()` can be made take an argument of `ValidationErrorVisibility` which is set to `automatic` on cancel. `wasFocused` wouldn't need to change because once it is set to `true` it cannot change to a `false`. This essentially makes the change from `automatic` to `always` for a FeatureForm view irreversible.

kaushikrw:
Merging this with the follow up `#380`.

PR #295, opened 2024-01-18
`Forms` : Added discardEdits API and UI

kaushikrw:
### Summary of changes

- Microapp now uses the api `FeatureForm.discardEdits()` instead of custom code.
- A confirmation dialog is displayed when the user clicks on the "cross" icon to discard the current edits to the feature.
- Fixes `369` where the form reloads when a feature form is being edited and the user taps on another feature.

### DIFF
--- a/microapps/FeatureFormsApp/app/src/main/res/values/strings.xml
+++ b/microapps/FeatureFormsApp/app/src/main/res/values/strings.xml
@@ -17,4 +17,7 @@
     <string name="sign_in_with_default_credentials">Sign in using built-in Credentials</string>
     <string name="sign_in_with_enterprise">Sign in with ArcGIS Enterprise</string>
     <string name="skin_sign_in">Skip sign in</string>
+    <string name="discard">Discard</string>
+    <string name="discard_edits">Discard Edits?</string>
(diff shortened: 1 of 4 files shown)


microapps/FeatureFormsApp/app/src/main/res/values/strings.xml
```diff
@@ -17,4 +17,7 @@
     <string name="sign_in_with_default_credentials">Sign in using built-in Credentials</string>
     <string name="sign_in_with_enterprise">Sign in with ArcGIS Enterprise</string>
     <string name="skin_sign_in">Skip sign in</string>
+    <string name="discard">Discard</string>
+    <string name="discard_edits">Discard Edits?</string>
+    <string name="all_changes_will_be_lost">All changes made will be lost.</string>
```
sorenoid, line 22: local changes to attributes made before invoking the form will not be lost, only edits made through the form.

```suggestion
    <string name="all_changes_will_be_lost">All changes made within the form will be lost.</string>
```

sorenoid (changes requested):

kaushikrw:
@sorenoid, ready for another look.

sorenoid (approved):

PR #306, opened 2024-01-23
SceneView: analysis overlay microapp

hud10837:
A basic demonstration of AnalysisOverlay. I considered making this more complex, but ended up reducing it to just displaying and hiding a viewshed.


microapps/SceneViewAnalysisOverlayApp/app/src/main/java/com/arcgismaps/toolkit/sceneviewanalysisoverlayapp/screens/MainScreen.kt
```diff
@@ -0,0 +1,140 @@
...
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun MainScreen() {
+    val arcGISScene = remember {
+        ArcGISScene(BasemapStyle.ArcGISImagery).apply {
+            val surface = Surface()
+            surface.elevationSources.add(
+                ArcGISTiledElevationSource(
+                    "https://elevation3d.arcgis" +
```
ruiqima, line 61: Why do we separate the uri string into two parts? Is it because we want to separate _host_ and _path_?

microapps/SceneViewAnalysisOverlayApp/app/src/main/java/com/arcgismaps/toolkit/sceneviewanalysisoverlayapp/screens/MainScreen.kt
```diff
@@ -0,0 +1,140 @@
...
+            120.0,
+            120.0,
+            0.0,
+            1500.0
+        )
+    }
+    val analysisOverlay = remember { AnalysisOverlay(listOf(viewshed)).apply { isVisible = true } }
+    val analysisOverlays = remember { listOf(analysisOverlay) }
+
+    val graphicsOverlay = GraphicsOverlay(
```
ruiqima, line 92: Just curious that if there's a more preferable way between `val graphicsOverlay = GraphicsOverlay(listOf(oneGraphic))` and `val graphicsOverlay = GraphicsOverlay().graphics.add(oneGraphic))` if we only have one graphic to add?

ruiqima (approved):
Only a few comments, LGTM 👍

microapps/SceneViewAnalysisOverlayApp/app/src/main/java/com/arcgismaps/toolkit/sceneviewanalysisoverlayapp/screens/MainScreen.kt
puneet-pdx: I don't think one is better than the other both are just slightly different way of doing the same thing.

microapps/SceneViewAnalysisOverlayApp/app/src/main/java/com/arcgismaps/toolkit/sceneviewanalysisoverlayapp/screens/MainScreen.kt
puneet-pdx: Good catch.
No, it is just an accidental enter press at that location and the IDE splitting the string into two.

microapps/SceneViewAnalysisOverlayApp/app/src/main/java/com/arcgismaps/toolkit/sceneviewanalysisoverlayapp/screens/MainScreen.kt
```diff
@@ -0,0 +1,139 @@
...
+import com.arcgismaps.mapping.symbology.SimpleMarkerSymbolStyle
+import com.arcgismaps.mapping.view.AnalysisOverlay
+import com.arcgismaps.mapping.view.Graphic
+import com.arcgismaps.mapping.view.GraphicsOverlay
+import com.arcgismaps.mapping.view.OrbitLocationCameraController
+import com.arcgismaps.toolkit.geocompose.SceneView
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun MainScreen() {
```
changanxian, line 55: Could you add a description document for this function?

PR #308, opened 2024-01-30
`Forms` : Update disabled fields style

kaushikrw:
### Summary of changes

- Non editable fields are rendered as read-only fields but with the colors of disabled fields. This is so that the users can copy the text from a non-editable field but to establish a visual distinction they are presented as disabled fields.
- The color and alpha values are taken from the text field [material 3 design](https://m3.material.io/components/text-fields/specs#e4964192-72ad-414f-85b4-4b4357abb83c).
- Placeholder text is also rendered with a lighter alpha compared to the input text.

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseTextFieldDefaults.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseTextFieldDefaults.kt
@@ -16,19 +16,30 @@
 
 package com.arcgismaps.toolkit.featureforms.components.base
 
+import androidx.compose.foundation.interaction.InteractionSource
+import androidx.compose.foundation.interaction.collectIsFocusedAsState
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.OutlinedTextFieldDefaults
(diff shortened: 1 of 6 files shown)


sorenoid (approved):
👍 . UI looks good and works well to provide copy ability without changing due to focus. test improvements are great. thank you!

kaushikrw:
Thanks @sorenoid, merging.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/base/BaseTextFieldDefaults.kt
```diff
@@ -22,52 +22,53 @@ import androidx.compose.material3.TextFieldColors
...
-            MaterialTheme.colorScheme.primary
-        else
-            MaterialTheme.colorScheme.secondary.copy(alpha = 0.8f)
-    
-    @Composable
-    fun textColor(isEditable: Boolean, isEmpty: Boolean, isPlaceHolderEmpty: Boolean): Color {
-        val color = if (isEmpty && !isPlaceHolderEmpty) {
-            Color.Gray
-        } else {
-            MaterialTheme.colorScheme.secondary
```
kaushikrw, line 68: Cause for incorrect color of text.

PR #313, opened 2024-02-07
`Forms`: Attachments API

kaushikrw:
### Summary of changes

- Adds `AttachmentFormElement` and its associated classes which serves as an API until the actual API is available through the SDK.

#### Usage
- A new `AttachmentFormElement` can be created using its asynchronous factory 

```kotlin
AttachmentFormElement.createOrNull(feature: ArcGISFeature, filesDir: String) : AttachmentFormElement?
```
- Since this implementation is built on top of the SDK, caching of attachments is external. Hence the `filesDir` parameter is important as it provides a directory in the file system to cache those attachments.

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/api/AttachmentFormElement.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/api/AttachmentFormElement.kt
@@ -78,6 +78,7 @@ internal class AttachmentFormElement private constructor(
             val attachment = feature.addAttachment(name, contentType, data).getOrNull()
             return if (attachment != null) {
                 val formAttachment = FormAttachment(attachment, filesDir)
+                formAttachment.load()
                 _attachments.add(formAttachment)
                 Result.success(formAttachment)
             } else {
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/api/AttachmentFormElement.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/api/AttachmentFormElement.kt
@@ -16,6 +16,7 @@
 
 package com.arcgismaps.toolkit.featureforms.api
 
+import android.content.Context
 import android.graphics.Bitmap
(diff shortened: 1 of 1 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/api/AttachmentFormElement.kt
sorenoid, line 1: should this class include `fetchAttachmentsAsync()` and `deleteAttachmentAsync()` prototypes?

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/api/AttachmentFormElement.kt
```diff
@@ -0,0 +1,267 @@
...
+    }
+}
+
+/**
+ * Represents an attachment belonging to a feature form. Wraps the Attachment object and adds additional
+ * properties and methods to support displaying attachments in a feature form.
+ *
+ * The [FormAttachment] must be loaded before calling [createFullImage] or [createThumbnail].
+ */
+internal class FormAttachment(
```
sorenoid, line 175: does this class need a prototype of `setName()`?

sorenoid (changes requested):
a couple questions about missing api prototypes.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/api/AttachmentFormElement.kt
kaushikrw: Since the SDK does not expose any public method/property to set the attachment name I did not include it. Would you prefer the api `setName()` is present but it does not do anything?

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/api/AttachmentFormElement.kt
kaushikrw: added `fetchAttachmentsAsync()`

sorenoid (approved):
👍

PR #314, opened 2024-02-07
Add version to API ref

shubham7109:
### Description

PR to add the Dokka versioning plugin to the Toolkit project, to be able to display the ArcGIS version on the KDoc.

image

### Testing

Run the gradle command: `./gradlew :kdoc:dokkaHtml`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -28,6 +28,10 @@ plugins {
 
 buildscript {
     dependencies {
+        // there doesn't appear to be a better way to provide this to subprojects.
+        // this is what lets us put the version number dropdown list in the generated dokka.
+        // it is a "dokka plugin" which is not a gradle plugin, it needs to be on the classpath
+        // before any dependent subproject uses its symbols to configure a dokka task.
         classpath(libs.dokka.versioning)
     }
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -25,3 +25,9 @@ plugins {
     alias(libs.plugins.kotlin.serialization) apply false
     alias(libs.plugins.dokka) apply false
 }
+
+buildscript {
+    dependencies {
+        classpath(libs.dokka.versioning)
```
gunt0001, line 31: Might be worth to add these explanatory comments:
```suggestion
        // there doesn't appear to be a better way to provide this to subprojects.
        // this is what lets us put the version number dropdown list in the generated dokka.
        // it is a "dokka plugin" which is not a gradle plugin, it needs to be on the classpath
        // before any dependent subproject uses its symbols to configure a dokka task.
        classpath(libs.dokka.versioning)
```

gunt0001 (approved):
@shubham7109 - looks good, one minor comment.

fyi - @sorenoid

sorenoid (approved):

PR #317, opened 2024-02-12
attachment UI functional prototype

sorenoid:
The UI I think looks the way we want it to look according to [the discussion](https://devtopia.esri.com/runtime/apollo/discussions/449).

This will need changes to be integrated with the attachments API and state object patterns.

The use of coil for async image loading may be backed out when we have API Attachment objects that are `Loadable`.


kaushikrw (commented):
@sorenoid This looks good. Just some minor UI adjustments.

Also do we need to use coil for the final implementation? Since the images will come from core as a Bitmap, we could do it with simple `Image` swapping.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/formelement/AttachmentFormElement.kt
```diff
@@ -0,0 +1,448 @@
...
+
+@Composable
+private fun ThumbnailMenu(expanded: Boolean, onDismiss: () -> Unit = {}) {
+    MaterialTheme(shapes = MaterialTheme.shapes.copy(extraSmall = RoundedCornerShape(8.dp))) {
+        DropdownMenu(
+            expanded = expanded,
+            offset = DpOffset(15.dp, (-5).dp),
+            onDismissRequest = onDismiss
+        ) {
+            Row(
```
kaushikrw, line 264: material 3 has a `DropdownMenuItem` that can be used here.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/formelement/AttachmentFormElement.kt
```diff
@@ -146,14 +150,14 @@ private fun AttachmentFormElement(
...
-            modifier = Modifier.padding(horizontal = 10.dp, vertical = 5.dp)
+            modifier = Modifier.padding(horizontal = 10.dp, vertical = 10.dp)
         ) {
             AttachmentElementHeader(
                 title = state.title,
                 description = state.description,
-                keyword = state.keyword,
                 editable = state.editable
             )
+            Spacer(modifier = Modifier)
```
kaushikrw, line 160:
```suggestion
            Spacer(modifier = Modifier.height(10.dp))
```

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/formelement/AttachmentFormElement.kt
```diff
@@ -0,0 +1,458 @@
...
+@Composable
+private fun AttachmentElementHeader(
+    title: String,
+    description: String,
+    modifier: Modifier = Modifier,
+    showingDetails: Boolean = false,
+    editable: Boolean = true
+) {
+    Row(
+        modifier = modifier.height(84.dp),
```
kaushikrw, line 425:
```suggestion
        modifier = modifier.wrapContentHeight(),
```
`wrapContentHeight()` might be better here so it dynamically resizes the header height.

kaushikrw (approved):
LGTM. Approving with some minor comments.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/formelement/AttachmentFormElement.kt
sorenoid: now it looks correct to me. good tip 👍

PR #326, opened 2024-02-16
feature form recomposition when a new Feature is selected

sorenoid:
toolkit behavior change to make state object key on the API FeatureForm object in their `remember` functions.

app level changes to present a discard edits dialog when a new Feature is selected and the UIState is editing.

### DIFF
--- a/microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapViewModel.kt
+++ b/microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapViewModel.kt
@@ -205,8 +205,6 @@ class MapViewModel @Inject constructor(
                             }?.let {
                                 if (_uiState.value is UIState.Editing) {
                                     val currentState = _uiState.value as UIState.Editing
-                                    //set state to trigger DiscardEditsDialog with lambdas to discard edits and unselect the old feature and select the new one
-                                    // or just cancel and do nothing
                                     val newFeature = it as ArcGISFeature
                                     _uiState.value = UIState.Switching(
                                         oldState = currentState,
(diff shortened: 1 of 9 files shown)


microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapViewModel.kt
```diff
@@ -155,6 +163,23 @@ class MapViewModel @Inject constructor(
         )
         return Result.success(Unit)
     }
+    
+    fun selectNewFeature() {
+        require(_uiState.value is UIState.Switching)
```
kaushikrw, line 168:
```suggestion
        if(_uiState.value is UIState.Switching)
```
It might be better to use an `if` here since this can be invoked from UI.

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapViewModel.kt
```diff
@@ -155,6 +163,23 @@ class MapViewModel @Inject constructor(
...
+        currentState.oldState.featureForm.discardEdits()
+        val layer = currentState.oldState.featureForm.feature.featureTable?.layer as FeatureLayer
+        layer.clearSelection()
+        layer.selectFeature(currentState.newFeature)
+        _uiState.value =
+            UIState.Editing(featureForm = FeatureForm(currentState.newFeature, layer.featureFormDefinition!!))
+    }
+    
+    fun continueEditing() {
+        require(_uiState.value is UIState.Switching)
```
kaushikrw, line 179:
```suggestion
        if(_uiState.value is UIState.Switching)
```
Similarly, here too.

kaushikrw (changes requested):
Looks good! Just a minor comment.

kaushikrw (approved):

PR #329, opened 2024-02-16
cs/Move the time extent to the GeoProxy

changanxian:
Refactor the `timeExtent` property to the GeoProxy.


changanxian:
closed this as not necessary

PR #330, opened 2024-02-19
Composable MapView/SceneView: remember default parameters

hud10837:
Previously in the MapView/SceneView, default parameter values were not remembered. This could cause issues on recomposition as new instances of the objects were being created. For example, the SceneView's `cameraController` parameter was defaulted to `GlobeCameraController()`. When the SceneView recomposes, a new instance would be created and set on the underlying sceneView in the `update` block. If the SceneView was in the middle of navigating to a viewpoint, this would interrupt the navigation.

This PR updates the default parameters which should not be recalculated in recomposition when left undefined by the user. This includes types which have reference equality (ie. are equal if the instances are equal) and types that probably shouldn't be recalculated if unchanged (namely `PaddingValues` and `Instant`). It does not change, for example, `WrapAroundMode` (because the SDK types like this which act like enums are actually defined as objects) nor `Color` (because this is actually a value class).

Also updates the `SceneViewSetViewpointApp` to have a progress indicator like the MapView version, and fixes compile errors in the `SceneViewLightingOptionsApp`, which was still using `SceneViewpointOperation`.


toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -124,12 +124,12 @@ public fun MapView(
...
     geometryEditor: GeometryEditor? = null,
     mapViewProxy: MapViewProxy? = null,
-    mapViewInteractionOptions: MapViewInteractionOptions = MapViewInteractionOptions(),
-    viewLabelProperties: ViewLabelProperties = ViewLabelProperties(),
-    selectionProperties: SelectionProperties = SelectionProperties(),
-    insets: PaddingValues = PaddingValues(),
+    mapViewInteractionOptions: MapViewInteractionOptions = remember { MapViewInteractionOptions() },
+    viewLabelProperties: ViewLabelProperties = remember { ViewLabelProperties() },
+    selectionProperties: SelectionProperties = remember { SelectionProperties() },
+    insets: PaddingValues = remember { PaddingValues() },
```
gunt0001, line 130: Since `PaddingValues` is immutable, we can define a global static property for it, instead of using remember, i.e.:
```suggestion
    insets: PaddingValues = MapViewDefaults.NoPadding,
```
and elsewhere in MapView.kt:
```kotlin
object MapViewDefaults {
    public val NoPadding = PaddingValues()
   ...
}
```

See [guidelines](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-component-api-guidelines.md#default-expressions) for default expressions.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/SceneView.kt
```diff
@@ -118,20 +118,20 @@ public fun SceneView(
...
-    cameraController: CameraController = GlobeCameraController(),
+    cameraController: CameraController = remember { GlobeCameraController() },
     analysisOverlays: AnalysisOverlayCollection = rememberAnalysisOverlayCollection(),
     imageOverlays: ImageOverlayCollection = rememberImageOverlayCollection(),
     atmosphereEffect: AtmosphereEffect = AtmosphereEffect.HorizonOnly,
     timeExtent: TimeExtent? = null,
     onTimeExtentChanged: ((TimeExtent?) -> Unit)? = null,
     spaceEffect: SpaceEffect = SpaceEffect.Stars,
-    sunTime: Instant = Instant.parse("2000-09-22T12:00:00Z"),
+    sunTime: Instant = remember { Instant.parse("2000-09-22T12:00:00Z") },
```
gunt0001, line 134: As above with PaddingValues, `Instant` is immutable. so we could define a global static property for it, something like:
```suggestion
    sunTime: Instant = SceneViewDefaults.DefaultSunTime
```
and elsewhere in MapView.kt:
```kotlin
object SceneViewDefaults {
    public val DefaultSunTime: Instant = Instant.parse("2000-09-22T12:00:00Z")
   ...
}
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/SceneView.kt
```diff
@@ -118,20 +118,20 @@ public fun SceneView(
...
     sunLighting: LightingMode = LightingMode.NoLight,
     ambientLightColor: Color = Color(220, 220, 220, 255),
```
gunt0001, line 136: As above, I suggest to define a global static property for it:
```suggestion
    ambientLightColor: Color = SceneViewDefaults.DefaultAmbientLightColor,
```

gunt0001 (approved):
Nice 👍

changanxian:
LGTM 👍

changanxian (approved):

PR #331, opened 2024-02-21
Update map and scene parameters to be non-nullable

shubham7109:
PR to update `arcGISMap` & `arcGISScene` to be non-nullable parameters.


toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -116,7 +116,7 @@ import kotlinx.coroutines.launch
 @Composable
 public fun MapView(
     modifier: Modifier = Modifier,
-    arcGISMap: ArcGISMap? = null,
+    arcGISMap: ArcGISMap,
```
gunt0001, line 119: Non-defaulted parameters should go before default parameters, otherwise you will have to explicitly name the `arcGISMap` parameter in order to make use of the defualt for `modifier`, i.e. with what you have currently, a user is forced to make a call like `MapView(arcGISMap = myMap)` instead of `MapView(myMap)`.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/SceneView.kt
```diff
@@ -113,7 +113,7 @@ import java.time.Instant
 @Composable
 public fun SceneView(
     modifier: Modifier = Modifier,
-    arcGISScene: ArcGISScene? = null,
+    arcGISScene: ArcGISScene,
```
gunt0001, line 116: As above, the `arcGISScene` needs to be moved above `modifier`.

shubham7109:
Added the changes, ready for review @gunt0001

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -115,8 +115,8 @@ import kotlinx.coroutines.launch
  */
 @Composable
 public fun MapView(
+    arcGISMap: ArcGISMap,
     modifier: Modifier = Modifier,
-    arcGISMap: ArcGISMap? = null,
```
gunt0001, line 119: You also need to update the order of the parameters in the doc.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/SceneView.kt
```diff
@@ -112,8 +112,8 @@ import java.time.Instant
  */
 @Composable
 public fun SceneView(
+    arcGISScene: ArcGISScene,
     modifier: Modifier = Modifier,
-    arcGISScene: ArcGISScene? = null,
```
gunt0001, line 116: You also need to update the order of the parameters in the doc.

shubham7109:
Good catch! Re-ordered doc params.

gunt0001 (commented):
@shubham7109 - this looks good but I suggest we also update all the micro apps with regard to parameter order. For example the MapView function is currently being called like this:
```
MapView(
   arcGISMap = arcGISMap,
   mapViewProxy = mapViewProxy
 )
```
I suggest we update this to:
```
MapView(
   arcGISMap,
   mapViewProxy = mapViewProxy
 )
```

gunt0001 (approved):
👍

ruiqima (approved):
LGTM 👍

PR #335, opened 2024-02-21
`Forms`: Remove experimental `softwarekeyboardcontroller`

kaushikrw:
### Summary of changes

- Removed experimental api `LocalSoftwareKeyboardController` since it is not needed anymore.
- `FocusManager.clearFocus()` clears the focus in a state-less way and hides the keyboard without causing a extra recomposition. It will still recompose to remove focus but only once, unlike before, which was twice.


sorenoid (approved):
LGTM

PR #338, opened 2024-02-23
`Forms`: Consume `validationErrors` stateflow

kaushikrw:
### Summary of changes

- Consumes new sdk changes to use `validationErrors` state flow.
- Moved the logic to produce `ValidationErrorState` from the errors list into its own flow `mapValidationErrors`. This removes a lot of the parsing logic from individual field states. As a consequence of this, `validate()` can be removed as well as `observeProperties` and all the `FieldFormElement` flows can be safely collected in the initializer of `BaseFieldState`.
- Moved all the custom state flows creation into `Flows.kt` since its unlikely they will be in core.
- Removed validation on `isRequired` since the `validationErrors` will emit errors if that changes.
- combined `_attribute` and `validationErrors` flow into one flow collector so that the UI always has the latest `value` and `errors`.
- consumes `title` changed event
- fix `attachmentformelement` reference
- passes all tests

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
@@ -15,6 +15,7 @@ import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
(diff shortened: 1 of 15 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
```diff
@@ -116,6 +117,12 @@ public fun FeatureForm(
     }
 }
 
+@Composable
+private fun FeatureFormTitle(featureForm: FeatureForm) {
```
sorenoid, line 121: 👍 thanks

sorenoid (approved):
this looks really good. I like the consolidation.

PR #341, opened 2024-02-26
updgrade to sdk 4160

sorenoid:
upgrade SDK again to get changes to exclude unapproved API from generation


sorenoid:
not required.

PR #342, opened 2024-02-27
`Forms` :  Update UI behavior for non-nullable fields

kaushikrw:
### Summary of changes

- For non-nullable and nullable string backed fields, when the value is cleared, the value is set to `null` via `updateValue(null)`
- ComboBox fields that are _required_ will not have a _clear_ action
- Updated `ComboBoxTests` Test case 3.5
- All tests are passing

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/codedvalue/ComboBoxField.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/codedvalue/ComboBoxField.kt
@@ -138,7 +138,13 @@ internal fun ComboBoxField(
             )
         },
         interactionSource = interactionSource,
-        onFocusChange = state::onFocusChanged
+        onFocusChange = state::onFocusChanged,
+        trailingContent = if (isRequired) {
+            // if required then do not show a clear icon
+            {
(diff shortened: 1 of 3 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/codedvalue/ComboBoxField.kt
```diff
@@ -138,7 +138,13 @@ internal fun ComboBoxField(
             )
         },
         interactionSource = interactionSource,
-        onFocusChange = state::onFocusChanged
+        onFocusChange = state::onFocusChanged,
+        trailingContent = if (isRequired) {
```
sorenoid, line 142: seems like `trailingContent` should become an abstract function on BaseTextField and just provide the composable specific to each element, but that is for another PR 👍

sorenoid (approved):

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/components/codedvalue/ComboBoxField.kt
kaushikrw: I thought of a similar change, but the effect was far reaching. So it might be better to do as a separate issue indeed.

PR #343, opened 2024-02-28
Composable GeoViews: Viewpoint Persistence

hud10837:
Adds viewpoint persistence to the MapView and SceneView. Viewpoint persistence refers to restoring the viewpoint after the MapView or SceneView is destroyed (for example, during a configuration change). The methods for doing this are slightly different for the MapView and SceneView.

For the MapView, we add a `viewpointPersistence` parameter with an enum type for the user to decide whether they want the viewpoint persisted by center and scale, bounding geometry, or not at all. We need to give the user the option of preserving either viewpoint type, so this parameter is the simplest solution. Under the hood, we use `rememberSaveable` to keep track of the last known viewpoint and restore it when the MapView first enters the composition. We pass a custom `Saver` to `rememberSaveable` which converts the Viewpoint to json and back.

For the SceneView, we don't add any parameter, instead we always persist. If users request the option not to persist in the future, we could add that. It doesn't make too much sense in the SceneView to preserve either Center/Scale or BoundingGeometry because they can not define tilt/pitch/roll. Instead, we persist the `Camera.transformationMatrix` and restore that to get an exact camera position with minimal data.

Note this PR also moves all the viewpoint events into the same functions as the viewpoint persistence (rather than using eg. `MapViewEventHandler()`) in order to minimize the number of collectors on `GeoView.viewpointChanged`.


toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -413,6 +398,91 @@ private fun MapViewEventHandler(
...
+                currentVisibleAreaChanged?.invoke(
+                    mapView.visibleArea
+                        ?: throw IllegalStateException("MapView visible area should not be null")
+                )
+                var currentViewpoint =
+                    if (currentViewpointPersistence != ViewpointPersistence.None) {
+                        val viewpointType =
+                            if (currentViewpointPersistence == ViewpointPersistence.ByCenterAndScale) ViewpointType.CenterAndScale else ViewpointType.BoundingGeometry
+                        mapView.getCurrentViewpoint(viewpointType)
+                    } else null
```
gunt0001, line 448: Would a when statement make this more readable?
```suggestion
                var currentViewpoint = when (currentViewpointPersistence) {
                   is ViewpointPersistence.ByCenterAndScale -> mapView.getCurrentViewpoint(ViewpointType.CenterAndScale)
                   is ViewpointType.BoundingGeometry -> mapView.getCurrentViewpoint(ViewpointType.BoundingGeometry)
                   else -> null
                }
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -413,6 +398,91 @@ private fun MapViewEventHandler(
...
+                currentViewpoint?.let { persistedViewpoint = it }
```
gunt0001, line 449: I think you need to set the `persistedViewpoint` here regardless of whether `currentViewpoint` is null or not. In fact, you could assign the result of the `when` statement suggested above directly to `persistedViewpoint`.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -413,6 +398,91 @@ private fun MapViewEventHandler(
...
+        )
+    ) {
+        mutableStateOf<Viewpoint?>(null)
+    }
+
+    LaunchedEffect(Unit) {
+        persistedViewpoint?.let { mapView.setViewpoint(it) }
+        launch {
+            mapView.viewpointChanged.collect {
+                currentVisibleAreaChanged?.invoke(
```
gunt0001, line 439: Could you move this logic down to where you also handle view point changed events for scale/bounding geometry viewpoints.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -413,6 +398,91 @@ private fun MapViewEventHandler(
...
+                    if (currentViewpoint?.viewpointType != ViewpointType.BoundingGeometry) {
+                        currentViewpoint =
+                            mapView.getCurrentViewpoint(ViewpointType.BoundingGeometry)
+                    }
+                    currentViewpoint?.let(callback)
+                }
+            }
+        }
+        launch {
+            snapshotFlow { currentViewpointPersistence }
```
gunt0001, line 466: I assume this is needed if `currentViewpointPersistence` changes during recomposition?

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/SceneView.kt
```diff
@@ -389,6 +373,97 @@ private fun SceneViewEventHandler(
...
+                // This is the default SceneView matrix
+                TransformationMatrix.createWithQuaternionAndTranslation(
+                    0.0,
+                    -0.0,
+                    -0.0,
+                    1.0,
+                    0.0,
+                    0.0,
+                    5.3969112064275116E7
+                )
```
gunt0001, line 439: Not sure we can do that. It might override the initial viewpoint specified on a scene. Instead, we may need to default to `null`.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/SceneView.kt
```diff
@@ -389,6 +373,97 @@ private fun SceneViewEventHandler(
...
+        )
+    }
+
+    LaunchedEffect(Unit) {
+        sceneView.setViewpointCamera(persistedCamera)
+        launch {
+            sceneView.viewpointChanged.collect {
+                val currentViewpointCamera = sceneView.getCurrentViewpointCamera()
+                currentOnCurrentViewpointCameraChanged?.invoke(currentViewpointCamera)
+                persistedCamera = currentViewpointCamera
```
gunt0001, line 450: Best to move this to the top of the `collect` block.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/ViewpointPersistence.kt
```diff
@@ -0,0 +1,37 @@
...
+ * or process recreation, for example, when the device is rotated or when the app is sent to the background
+ * and then brought back to the foreground.
+ *
+ * - None: No viewpoint persistence is applied. The viewpoint will not be saved or restored.
+ * - ByCenterAndScale: The viewpoint is persisted by its center and scale.
+ * - ByBoundingGeometry: The viewpoint is persisted by its bounding geometry.
+ *
+ * @since 200.4.0
+ */
+public enum class ViewpointPersistence {
```
gunt0001, line 33: Should we name it `MapViewpointPersistence` so we can introduce a `SceneViewpointPersistence` in the future if needed?

hud10837:
Thanks @gunt0001 , I believe I have responded to all of your comments

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -413,6 +400,115 @@ private fun MapViewEventHandler(
...
+                                center = center as Point,
+                                scale = viewpoint.targetScale,
+                                rotation = viewpoint.rotation
+                            )
+                        }
+                } else viewpoint
+                normalized?.toJson()
+            },
+            restore = {
+                mutableStateOf(Viewpoint.fromJsonOrNull(it))
```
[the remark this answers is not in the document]
hud10837: If you return `null` from the `save` lambda, then `restore` is not called. From the doc on `save`:

```
Defines how to convert the value into a saveable one. If null is returned the value will not be saved.
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -436,6 +521,14 @@ public inline fun rememberLocationDisplay(
     }
 }
 
+private fun MapView.getViewpointByPersistence(viewpointPersistence: ViewpointPersistence): Viewpoint? {
+    return when (viewpointPersistence) {
+        is ViewpointPersistence.None -> null
+        is ViewpointPersistence.ByCenterAndScale -> getCurrentViewpoint(ViewpointType.CenterAndScale)
+        is ViewpointPersistence.ByBoundingGeometry -> getCurrentViewpoint(ViewpointType.BoundingGeometry)
+    }
+}
```
gunt0001, line 530: Nitpick - can use functional style & missing doc.
```suggestion
private fun MapView.getViewpointByPersistence(viewpointPersistence: ViewpointPersistence): Viewpoint? =
    when (viewpointPersistence) {
        is ViewpointPersistence.None -> null
        is ViewpointPersistence.ByCenterAndScale -> getCurrentViewpoint(ViewpointType.CenterAndScale)
        is ViewpointPersistence.ByBoundingGeometry -> getCurrentViewpoint(ViewpointType.BoundingGeometry)
    }
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/ViewpointPersistence.kt
```diff
@@ -0,0 +1,68 @@
...
+ * and then brought back to the foreground.
+ *
+ * - None: No viewpoint persistence is applied. The viewpoint will not be saved or restored.
+ * - ByCenterAndScale: The viewpoint is persisted by its center and scale.
+ * - ByBoundingGeometry: The viewpoint is persisted by its bounding geometry.
+ *
+ * @since 200.4.0
+ */
+public sealed class ViewpointPersistence {
+    public object None : ViewpointPersistence()
```
gunt0001, line 34: Missing doc for `None`

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/ViewpointPersistence.kt
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Enum class representing the different types of viewpoint persistence on a composable [MapView].
+ *
+ * Viewpoint persistence determines how the viewpoint of a MapView is saved and restored across activity
+ * or process recreation, for example, when the device is rotated or when the app is sent to the background
+ * and then brought back to the foreground.
+ *
+ * - None: No viewpoint persistence is applied. The viewpoint will not be saved or restored.
+ * - ByCenterAndScale: The viewpoint is persisted by its center and scale.
+ * - ByBoundingGeometry: The viewpoint is persisted by its bounding geometry.
```
gunt0001, line 29: These bullet points are somewhat redundant since you repeat what is already documented on the individual enum classes.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -132,6 +141,7 @@ public fun MapView(
     grid: Grid? = null,
     backgroundGrid: BackgroundGrid = remember { BackgroundGrid() },
     wrapAroundMode: WrapAroundMode = WrapAroundMode.EnabledWhenSupported,
+    viewpointPersistence: ViewpointPersistence = MapViewDefaults.DefaultViewpointPersistence,
```
gunt0001, line 144: Should we move this parameter higher up, just below `onVisibleAreaChanged`? I feel it belongs to the viewpoint related lambdas...

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -413,6 +400,104 @@ private fun MapViewEventHandler(
...
+                        ?.let { normalizedGeometry ->
+                            if (viewpoint.viewpointType == ViewpointType.CenterAndScale) {
+                                Viewpoint(
+                                    center = normalizedGeometry as Point,
+                                    scale = viewpoint.targetScale,
+                                    rotation = viewpoint.rotation
+                                )
+                            } else {
+                                Viewpoint(
+                                    boundingGeometry = normalizedGeometry as Polygon,
```
gunt0001, line 445: This cast should not be necessary, potentially could cause a crash:
```suggestion
                                    boundingGeometry = normalizedGeometry,
```

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -413,6 +400,104 @@ private fun MapViewEventHandler(
...
+
+                // If the spatial reference is pannable or geographic, normalize the viewpoint's center
+                // If we don't do this, the viewpoint will not be restored correctly after rotation
+                // if the viewpoint has crossed the central meridian. However normalization can
+                // only be done with pannable or geographic spatial references.
+                val sr = viewpoint.targetGeometry.spatialReference
+                val viewpointToPersist = if (sr?.isPannable == true || sr?.isGeographic == true) {
+                    GeometryEngine.normalizeCentralMeridian(viewpoint.targetGeometry)
+                        ?.let { normalizedGeometry ->
+                            if (viewpoint.viewpointType == ViewpointType.CenterAndScale) {
```
gunt0001, line 437: Better use `when` instead of `if`, just in case the viewpoint type enum is enhanced in the future, then the compiler would catch that. In addition, the code may become more readable by factoring all of this out into an extension function, something like:
```
private fun Viewpoint.tryNormalize(): Viewpoint? {
   val sr = viewpoint.targetGeometry.spatialReference
   ...
}
```
Then the calling code would be as simple as:
```
val viewpointToPersist = viewpoint.tryNormalize() ?: viewpoint
```

microapps/TemplateApp/app/src/main/java/com/arcgismaps/toolkit/templateapp/screens/MainScreen.kt
hud10837, line 1: This app was not compiling because it was still using viewpoint operations, so updating it here.

hud10837:
Thanks @gunt0001 ready for re-review

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -413,6 +400,132 @@ private fun MapViewEventHandler(
...
+ */
+private fun tryNormalizeViewpoint(viewpoint: Viewpoint): Viewpoint {
+    // Normalization should only be done with pannable or geographic spatial references.
+    val sr = viewpoint.targetGeometry.spatialReference ?: return viewpoint
+    if (!sr.isPannable && !sr.isGeographic) return viewpoint
+    // if normalization fails, just return the viewpoint as is
+    val normalizedGeometry =
+        GeometryEngine.normalizeCentralMeridian(viewpoint.targetGeometry) ?: return viewpoint
+
+    val viewpointToPersist = when (viewpoint.viewpointType) {
```
gunt0001, line 498: "persist" doesn't make sense in the context of this function, I suggest:
```suggestion
    val normalizedViewpoint = when (viewpoint.viewpointType) {
```

gunt0001 (approved):
@hud10837 - looks great, one minor comment.

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/ViewpointPersistence.kt
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @since 200.4.0
+     */
+    public class ByCenterAndScale : ViewpointPersistence() {
+
+        // Note: ByCenterAndScale and ByBoundingGeometry could have been defined as singletons (object) but we
+        // want to keep the possibility open to add instance state (properties) to these classes in the future,
+        // thus we had to declare them as classes. This meant we had to override hashCode and equals in order to
+        // achieve the same equality behaviour as a singleton would do.
+        override fun hashCode(): Int {
```
changanxian, line 49: slightly better with ` override fun hashCode(): Int = 1 `

changanxian: we may not have such a case, but the constant can lead to poor performance in data structures like hash-based collections (e.g., HashMap)

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/ViewpointPersistence.kt
```diff
@@ -0,0 +1,70 @@
...
+            return 1
+        }
+        override fun equals(other: Any?): Boolean {
```
changanxian, line 52: slight better with `   override fun equals(other: Any?): Boolean = ( other is ByCenterAndScale )`

toolkit/geo-compose/src/main/java/com/arcgismaps/toolkit/geocompose/MapView.kt
```diff
@@ -413,6 +400,132 @@ private fun MapViewEventHandler(
...
+/**
+ * Normalizes the viewpoint's target geometry.
+ *
+ * If the spatial reference is null or not pannable or geographic, the viewpoint is returned as is.
+ * Otherwise, the viewpoint's target geometry is normalized and a new viewpoint is created with the
+ * normalized geometry.
+ *
+ * @since 200.4.0
+ */
+private fun tryNormalizeViewpoint(viewpoint: Viewpoint): Viewpoint {
```
changanxian, line 490: feel slightly better to change the function to `private fun Viewpoint.normalize(): Viewpoint`

changanxian (approved):
Looks good! A few small changes are up to you to take.

hud10837:
Thank you @changanxian and @gunt0001 for the reviews!

PR #345, opened 2024-02-29
`Forms` api backwards compatibility validation

sorenoid:
Adds the use of the [Kotlin-JVM binary compatibility validator gradle plugin](https://github.com/Kotlin/binary-compatibility-validator)

adds the `api` file which represents the FeatureForms public API
support for generating the `api` file by running gradle task `apiDump`
support for checking backwards compatibility violations by running gradle task `apiCheck`

### DIFF
--- a/toolkit/featureforms/build.gradle.kts
+++ b/toolkit/featureforms/build.gradle.kts
@@ -53,6 +53,9 @@ android {
 
 apiValidation {
     ignoredClasses.add("com.arcgismaps.toolkit.featureforms.BuildConfig")
+    // todo: remove when this is resolved https://github.com/Kotlin/binary-compatibility-validator/issues/74
+    // compose compiler generates public singletons for internal compose functions. this may be resolved in the compose
+    // compiler.
     val composableSingletons = listOf(
         "com.arcgismaps.toolkit.featureforms.components.base.ComposableSingletons\$BaseTextFieldKt",
         "com.arcgismaps.toolkit.featureforms.components.codedvalue.ComposableSingletons\$ComboBoxFieldKt",
--- a/toolkit/featureforms/build.gradle.kts
+++ b/toolkit/featureforms/build.gradle.kts
@@ -53,6 +53,19 @@ android {
 
 apiValidation {
     ignoredClasses.add("com.arcgismaps.toolkit.featureforms.BuildConfig")
+    val composableSingletons = listOf(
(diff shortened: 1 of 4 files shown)


toolkit/featureforms/build.gradle.kts
```diff
@@ -50,6 +51,27 @@ android {
...
+        "com.arcgismaps.toolkit.featureforms.components.codedvalue.ComposableSingletons\$RadioButtonFieldKt",
+        "com.arcgismaps.toolkit.featureforms.components.datetime.ComposableSingletons\$DateTimeFieldKt",
+        "com.arcgismaps.toolkit.featureforms.components.datetime.picker.ComposableSingletons\$DateTimePickerKt",
+        "com.arcgismaps.toolkit.featureforms.components.datetime.picker.date.ComposableSingletons\$DatePickerKt",
+        "com.arcgismaps.toolkit.featureforms.components.datetime.picker.time.ComposableSingletons\$TimePickerKt",
+        "com.arcgismaps.toolkit.featureforms.components.formelement.ComposableSingletons\$AttachmentFormElementKt",
+        "com.arcgismaps.toolkit.featureforms.components.formelement.ComposableSingletons\$GroupElementKt"
+    )
+    
+    ignoredClasses.addAll(composableSingletons)
```
kaushikrw, line 71:
```suggestion
    ignoredPackages.add("com.arcgismaps.toolkit.featureforms.components")
```
I think the entire package can be ignored.

kaushikrw (changes requested):
This plugin seems very helpful for the toolkit, especially as our compose apis become large. 

I was however wondering, what the workflow would be if and when this gets integrated into Github Checks. Perhaps we can document it into a README, how to verify and update the api compatibility.

gunt0001 (approved):
Looks good 👍

toolkit/featureforms/build.gradle.kts
sorenoid: > I think the entire package can be ignored. It might be a little tedious to keep track of all the internal compose functions.

@kaushikrw I'd like to discuss this point a bit offline.

kaushikrw: As discussed, this can go in now and we will later move all the internal components into an `internal` package for feature forms and ignore that package.

kaushikrw (approved):
LGTM

PR #346, opened 2024-03-02
`Forms` : Read only fields re-design

kaushikrw:
### Summary of changes

- Updated read-only fields design. This is implemented by introducing a `ReadOnlyTextField` that is used when the `BaseTextField` is called with `isEditable=false`
- Showing supportingText, character count and `*` on required field labels is moved into the `BaseTextField` to avoid code duplication and since it is common behavior for all field elements that use the `BaseTextField`.
- `BaseTextField` now takes an `isError` parameter that uses the material 3 spec for showing an error state.
- A `Divider` is now shown between form elements.
- Optimizations to the `trailingContent` of the `BaseTextField`.
- If no `supportingText` is provided and `showCharacterCount` is false, then no UI node exists in the composition. Otherwise if an empty text is created, it takes up a blank space which doesn't look consistent.
- Fixed tests to assert the above statement.
- All tests are passing.

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
@@ -11,11 +11,9 @@ import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.material3.Divider
 import androidx.compose.material3.LinearProgressIndicator
-import androidx.compose.material3.LocalTextStyle
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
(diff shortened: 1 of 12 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
```diff
@@ -173,6 +173,7 @@ private fun FeatureFormBody(
                             // other form elements are not created
                         }
                     }
+                    Divider()
```
sorenoid, line 176: It might be good to only surround read only fields with dividers, but it is hard to tell without looking at it and it would require a small amount of brain surgery here to see it. If we get feedback, let's look at it.

sorenoid (approved):
👍 this re-design makes a lot of sense visually and internal-logic wise. All in all a very good change for both of these reasons.

I have one lingering question if the dividers should only surround non-editable fields. When they surround editable FieldFormElements which display as OutlinedTextFields, and around GroupElements which display as Cards they seem redundant. The same will be true of AttachmentFormElements which will display as Cards.

Let's see if others provide feedback.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
kaushikrw: I agree, lets get some feedback and iterate.

kaushikrw:
Thanks @sorenoid.

I think the `GroupElement` might need to change to accommodate the dividers like in the iOS design. `AttachmentFormElement` might also follow the group changes.

Lets gather some feedback and improve the design.

Merging this for now!

PR #350, opened 2024-03-06
create an internal subpackage containing all non public packages.

sorenoid:
create an internal package containing all internal packages
update the `apiValidation` gradle configuration to reflect the changed package structure.

verified tests and microapp still pass/function.
verified no text references in comments or md files.
ran apiCheck successfully


kaushikrw (approved):
LGTM!

PR #351, opened 2024-03-06
replace clear button on required date time fields

sorenoid:
when a date time element is required, do not show a clear button. Instead show the edit calendar button (the non-editable calendar button is no longer needed since we show read only fields with no trailing icon.)

editable, required, empty or non-empty
show Icons.Rounded.EditCalendar

editable, not-required, empty
show Icons.Rounded.EditCalendar

editable, non-required, non-empty
show clear

non-editable 
show no icon


kaushikrw (approved):
Looks good!

PR #353, opened 2024-03-11
FloorFliter: silence exception when floorManager is not set

ruiqima:
The scope of this issue is to make sure that an exception is not thrown and the app does not crash if the GeoModel used to create it doesn't have the floorManager. The failure should eventually be handled by a callback, but that will be handled in a different issue.

#### Changes

- Replace error throwing with `return`.


mikewilburn:
@ruiqima we will likely merge our feature branch into `v.next` soon. Can this PR target `v.next` instead?

puneet-pdx:
@ruiqima the change looks good, but as @mikewilburn said you will probably have to create another PR that is based on `v.next` to have a clean commit history.

puneet-pdx (commented):
@ruiqima the change looks good, but as @mikewilburn said you will probably have to create another PR that is based on `v.next` to have a clean commit history.

PR #357, opened 2024-03-15
don't normalize viewpoints before persisting

hud10837:
This code that normalizes viewpoints before persisting them was in place to work around a core issue with setting the viewpoint during mapview initialization. The bug was fixed so this code can now be removed.


gunt0001 (approved):
👍

puneet-pdx (approved):

PR #358, opened 2024-03-15
don't persist null current viewpoints

hud10837:
Checks if a viewpoint is null before making it the `persistedViewpoint`

When the view-based `MapView` is initializing, it emits a `viewpointChanged` event with a `null` viewpoint. This will get persisted, so if you set the viewpoint, rotate the device once, it will be the expected viewpoint that is persisted, but if you then don't interact with the mapview and rotate again, the last viewpoint changed event would have been a null value and the mapview displays the max extent.


gunt0001 (approved):
@hud10837 - looks good. Make sure to test the following:
- MapView is in composition and a persisted viewpoint has been initialized
- ViewpointPersistence is changed to `None`, causing recomposition
- persisted viewpoint should be cleared

shubham7109 (approved):
Tested this out, and works as expected. Looks good to me

hud10837:
Thanks @gunt0001 @shubham7109

PR #359, opened 2024-03-15
FloorFliter: silence exception when floorManager is not set, targeting v.next

ruiqima:
The scope of this issue is to make sure that an exception is not thrown and the app does not crash if the GeoModel used to create it doesn't have the floorManager. The failure should eventually be handled by a callback, but that will be handled in a different issue. Related PR: #353 

#### Changes

- Replace error throwing with `return`.


puneet-pdx (approved):
@ruiqima looks good

shubham7109 (approved):
Looks good 👍

PR #362, opened 2024-03-19
`Forms`: Add theming support

kaushikrw:
### Summary of changes

- Added theming support for the FeatureForm component that includes customizing the ColorScheme and Typography for most of the fields.
- This theming support is built on top of Material3 to provide default values.
- Added tests to ensure the color scheme and typography are being passed down and used by the appropriate field types.
- Updates binary API

### Details

It provides a way to customize the color scheme and typography using two classes.
  - `FeatureFormColorScheme`
      - EditableTextFieldColors
      - ReadOnlyTextFieldColors
      - GroupElementColors
      - RadioButtonFieldColors 
  - `FeatureFormTypoGraphy`
      - EditableTextFieldTypography
      - ReadOnlyTextFieldTypography
      - GroupElementTypography
      - RadioButtonFieldTypography

These individual classes allow a user to fully customize the color and typography for all the fields independently. 

These properties are passed down the compose tree using the `LocalColorScheme` and `LocalLocalTypography` composition local.

All these classes provide a `createDefaults()` factory method that can only be called from a `Composable` context. The `createDefaults()` method hydrates the individual properties of these classes from `MaterialTheme`. It also provides parameters with default values, so that any developer can quickly glance at what `MaterialTheme` values are being set as defaults.

Due to this design, the `FeatureFormColorScheme` and `FeatureFormTypoGraphy` can only be created from a `Composable` context, since it relies on `MaterialTheme` for default values. Hence, as a convenience, an object class `DefaultThemeTokens` is introduced which provides a default `colorScheme` and `typography` based on reasonable hardcoded values. These hardcoded defaults are adapted from the `Material3.lightColorScheme` and `MaterialTheme.Typography`.

Furthermore, the `DefaultThemeTokens` are also useful in providing a good default value for the `LocalColorScheme` and `LocalLocalTypography`  composition local. This is particularly helpful for compose previews.


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/theme/DefaultThemeTokens.kt
```diff
@@ -0,0 +1,224 @@
...
+    val Secondary = PaletteTokens.Secondary40
+    val SecondaryContainer = PaletteTokens.Secondary90
+    val Surface = PaletteTokens.Neutral99
+    val SurfaceTint = Primary
+    val SurfaceVariant = PaletteTokens.NeutralVariant90
+    val Tertiary = PaletteTokens.Tertiary40
+    val TertiaryContainer = PaletteTokens.Tertiary90
+}
+
+private object PaletteTokens {
```
sorenoid, line 197: I think this needs to check `isSystemInDarkTheme()` and offer dark mode alternatives/

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/theme/FeatureFormTheme.kt
```diff
@@ -0,0 +1,707 @@
...
+ * defaults are populated using [MaterialTheme].
+ *
+ * Any nested elements within a GroupFormElement will also use the same color scheme specified as
+ * part of this class.
+ *
+ *
+ * Use [FeatureFormColorScheme.createDefaults] to create a new instance with the default values.
+ */
+@Immutable
+public data class FeatureFormColorScheme internal constructor(
```
sorenoid, line 122: All of the public types, properties and function need to have `since 200.5.0` tags.

sorenoid (changes requested):
@kaushikrw this looks great and fills a valuable need for FeatureForms. I have some very high level suggestions
* provide a dark mode color palette that is applied automatically when in dark mode
* add since tags to all public API
* provide a simple theme for the FeatureForm itself in addition to the individual components of the form.
    * this theme could be used for the body of GroupElements as well, perhaps with different defaults.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/base/BaseTextField.kt
```diff
@@ -248,24 +257,92 @@ private fun ReadOnlyTextField(
...
 
+@Composable
+internal fun EditableTextFieldColors.toTextFieldColors(
+    textIsEmpty: Boolean,
+    placeHolderIsEmpty: Boolean
+): TextFieldColors {
+    // transform placeholder colors into text colors
+    val transformPlaceHolderColors = textIsEmpty && !placeHolderIsEmpty
+    val transformedFocusedTextColor = if (transformPlaceHolderColors) {
+        // if placeholder is visible, make it lighter than the actual input text color
```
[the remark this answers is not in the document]
kaushikrw: I don't think we ever created an issue, but this behavior also exists in the pre-custom theme system.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/theme/DefaultThemeTokens.kt
kaushikrw: I'm not sure this is needed. The `DefaultThemeTokens` only needs to provide default values for `LocalColorScheme` and `LocalTypography`. These defaults are only utilized in internal compose previews. During runtime the actual values are always taken from `MaterialTheme`.

kaushikrw:
@sorenoid I've updated this based on our new design standard and addressed some of the feedback. Please have another look!

sorenoid (approved):
LGTM!

kaushikrw:
Thanks @sorenoid, merging.

PR #366, opened 2024-03-20
`Forms`: Updated radio button tests

kaushikrw:
### Summary of changes

- Updates the `testRadioButtonSelection` test to not assert a no value label, since the field is now required because of the `ArcGISFeatureTable.typeIdField`.
- Adds 3 more tests based on the common-toolkit test design.


sorenoid (approved):
👍

PR #367, opened 2024-03-21
add groups to final error validation reporting in the app

sorenoid:
look over field form elements nested in groups when accumulating errors for the apply edits dialog.


kaushikrw (approved):
LGTM.

PR #368, opened 2024-03-21
FeatureForm composable KDoc edit

sorenoid:
Just a bit of cleanup based on suggestions from @gunt0001

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
@@ -93,8 +93,8 @@ public sealed class ValidationErrorVisibility {
 
 /**
  * A composable Form toolkit component that enables users to edit field values of features in a
- * layer using forms that have been configured externally (using either in the the Web Map Viewer
- * or the Fields Maps web app).
+ * layer using forms that have been configured externally. Forms may be configured in the Web Map Viewer
+ * or Fields Maps Designer).
  *
  * @param featureForm The [FeatureForm] configuration.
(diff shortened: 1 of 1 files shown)


kaushikrw (approved):

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
```diff
@@ -93,8 +93,8 @@ public sealed class ValidationErrorVisibility {
 
 /**
  * A composable Form toolkit component that enables users to edit field values of features in a
- * layer using forms that have been configured externally (using either in the the Web Map Viewer
- * or the Fields Maps web app).
+ * layer using forms that have been configured externally. Forms may be configured in the Web Map Viewer
+ * or Fields Maps Designer).
```
gunt0001, line 97: Would linking make sense here? I'm ok if you prefer not to for some reason...
```suggestion
 * layer using forms that have been configured externally. Forms may be configured in the [Web Map Viewer](https://www.arcgis.com/home/webmap/viewer.html)
 * or [Fields Maps Designer](https://www.arcgis.com/apps/fieldmaps/)).
```

gunt0001: Can forms be configured in ArcGIS Pro? If so, I would add that as well.

gunt0001 (approved):
@sorenoid - some suggestion/question.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
sorenoid: nope

PR #369, opened 2024-03-22
`Forms` : Optimize FeatureForm

kaushikrw:
### Summary

- Adds `@Stable` and `@Immutable` annotations to classes that are indeed stable and immutable so that compose compiler can enable smart recompositions by making the Composables that use these classes skippable.
- `FeatureForm` composable now uses an internal `StateData` class which is marked as `Immutable` as a internal mechanism to enable smart recompositions.
- Updated `MapScreen` to remember the form inside the `AnimatedVisibility` so that exit animation is played correctly.


sorenoid (approved):
💪 this is great. thanks for doing this in a way where the API remains the same.

sorenoid (approved):

kaushikrw:
Thanks @sorenoid, merging.

PR #373, opened 2024-03-26
Format screenshots

shubham7109:
### Description

PR to format all the screenshots used in the readmes to allow for a max image height of `800px`


hud10837 (commented):
Thanks for making this change. I notice that in some readmes we place the screenshot in a table, while in others we leave them inline. Could we normalize this with this PR? I think except for the Floor Filter microapp which uses a lot of screenshots, it makes sense to have them outside a table.

shubham7109:
@hud10837 Good suggestion, I updated the screenshots to have them outside a table.

hud10837 (approved):
Thanks @shubham7109

changanxian (approved):
👍

PR #375, opened 2024-03-27
Makes UsernamePassword and ServerTrust prompt scrollable so that the content is displayed correctly

eri9000:


hud10837 (commented):
@eri9000 I think this looks good, but I also think we should reduce the font size of the prompt, which is quite large. Instead of `headlineSmall`, maybe we can use `titleLarge` or `titleSmall`?

hud10837 (approved):
Thanks for that change 👍

puneet-pdx (approved):
@eri9000 LGTM 👍🏽

PR #377, opened 2024-03-28
agree with exception name change

sorenoid:
There's a breaking change to the name of a FeatureFormValidationException in SDK build 200.4.0 as of 4191.

This will accommodate the change. Do not merge until after 4191 is built!


kaushikrw (approved):
LGTM

PR #378, opened 2024-03-28
use geo-compose in FeatureFormsApp

sorenoid:
remove some unnecessary coroutine logic and nullability in MapViewModel too.


microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/di/MapViewModule.kt
```diff
@@ -0,0 +1,48 @@
...
+import javax.inject.Qualifier
+import javax.inject.Singleton
+
+
+/**
+ * Provide an annotation to inject the MapViewProxy
+ */
+@Qualifier
+@Retention(AnnotationRetention.SOURCE)
+annotation class MapViewProxy
```
kaushikrw, line 32: Why does the `MapViewProxy` need to be injected via Hilt? I doesn't need to be a singleton either. I think this can be created directly in the `MapViewModel`.

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapViewModel.kt
```diff
@@ -94,23 +96,22 @@ data class ErrorInfo(val fieldName: String, val error: FeatureFormValidationExce
...
 class MapViewModel @Inject constructor(
     savedStateHandle: SavedStateHandle,
-    private val portalItemRepository: PortalItemRepository
-) : ViewModel(),
-    MapInterface by MapInterfaceImpl(ArcGISMap()) {
+    portalItemRepository: PortalItemRepository,
+    val proxy: MapViewProxy,
+    private val application: Application,
+    @ApplicationScope private val scope: CoroutineScope
+) : ViewModel() {
```
kaushikrw, line 103:
```suggestion
    application: Application,
    @ApplicationScope private val scope: CoroutineScope
) : AndroidViewModel(application) {
```
`AndroidViewModel` can be used here just to make it more explicit.

sorenoid: I had to add a base class which takes only the singular application parameter to do this.

kaushikrw (approved):
Looks good!

sorenoid:
thanks @kaushikrw

PR #379, opened 2024-03-28
Moves rememberLauncherForActivityResult outside of conditional

eri9000:
Fixes following issue:
When the OAuth Sign in page is rotated , the activity does not complete the result contract and the activity hangs indefinitely.


hud10837 (approved):
This fix is working for me. It looks like in the last commit that went into this file, the launcher definition was moved into the else statement as a cleanup but this was not properly tested. This was probably done as an optimization (so a launcher is not created when `onPendingOAuthUserSignIn` is null) but I think it's better to have the fix than an optimization 😅

changanxian (approved):
The changes look good.

PR #382, opened 2024-04-02
`Forms`: fix title padding

kaushikrw:
- Adds a horizontal padding for the `FeatureFormTitle`.
- Use `MaterialTheme.typography` for `FeatureFormTitle`

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
@@ -32,6 +32,7 @@ import androidx.compose.material3.LinearProgressIndicator
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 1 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
```diff
@@ -169,9 +170,18 @@ internal fun FeatureForm(
...
+    Text(
+        text = title,
+        style = TextStyle(fontWeight = FontWeight.Bold),
+        modifier = modifier
+    )
+    DisposableEffect(key1 =) {
+        onDispose {
+
+        }
+    }
```
sorenoid, line 184: this looks unintentional, and like it doesn't compile?

sorenoid (changes requested):
I think there is some unintentional code in your commit?

kaushikrw:
@sorenoid fixed it. Also noticed that the title wasn't using material theme typography which is addressed in this PR.

sorenoid (approved):

PR #383, opened 2024-04-02
Fixes configuration change bug when MainActivity is set up to intercept the redirect URI

eri9000:
Fixes bug:
If the device is rotated when the OAuth login page is being displayed and a subsequent login attempt is tried then the OAuth page will not get launched.


changanxian (approved):
LGTM

shubham7109 (approved):
Cannot repro the bug with this fix. Looks good 👍

PR #387, opened 2024-04-02
Fix layout issues

puneet-pdx:
- Fixes the layout issue with Authenticator and FloorFilter because of long strings in different locales.
- adds context to the strings whose translation have caused layout issues for future reference.

### DIFF
--- a/toolkit/authentication/src/main/res/values/strings.xml
+++ b/toolkit/authentication/src/main/res/values/strings.xml
@@ -18,8 +18,9 @@
 
 <resources>
     <string name="server_trust_message">The certificate provided by %1$s is not signed by a trusted authority.</string>
-    <!-- This text is on a button that can fit 24 chars -->
+    <!-- This text is on a button that can fit 12 chars -->
     <string name="cancel">Cancel</string>
+    <!-- This text is on a button that can fit 40 chars -->
     <string name="allow_connection">Dangerous: Allow Connection</string>
     <string name="username_password_login_message">Please login to %1$s</string>
     <!-- This text is a label on a textfield that can fit 23 chars -->
--- a/toolkit/authentication/src/main/res/values/strings.xml
+++ b/toolkit/authentication/src/main/res/values/strings.xml
@@ -18,10 +18,14 @@
 
 <resources>
     <string name="server_trust_message">The certificate provided by %1$s is not signed by a trusted authority.</string>
+    <!-- This text is on a button that can fit 24 chars -->
     <string name="cancel">Cancel</string>
     <string name="allow_connection">Dangerous: Allow Connection</string>
     <string name="username_password_login_message">Please login to %1$s</string>
+    <!-- This text is a label on a textfield that can fit 23 chars -->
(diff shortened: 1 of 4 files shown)


toolkit/indoors/src/main/res/values/strings.xml
```diff
@@ -2,7 +2,9 @@
 <resources>
     <string name="floor_filter_view_filter_hint_sites">Filter sites&#8230;</string>
     <string name="floor_filter_view_filter_hint_facilities">Filter facilities&#8230;</string>
+    <!-- This text is a label on a textfield that can fit 23 chars -->
```
hud10837, line 5: Is this true if we reduce the text size above 23? Surely it can then fit more

toolkit/authentication/src/main/res/values/strings.xml
```diff
@@ -18,10 +18,14 @@
 
 <resources>
     <string name="server_trust_message">The certificate provided by %1$s is not signed by a trusted authority.</string>
+    <!-- This text is on a button that can fit 24 chars -->
```
hud10837, line 21: these are good comments to have. Maybe we should consider having them on the `cancel` and `allow_connection` buttons as well. Did you test the server trust authenticator at all?

hud10837 (commented):
Looks good, a couple thoughts

toolkit/authentication/src/main/res/values/strings.xml
hud10837: Server trust in Ukrainian looks okay, so we can probably ignore it.

image

hud10837 (approved):
Since server trust looks okay with long text, I think this can move forward 👍

toolkit/indoors/src/main/res/values/strings.xml
puneet-pdx: I saw weird behavior for text around 24-25 characters where it would either not show the last character or wrap it on the next line. At this font size it can only fit 23 characters on a single line. When the text wraps on to the 2nd line that's when this header starts to look bad. So I capped it at 23 characters.

shubham7109 (approved):
Tested by changing device locale to `Vietnamese` which seems to be the longest translation text among the others. Don't see any layout issues. Looks good 👍 

One question, I noticed if I were to login without typing any credentials the error message displayed will always be in English. Seems like this string is hardcoded in `AuthenticatorState.kt`. Should this also be translated as part of the Authenticator UI?  

https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/blob/7ee9b1f9759bf12972e8838a36e3617a496fd9cb/toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/AuthenticatorState.kt#L224

- Screenshot using Vietnamese:
image

shubham7109:
@puneet-pdx Just read the original issue comments, you can ignore my comment on the invalid message.

puneet-pdx:
@hud10837 @shubham7109 thanks! merging.

PR #389, opened 2024-04-06
Fixes BackHandler Bug

eri9000:
Fix the following: 
Any other BackHandler defined in the app will not work when the Authenticator composable is placed last as the Authenticator will have the inner most BackHandler enabled.
```kotlin
@Composable
fun MyApp() {
    val authenticatorState: AuthenticatorState = remember { AuthenticatorState() }
    MyAppContent()
    Authenticator(authenticatorState)
}
```


hud10837 (approved):
This fix works for me 👍

PR #393, opened 2024-04-15
`Forms` : Add OAuth login

kaushikrw:
### Summary of changes

- The enterprise login will now use OAuth to sign in.
- Created an OAuth configuration on the developer dashboard that is linked to the micro app.
- This OAuth config `clientId` must be specified in the `local.properties` file that gets picked up by gradle secrets.

### DIFF
--- a/microapps/FeatureFormsApp/README.md
+++ b/microapps/FeatureFormsApp/README.md
@@ -17,4 +17,11 @@ webMapUser=XXX
 webMapPassword=YYY
 ```
 
+To authenticate using OAuth, provide the Client ID value in `local.properties`. If an OAuth client id
+is not provided, the app will prompt a username and password dialog.
+
+```
+clientId=YOUR_CLIENT_ID
+```
+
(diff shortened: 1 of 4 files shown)


sorenoid (changes requested):
This works great and is really helpful to see other maps! 

A default non-functionalvalue for `clientId` should be placed in `secrets.defaults.properties`. Other than that I have suggestions but would otherwise approve at this point!

For users we build this app for, i.e. internal testers, it would be good to indicate that oauth is an option on the sign in screen, rather than or in addition to `Sign in with ArcGIS Enterprise`. 

One idea is that if the clientId is present in the BuildConfig and not the default value, another button can be displayed that says `Sign in to AGOL with oauth` that leads to the same custom chrome tabs experience.

kaushikrw:
@sorenoid Updated! OAuth will now only work when a client id is provided, otherwise it will use the Username/Password Authentication.

sorenoid (approved):
very nice improvement!

kaushikrw:
Thanks @sorenoid, merging.

PR #394, opened 2024-04-16
Changes to abstract strings used for accessibility

puneet-pdx:
- I tested the Floorfilter app with the TalkBack feature and kept only the strings that were being used. 

- There are also some changes that update the formatting 
- Also have removed unicode characters from String.xml as I think they are causing confusion amongst translators

### DIFF
--- a/toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/FloorFilter.kt
+++ b/toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/FloorFilter.kt
@@ -47,6 +47,7 @@ import androidx.compose.ui.draw.clip
 import androidx.compose.ui.geometry.Size
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.contentDescription
 import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.text.font.FontWeight
@@ -115,7 +116,6 @@ public fun FloorFilter(
     val uiProperties = floorFilterState.uiProperties
 
     Surface(
-        modifier = modifier.semantics { contentDescription = "FloorFilterComponent" },
         shadowElevation = 10.dp,
         color = Color.Transparent
     ) {
@@ -279,7 +279,9 @@ internal fun FloorListColumn(
         if (measureHeight == null)
             modifier.fillMaxWidth()
         else
-            modifier.fillMaxWidth().height(measureHeight),
+            modifier
+                .fillMaxWidth()
+                .height(measureHeight),
         reverseLayout = true,
         userScrollEnabled = true
     ) {
@@ -312,7 +314,7 @@ internal fun SiteFacilityButton(
     onFacilitiesSelectorVisible: (Boolean) -> Unit
 ) {
     Box(
-        modifier = modifier.semantics { contentDescription = "SiteFacilityButton" }
+        modifier = modifier
             .height(uiProperties.buttonSize.height.dp)
             .clickable {
                 onSiteFacilitySelectorVisibilityChanged(true)
@@ -320,7 +322,7 @@ internal fun SiteFacilityButton(
         Icon(
             painter = painterResource(id = R.drawable.ic_site_facility_24),
             tint = uiProperties.selectedForegroundColor,
-            contentDescription = "Facilities icon",
+            contentDescription = stringResource(R.string.site_facility_button),
             modifier = modifier
                 .height(uiProperties.buttonSize.height.dp)
                 .width(uiProperties.buttonSize.width.dp)
@@ -352,15 +354,15 @@ internal fun FloorListCloseButton(
     buttonSize: Size,
     onClick: (Unit) -> Unit
 ) {
-    Box(modifier
-        .fillMaxWidth()
-        .height(buttonSize.height.dp)
-        .clickable { onClick(Unit) }
-        .semantics { contentDescription = "FloorListCloseButton" }) {
+    Box(
+        modifier
+            .fillMaxWidth()
+            .height(buttonSize.height.dp)
+            .clickable { onClick(Unit) }) {
         Icon(
             modifier = modifier.align(Center),
             painter = painterResource(id = R.drawable.ic_x_24),
-            contentDescription = "Close icon"
+            contentDescription = stringResource(R.string.close)
         )
     }
 }
@@ -379,6 +381,7 @@ internal fun FloorLevelSelectButton(
     uiProperties: UIProperties,
     onFloorLevelSelected: (Int) -> Unit,
 ) {
+    val floorLevelSelectButton = stringResource(R.string.floor_level_select_button)
     Text(
         text = floorText,
         textAlign = TextAlign.Center,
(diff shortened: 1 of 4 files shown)


toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/FloorFilter.kt
```diff
@@ -115,7 +116,6 @@ public fun FloorFilter(
     val uiProperties = floorFilterState.uiProperties
 
     Surface(
-        modifier = modifier.semantics { contentDescription = "FloorFilterComponent" },
```
shubham7109, line 118: The change will break the `FloorFilterTests` instrumented test. Should this also be abstracted to `strings.xml`?

puneet-pdx: This particular string is not needed. I have modified the test so now it will perform the same test without making use of this.

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/FloorFilter.kt
```diff
@@ -312,15 +314,15 @@ internal fun SiteFacilityButton(
     onFacilitiesSelectorVisible: (Boolean) -> Unit
 ) {
     Box(
-        modifier = modifier.semantics { contentDescription = "SiteFacilityButton" }
```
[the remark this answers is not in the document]
puneet-pdx: This is also not needed for the test.

puneet-pdx:
@shubham7109 I have only kept the strings that makes sense from accessibility/testing point of view. The ones I have weeded out did not make sense from accessibility point of view and also were not needed for the automated test. I have not modified the test steps but just the content strings for the nodes being tested.

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/SiteAndFacilitySelector.kt
```diff
@@ -466,19 +482,25 @@ internal fun ListOfSitesOrFacilities(
 @Composable
 internal fun SiteOrFacilityItem(
     name: String,
-    isSelected: Boolean,
-    onSelected: (Int) -> Unit,
     index: Int,
+    isSelected: Boolean,
```
puneet-pdx, line 486: I have shuffled this fun's parameters to match the order in which they are from the only invoking site in the code.

toolkit/indoors/src/main/res/values/strings.xml
```diff
@@ -1,11 +1,29 @@
...
+    <!-- This text is used as a hint on a text field to add text to filter a list-->
+    <string name="floor_filter_view_filter_hint_facilities">Filter facilities…</string>
+    <!-- This text is a label on a text field that can fit 24 chars -->
     <string name="floor_filter_select_site">Select a site</string>
-    <!-- This text is a label on a textfield that can fit 23 chars -->
+    <!-- This text is a label on a text field that can fit 23 chars -->
     <string name="floor_filter_select_facility">Select a facility</string>
     <string name="floor_filter_no_results">No results found.</string>
-    <string name="floor_filter_site_selector_top_bar">Site\u0020-\u0020</string>
+    <string name="not_available">not available</string>
```
shubham7109, line 12: Not sure, but I think this would need to be capitalized: 
```suggestion
    <string name="not_available">Not available</string>
```

shubham7109 (approved):
Looks good 👍, just the one comment.

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/SiteAndFacilitySelector.kt
```diff
@@ -466,19 +482,25 @@ internal fun ListOfSitesOrFacilities(
...
+    isSiteItem: Boolean,
     uiProperties: UIProperties,
-    isSiteItem: Boolean
+    onSelected: (Int) -> Unit
 ) {
+    val itemContentDescription =
+        if (isSiteItem) stringResource(id = R.string.site_item) else stringResource(id = R.string.facility_item)
```
ruiqima, line 492: I tried with TalkBack and this works well with "facility item Azimuth in list". Just curious how we manage to have it read the option's name (e.g. Azimuth), is there a specific rule for TalkBack to do this?

The reason I'm thinking this is, when tapping on a floor level, it only said "floor level select button", without the specific level number. It could be better if we can have it speak out the exact floor number as in the above example. But I'm not sure if we could achieve this or not.

ruiqima (approved):
LGTM 👍 Tried with TalkBack and it works well. Left one comment.

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/SiteAndFacilitySelector.kt
ruiqima: Another interesting behavior for TalkBack I found is, if you swipe up or down on the screen, it may gave you weird english letters. Don't understand what it is doing. But I think this is TalkBack thing, not from our FloorFilter. Just want to share this discovery.

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/SiteAndFacilitySelector.kt
```diff
@@ -253,20 +263,24 @@ internal fun FacilitySelectorTopBar(
 
             Text(
                 text = stringResource(R.string.floor_filter_site_selector_top_bar) +
-                        (floorFilterState.getSelectedSite()?.name ?: "not available"),
+                        (floorFilterState.getSelectedSite()?.name
+                            ?: stringResource(id = R.string.not_available)),
```
sorenoid, line 267: In a RTL language shouldn't this text be `<name> - Site`? If I understand correctly, then the current string will not be correct in a RTL language, regardless of translation. It seems like the name of the site should be passed into stringResource as a format parameter.
```suggestion
val name =  floorFilterState.getSelectedSite()?.name ?: stringResource(id = R.string.not_available)
Text(
    text = stringResource(R.string.floor_filter_site_selector_top_bar, name)
```

and the string resource should be

```xml
<!-- This text is a label on a text field that displays the name of the site after the hyphen -->
<string name="floor_filter_site_selector_top_bar">Site - %1$s</string>
```

so that it can be translated to a RTL language
```
<string name="floor_filter_site_selector_top_bar">%1$s - Site</string>
```

sorenoid (changes requested):
looks good. just one question.

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/SiteAndFacilitySelector.kt
puneet-pdx: @ruiqima that happens when it starts reading the text on an element. I starts to spell it.

puneet-pdx: also there we are not specifying how TalkBack is reading the list. Maybe we can look into it further for the floor level select button.

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/SiteAndFacilitySelector.kt
puneet-pdx: Good call. Updated

sorenoid (approved):

puneet-pdx:
Thanks @sorenoid @shubham7109 @ruiqima. Merging..

PR #395, opened 2024-04-16
Separate creation of component and microapps

hud10837:
Modifies the new component script to separate creating a microapp into its own script

- Removes the `-n` name flag and instead adds a prompt for the name after running the script
- Adds a prompt for optionally creating a microapp
- Fixes some issues where the plugins block which is added dynamically to new modules was not working correctly, now it finds the exact location of the existing plugins block and replaces it.
- Removes the dependency on `template` in `TemplateApp`. This means new microapps don't automatically have a dependency on a toolkit component, but I think it's easy to add manually and now these are separate scripts we don't always know what the author will want to add as. a dependency


shubham7109:
I tried using the script, and it works well. Some thoughts:
- The package name used and the package folders are inconsistent 
- The copyright year is `2023` would be nice if it was dynamically updated. 
- It looks like the script is using the `TemplateApp` to spin up the code, would you say the eventual goal is to optimize this module to use a standard theme, toolbar, etc. 
- Should the script support making a microapp without the need to create a toolkit component? For example, the GeoCompose module is used across multiple microapps, so I think would be nice to be able to create an isolated microapp.

shubham7109 (approved):
Nice work @hud10837 👍

hud10837:
Thanks @sorenoid @shubham7109

PR #396, opened 2024-04-16
Fix for not showing random floor data

puneet-pdx:
Earlier implementation was making all the floor levels visible when clearing map filter, which is not the right thing to do as it leads to unintended results.  
The fix here removes clearMapFilter logic to match javascript api implementation


shubham7109 (approved):
Tested with `Datum` and `Meridian` buildings and floors match with the online [webmap](https://www.arcgis.com/apps/mapviewer/index.html?webmap=b4b599a43a474d33946cf0df526426f5).

eri9000 (approved):
LGTM 👍🏼

puneet-pdx:
Thanks @eri9000 and @shubham7109. Merging..

PR #399, opened 2024-04-18
`Forms`: Check if map has a FeatureFormDefinition

kaushikrw:
### Summary of changes

- When a Map is opened, a check is performed to see if any of its `FeatureLayer`s contains a `FeatureFormDefinition`. If none exists, an alert dialog is shown informing the user.
- Bumps SDK version to pick up the latest changes


microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapViewModel.kt
```diff
@@ -115,10 +125,34 @@ class MapViewModel @Inject constructor(
...
+            // check if this map has a FeatureFormDefinition on any of its layers
+            checkFeatureFormDefinition()
+        }
+    }
+
+    private suspend fun checkFeatureFormDefinition() {
+        map.load()
+        val layer = map.operationalLayers.firstOrNull {
+            if (it is FeatureLayer) {
+                it.load()
```
sorenoid, line 143: I don't think it is necessary to load the layer here. The `featureFormDefinition` will be non-null before the layer is loaded if it exists. It could be loaded below in the else case so that only the layer that is needed is loaded explicitly here.

sorenoid (changes requested):
Looks good, I suggest one small change

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapViewModel.kt
kaushikrw: @sorenoid It seems like the layer needs to be loaded at this time. I can update this when that changes.

sorenoid (approved):
based on the featureFormDefinition's availability before layer loading, approved.

kaushikrw:
thanks @sorenoid

PR #405, opened 2024-04-25
Shubham/microapp resources

shubham7109:


microapps/AuthenticationApp/app/src/main/java/com/arcgismaps/toolkit/authenticationapp/MainActivity.kt
```diff
@@ -60,28 +63,33 @@ class MainActivity : ComponentActivity() {
...
-        PortalDetails(
-            url = authenticationAppViewModel.url.collectAsState().value,
-            onSetUrl = authenticationAppViewModel::setUrl,
-            useOAuth = authenticationAppViewModel.useOAuth.collectAsState().value,
-            onSetUseOAuth = authenticationAppViewModel::setUseOAuth,
-            onSignOut = authenticationAppViewModel::signOut,
-            onLoadPortal = authenticationAppViewModel::loadPortal
-        )
-        InfoScreen(text = infoText, isLoading = isLoading)
+    Scaffold(
```
sorenoid, line 77: I am wondering if this is the right thing to do. Wouldn't this change mask any themimg lapses within the component? I am wondering if we want to deliberately expose such lapses so we catch them?

microapps/FloorFilterApp/app/src/main/java/com/arcgismaps/toolkit/floorfilterapp/MainActivity.kt
```diff
@@ -20,35 +20,44 @@ package com.arcgismaps.toolkit.floorfilterapp
...
     }
 }
 
+@OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun FloorFilterApp() {
-    MainScreen()
+    Scaffold(topBar = { TopAppBar(title = { Text("FloorFilter App") }) }) {
+        MainScreen(Modifier.padding(it))
+    }
```
sorenoid, line 54: I have the same question here: don't we want to know if the theme is not being respected by the component? My understanding is that placing it in a Scaffold will mask any shortcoming in the component of applying the app theme.

sorenoid (commented):
I have a question about microapps that do not showcase GeoView functionality. My understanding of the Authenticator was that placing it in a Scaffold masked the shortcomings in the Authenticator of applying the app's theme. If this is true, don't we want to avoid placing components like Authenticator and FloorFilter in Scaffolds in our microapps, so we will know if there are regressions?

microapps/AuthenticationApp/app/src/main/java/com/arcgismaps/toolkit/authenticationapp/MainActivity.kt
shubham7109: Removed the use of Scaffold from the Authenticator App

microapps/FloorFilterApp/app/src/main/java/com/arcgismaps/toolkit/floorfilterapp/MainActivity.kt
shubham7109: Removed the use of Scaffold from the FoorFilter App

sorenoid (approved):

shubham7109:
Thanks for the review @sorenoid, After our discussion, I have updated to keep the scaffold as part of these micro apps.

hud10837 (commented):
@shubham7109 can you comment why you reverted removing the Scaffolds? I don't disagree but am curious about the explanation

hud10837 (changes requested):
Could you also please make this change to the `TemplateApp`, so that any new microapps will automatically use the new theme.

shubham7109:
@hud10837 Sure thing, the concern was, would we have known about the Authenticator theme issue (#415) if the component were contained in a Scaffold in its microapp? > Yes, it was only by using a Scaffold on the MicroApp highlighted the drawback of the Authenticator component not applying the correct `contentColorFor`. This ensures that the colorScheme is respected across light/dark modes for all material components. 

The Scaffold does help update the `contentColorFor` through the default [parameter](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/Scaffold.kt;l=69?q=file:androidx%2Fcompose%2Fmaterial3%2FScaffold.kt%20function:Scaffold), and so does [Surface](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/Surface.kt;l=87). I felt it would be best to use `Scaffold` for microapps, and `Surface` in the case of the Auth Toolkit.

Now with the changes in (#415) with or without using a scaffold on the MicroApp, we can guarantee that the Authenticator component will display the right expected color scheme as the `Surface` will set the `contentColorFor`. Making it independent of the app level `Scaffold`: 

- DarkMode using `MicroAppTheme` without `Scaffold`

<image src="https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/assets/14892474/0f9a0fae-dd7b-40e2-8762-bddca9be640a" height=500>

- DarkMode using `MicroAppTheme` with `Scaffold`

<image src="https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/assets/14892474/01867064-3144-423c-9dad-5c1242e39d0b" height=500>

We decided to keep the Scaffold as it only provides cosmetic change for the MicroApp without affecting the Toolkit. @sorenoid Hope this summarizes our conversation accurately.

shubham7109:
Nice catch @hud10837. Updated the TemplateApp to use the common theme. Ready for your review.

hud10837 (approved):
Thanks for the explanation. LGTM

PR #409, opened 2024-04-29
`Forms` : Attachment Form Element

kaushikrw:
Summary of changes

Issue - https://devtopia.esri.com/runtime/apollo/issues/409

- Adds `AttachmentFormElement` composable with the api object and the state objects.
- Supports only display of attachments.
- This also includes loading the attachments and transition between different load states.
- Moved the attachment related files into their own sub package.


sorenoid (approved):
👍

PR #410, opened 2024-04-30
Automate authenticator test case 4

hud10837:
Automates the manual test steps for server trust prompts in the Authenticator.

Note these tests differ in a couple of ways to the manual tests:
- The manual tests verify behavior when putting the app in the background and returning to it, and also when rotating the device. In these tests we instead use a `StateRestorationTester`, which works slightly different (it only simulates disposing and recreating the composition) but vastly simplifies the tests. If we decide in the future that we need to do the bigger tests, we can add them then.
- The manual tests have a step for tapping outside of the alert dialog to dismiss. This behavior works with the authenticator but it is very difficult to test and is functionally the same as pressing the back button, so I've excluded that test.


toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/ServerTrustTests.kt
```diff
@@ -0,0 +1,169 @@
...
+ * Tests for server trust challenges.
+ *
+ * @since 200.4.0
+ */
+class ServerTrustTests {
+
+    @get:Rule
+    val composeTestRule = createAndroidComposeRule<ComponentActivity>()
+
+    private val authenticatorState = AuthenticatorState()
```
gunt0001, line 53: I think this can be declared locally in `testServerTrustChallengeWithStateRestoration()`

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/ServerTrustTests.kt
```diff
@@ -0,0 +1,169 @@
...
+        }.await()
+        assert(response is NetworkAuthenticationChallengeResponse.Cancel)
+    }
+
+    /**
+     * Places a DialogAuthenticator in the composition and issues a server trust challenge.
+     * Once the dialog is displayed, [userInputOnDialog] will be called to simulate user input.
+     * The function will return the response to the server trust challenge from the authenticator in a deferred.
+     *
+     * Note that this function will also simulate disposing and restoring the state of the DialogAuthenticator.
```
gunt0001, line 125: I would add here:
```suggestion
     * Note that this function will also simulate disposing and restoring the state of the DialogAuthenticator
     * before interaction with the server trust dialog.
```

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/ServerTrustTests.kt
```diff
@@ -0,0 +1,169 @@
...
+        userInputOnDialog()
+
+        advanceUntilIdle()
+        // ensure the dialog has disappeared
+        assert(authenticatorState.pendingServerTrustChallenge.value == null)
+        composeTestRule.onNodeWithText(serverTrustMessage).assertDoesNotExist()
+
+        // return the response deferred
+        return challengeResponse
+    }
```
gunt0001, line 168: You could wrap the code into a `with()` scope function:
```suggestion
       // StateRestorationTester simulates the state restoration process
        with(StateRestorationTester(composeTestRule)) {
            setContent {
                DialogAuthenticator(authenticatorState = authenticatorState)
            }
            // issue the server trust challenge
            val certificateHostName = "https://server-trust-tests.com/"
            val challengeResponse = async {
                authenticatorState.handleNetworkAuthenticationChallenge(
                    NetworkAuthenticationChallenge(
                        certificateHostName,
                        NetworkAuthenticationType.ServerTrust,
                        CertificateException("Test exception")
                    )
                )
            }

            val serverTrustMessage =
                composeTestRule.activity.getString(R.string.server_trust_message, certificateHostName)
            // ensure the dialog prompt is displayed as expected
            advanceUntilIdle()
            assert(authenticatorState.pendingServerTrustChallenge.value != null)
            composeTestRule.onNodeWithText(serverTrustMessage).assertIsDisplayed()

            // simulate a configuration change
            emulateSavedInstanceStateRestore()

            // perform the test action
            userInputOnDialog()

            advanceUntilIdle()
            // ensure the dialog has disappeared
            assert(authenticatorState.pendingServerTrustChallenge.value == null)
            composeTestRule.onNodeWithText(serverTrustMessage).assertDoesNotExist()

            // return the response deferred
            return challengeResponse
        }
    }
```

gunt0001 (approved):

shubham7109 (approved):
Looks good to me 👍 

One comment, would it be helpful to delete the `ExampleInstrumentedTest` and `ExampleUnitTest` for this module? I believe we made a similar change with FloorFilterTests.

eri9000 (approved):
LGTM 👍🏼

hud10837:
Thanks @eri9000 @shubham7109 @gunt0001 for the reviews

PR #412, opened 2024-04-30
`sprint x`: fix warnings

sorenoid:
address all warnings in the toolkit.

fix usage of bundles in test implementations that also require the compose bom dependency.

fix usage of deprecated API introduced by version upgrades.

feature forms: fix content descriptions warnings and other various warnings due to version upgrades, add compose bom dependency.

address a build failure in the kdoc project when building after clean: add source roots in the dokka configuration, rather than adding regular source sets. Only add dependencies needed for kdoc generation (exposed API).


eri9000 (approved):
LGTM

kaushikrw (approved):
LGTM

PR #413, opened 2024-05-01
`Popup`: Adds TextPopupElement

eri9000:
Adds TextPopupElement composable.
This composable can render HTML text.


toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/ExpandableCard.kt
```diff
@@ -0,0 +1,79 @@
...
+                    )
+                    IconButton(onClick = { expanded = !expanded }) {
+                        Icon(
+                            imageVector = if (expanded) Icons.Default.KeyboardArrowUp else Icons.Default.KeyboardArrowDown,
+                            contentDescription = "More"
+                        )
+                    }
+                }
+            }
+
```
sorenoid, line 60: There should probably be a Spacer here.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/ExpandableCard.kt
```diff
@@ -0,0 +1,79 @@
...
+    ) {
+        Column {
+            if (title.isNotEmpty()) {
+                Row(modifier = Modifier.fillMaxWidth(),horizontalArrangement = Arrangement.SpaceBetween) {
+                    Text(
+                        title,
+                        modifier = Modifier.padding(8.dp),
+                        style = MaterialTheme.typography.titleMedium,
+                    )
+                    IconButton(onClick = { expanded = !expanded }) {
```
sorenoid, line 52: Is IconButton necessary? It seems like an Icon wrapped in a CrossFade animation will work. I think at any rate the animation is necessary.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/ExpandableCard.kt
```diff
@@ -0,0 +1,79 @@
...
+                        Icon(
+                            imageVector = if (expanded) Icons.Default.KeyboardArrowUp else Icons.Default.KeyboardArrowDown,
```
sorenoid, line 54: I think the carets are not centered in their bitmaps. Better to use `Icons.Rounded.ExpandLess` and `Icons.Rounded.ExpandMore`

sorenoid (changes requested):
Looks nice. Some small suggestions.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/ExpandableCard.kt
eri9000: I've added a spaced only when there's a title.

sorenoid (approved):

PR #416, opened 2024-05-01
`Forms` : Add attachments

kaushikrw:
### Summary of changes

- Added "Add attachment" feature which lets a user add an Image attachment via Camera capture or Gallery picker.
- This requires `CAMERA` permissions and the `FeatureFormsApp` requests the permissions and shows a dialog if necessary permissions are not granted.
- The `FeatureForm` composable will not display `Take Photo` option for attachments if the `Camera` permission is not granted.
- "Add attachment" is only available if the form element is editable.
- Committing of edits with Attachments need to happen on the `ServiceGeoDatabase` if it supports edits.
- Fixed an issue with finding a `FeatureFormDefinition` if the layer is a `GroupLayer`.


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/utils/AttachmentCaptureFileProvider.kt
```diff
@@ -0,0 +1,43 @@
...
+
+internal class AttachmentCaptureFileProvider :
+    FileProvider(R.xml.feature_forms_captured_attachments) {
+    companion object {
+        private const val AUTHORITY = "com.arcgismaps.toolkit.featureforms.capturefileprovider"
+
+        fun getImageUri(file: File, context: Context): Uri {
+            val directory = File(context.cacheDir, "feature_forms_attachments")
+            directory.mkdirs()
+            // The authority string must be unique per device. Therefore use of this provider with two
```
sorenoid, line 33: I've noticed this will cause previews to fail to install.

kaushikrw: Perhaps this is okay for now. And we can move to use the camera2 or cameraX api to bring the capture into the component itself.

sorenoid (approved):

kaushikrw:
thanks @sorenoid!

PR #417, opened 2024-05-06
`Forms`: Delete & rename attachments

kaushikrw:
### Summary of changes

- Introduced renaming and deleting of attachments.
- Renaming an attachment happens via a `RenameAttachmentDialog` as part of the `DialogRequester`.

Note - delete attachments is not working as expected for existing attachments. The deleted attachment does not get removed from the `AttachmentFormElement.attachments` list until after the feature is saved and the map is re-opened.


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
```diff
@@ -205,6 +234,23 @@ internal class FormAttachmentState(
...
+
+    override fun hashCode(): Int {
+        return Objects.hash(name, size, type)
+    }
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as FormAttachmentState
```
sorenoid, line 246: I didn't realize this was possible but it makes sense 👍

sorenoid (approved):

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
```diff
@@ -182,19 +203,27 @@ internal class FormAttachmentState(
...
+    constructor(
+        element: AttachmentElementState,
+        attachment: FormAttachment,
+        scope: CoroutineScope
+    ) : this(
         name = attachment.name,
         size = attachment.size,
+        elementStateId = element.id,
         loadStatus = attachment.loadStatus,
         onLoadAttachment = attachment::load,
```
sorenoid, line 215: If a Loadable is FailedToLoad state, calling `load()` on it will immediately fail. You must call retryLoad() to load a previously failed Loadable.

```suggestion
        onLoadAttachment = attachment::retryLoad,
```

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentTile.kt
```diff
@@ -21,113 +21,202 @@ import androidx.compose.foundation.BorderStroke
...
+                    }
+                }
+
+                is PressInteraction.Release -> {
+                    if (!wasALongPress) {
+                        haptic.performHapticFeedback(HapticFeedbackType.TextHandleMove)
+                        // handle single tap
+                        if (loadStatus is LoadStatus.NotLoaded || loadStatus is LoadStatus.FailedToLoad) {
+                            // load attachment
+                            state.load()
```
sorenoid, line 213: this needs to invoke `retryLoad()`

sorenoid (changes requested):

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
kaushikrw: thanks, updated!

sorenoid (approved):

PR #418, opened 2024-05-07
`Forms` : Fix state references

kaushikrw:
### Summary of changes

- `DialogRequester` saves the references to the states that it is passed when a field requests a dialog. In case of a configuration change, new states are created but the `DialogRequester` is still referencing the old state.
- This PR fixes this issue by making the `DialogRequester` accept a `stateId` instead of the state object. This `stateId` can then be used to fetch the actual state from the `FormStateCollection`.
- The `stateId` mechanism is relying in the `FormElement.hashCode()` implementation which ensures that even after an orientation change the `hashCode()` will not change because the underlying core object is the same.


sorenoid (approved):
exactly as discussed. very nice.

kaushikrw:
thanks @sorenoid

PR #420, opened 2024-05-08
`Popup`: state collection

sorenoid:
* Provides a collection of Immutable, Parcelable state objects for each PopupElement (so far).

The benefit of passing state objects to composables are that they are immutable and enable smart recomoposition of the Popup within the compose runtime.

There is a bit of complexity with the creation of state objects:
they must be maintained at the top level of the composition so that they don't get recreated when the lazy list of elements scrolls an element out and into view.
they need to be saveable to survive config changes. Luckily we can just make them Parcelable for this purpose.

* Reorganizes file structure and packages a bit.
* introduce a ViewModel for the microapp to ensure app state survives config changes (so we can test support for config changes within the Popup composable).


microapps/PopupApp/app/src/main/java/com/arcgismaps/toolkit/popupapp/screens/mapscreen/MapViewModel.kt
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Simple android view model for the Popup app map screen.
+ * Provides persistence of state across orientation changes etc. This allows us to test
+ * persistence of internal composable Popup() state across orientation changes etc.
+ */
+@Suppress("unused_parameter")
+class MapViewModel(
+    savedStateHandle: SavedStateHandle,
+    application: Application,
+    private val coroutineScope: CoroutineScope = CloseableCoroutineScope()
```
eri9000, line 60: dose this need to marked as `private val`?

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/Popup.kt
```diff
@@ -163,3 +171,34 @@ internal fun InitializingExpressions(
         LinearProgressIndicator(modifier)
     }
 }
+
+/**
+ * Creates and remembers state objects for all the supported element types that are part of the
+ * provided FeatureForm. These state objects are returned as part of a [PopupElementStateCollection].
```
eri9000, line 177:
```suggestion
 * provided Popup. These state objects are returned as part of a [PopupElementStateCollection].
```

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/textelement/TextElementState.kt
```diff
@@ -0,0 +1,51 @@
...
+import kotlinx.parcelize.Parcelize
+
+/**
+ * A class to handle the state of a [TextPopupElement].
+ *
+ * @param value the text of the element.
+ */
+@Immutable
+@Parcelize
+internal class TextElementState(
```
eri9000, line 35: I am getting a build error here: 
```
Class 'TextElementState' is not abstract and does not implement abstract base class member public abstract fun describeContents(): Int defined in com.arcgismaps.toolkit.popup.internal.element.state.PopupElementState
```

eri9000 (commented):
Good work, I have a few suggestions

eri9000 (approved):

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/textelement/TextElementState.kt
sorenoid: the @Parcelize annotation should provide this. I am able to build this, but I think to move forward I can just remove the `Parcelable` super class from the base class, which I need to do anyway for attachment states which cannot be Parcelable.

microapps/PopupApp/app/src/main/java/com/arcgismaps/toolkit/popupapp/screens/mapscreen/MapViewModel.kt
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Base class for context aware AndroidViewModel. This class must have only a single application
+ * parameter.
+ */
+open class BaseMapViewModel(application: Application) : AndroidViewModel(application)
+
+/**
+ * Simple android view model for the Popup app map screen.
+ * Provides persistence of state across orientation changes etc. This allows us to test
```
puneet-pdx, line 53: Do we need to mention this / can this be removed ?
"This allows us to test persistence of internal composable Popup() state across orientation changes etc."

puneet-pdx (approved):
@sorenoid just a question about a comment. Looks good otherwise

PR #422, opened 2024-05-10
`Forms` : View Attachments

kaushikrw:
### Summary of changes

- Added attachment viewer by invoking an intent to display the downloaded attachment via its appropriate file type.
- Changed the `AttachmentFileProvider` authority to be unique by prefixing it with the `applicationId`. With this multiple apps can exist on the same device that use the library.
- `FormAttachmentState` is now loadable to include the time taken to save the loaded attachment data to disk.


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
```diff
@@ -176,21 +204,23 @@ internal class AttachmentElementState(
...
+    val contentType: String,
     val elementStateId: Int,
-    val loadStatus: StateFlow<LoadStatus>,
-    private val onLoadAttachment: suspend () -> Result<Unit>,
-    private val onLoadThumbnail: suspend () -> Result<BitmapDrawable?>,
     val deleteAttachment: suspend () -> Unit,
+    private val filesDir: String,
     private val scope: CoroutineScope,
-) {
+    private val formAttachment: FormAttachment? = null
```
sorenoid, line 222: Why is this nullable?

sorenoid (changes requested):
Looks good. I have a couple questions

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
kaushikrw: This is to enable compose previews for `AttachmentTile`. For the preview null can be passed while at runtime during state creation, this will be populated.

sorenoid (approved):

PR #425, opened 2024-05-13
`Forms` : Add attachment fixes

kaushikrw:
### Summary of changes

[Issue](https://devtopia.esri.com/runtime/apollo/issues/678)

- Fixes an issue where loaded attachments become _not-loaded_ after a new attachment is added.
- This was due to the fact that the `FormAttachmentState` list is rebuilt after calling `fetchAttachments` and hence any previous UI state is lost.
- The fix is to re-use the `FormAttachmentState` objects after a call to `fetchAttachments`.
- Also fixes a possible bug in `Dialog` where the state object can be empty.


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
```diff
@@ -80,7 +82,7 @@ internal class AttachmentElementState(
     /**
      * The attachments associated with the form element.
      */
-    val attachments = SnapshotStateList<FormAttachmentState>()
+    var attachments by mutableStateOf(emptyMap<Int, FormAttachmentState>())
```
sorenoid, line 85: FormAttachments are unique, persistent, and implement equals and hashCode. You can key on them, and if that doesn't work there is likely a bug somewhere.
```suggestion
    var attachments by mutableStateOf(emptyMap<FormAttachment, FormAttachmentState>())
```

sorenoid (changes requested):
looks good, just one suggestion to simplify the mapping of FormAttachment to FormAttachmentState.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
kaushikrw: I believe this should work too, but it the adds a `FormAttachment` dependency to the `AttachmentFormElement`. This would mean previews cannot be generated with a list of attachments.

```
@Composable
AttachmentFormElement(
    ..
    attachments: Map<FormAttachment, FormAttachmentState>,
    ..
)

sorenoid (approved):

PR #427, opened 2024-05-14
Configure aggregated local tests

gunt0001:
issue: https://devtopia.esri.com/runtime/kotlin/issues/3991

### Description

Applies and configures the [gmazzo test aggregation plugin](https://github.com/gmazzo/gradle-android-test-aggregation-plugin) to get an aggregated test report of all the local tests in all the modules of this project.

Note:
- This plugin does not work for Android connected tests. For connected tests there will be a follow up PR that handles test reports of connected tests differently.
- I tried to use Gradle's [Test Report Aggregation Plugin](https://docs.gradle.org/current/userguide/test_report_aggregation_plugin.html), but it doesn't work in combination with the Android Gradle Plugin :(
- Some modules such as `bom`, `kdoc` etc have been excluded from test aggregation.
- To run aggregated local tests:
```
./gradlew testAggregatedReport
```
Test report will be in `arcgis-maps-sdk-kotlin-toolkit/build/reports`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Please see the [package structure](doc/general/developer_setup.md#package-struct
 
 ### Testing
 
-In order to run *local* tests of all modules and get an aggregated test report, run the following at the root folder of the project:
+In order to run *local* (non-instrumented) tests of all modules and get an aggregated test report, run the following at the root folder of the project:
 ```
 ./gradlew testAggregatedReport
 ```
--- a/README.md
+++ b/README.md
@@ -97,6 +97,14 @@ The template and TemplateApp modules are for bootstrapping new modules.
 
 Please see the [package structure](doc/general/developer_setup.md#package-structure) documentation for more details.
 
+### Testing
+
+In order to run *local* tests of all modules and get an aggregated test report, run the following at the root folder of the project:
+```
+./gradlew testAggregatedReport
+```
+The test report for local tests can be located under `arcgis-maps-sdk-kotlin-toolkit/build/reports`.
+
 ## Licensing
 
 Copyright 2019-2022 Esri
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,3 +1,4 @@
+
 /*
  *
  *  Copyright 2023 Esri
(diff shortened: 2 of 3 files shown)


hud10837 (approved):
👍 I was able to run the script and get the test report for all modules

build.gradle.kts
```diff
@@ -38,3 +39,38 @@ buildscript {
         classpath(libs.dokka.versioning)
     }
 }
+
+/**
+ * Configures the [gmazzo test aggregation plugin](https://github.com/gmazzo/gradle-android-test-aggregation-plugin)
+ * with all local tests to be aggregated into a single test report.
+ * Note: This works only for local tests, not for connected tests.
+ * To run aggregated local tests:
```
TADraeseke, line 47:
```suggestion
 * To run aggregated local tests, run the following at the root folder of the project:
```

README.md
```diff
@@ -97,6 +97,14 @@ The template and TemplateApp modules are for bootstrapping new modules.
 
 Please see the [package structure](doc/general/developer_setup.md#package-structure) documentation for more details.
 
+### Testing
+
+In order to run *local* (non-instrumented) tests of all modules and get an aggregated test report, run the following at the root folder of the project:
+```
+./gradlew testAggregatedReport
+```
+The test report for local tests can be located under `arcgis-maps-sdk-kotlin-toolkit/build/reports`.
```
TADraeseke, line 106:
```suggestion
The tests take about 5-10 minutes to run and the final report can be found in `arcgis-maps-sdk-kotlin-toolkit/build/reports`.
```

I find it helpful to know how long scripts take to run (do I have time and when do I start worrying it's stuck)

TADraeseke (approved):
LGTM! Ran it myself and got the report. Just a couple small optional comments.

README.md
gunt0001: I don't think we should add this type of info here. As we add more tests, this time will likely increase, so the numbers wouldn't reflect reality anyway, unless we keep this info updated, which is not the point of this doc from my POV.

TADraeseke (approved):
LGTM!

PR #429, opened 2024-05-15
Fix for the rotation bug

puneet-pdx:
- Add logic to only initialize the Callout when the MapView is done with initial draw
- Move the onViewPointChanged listener from MapView to Callout to prevent unnecessary MapView recompositions.
- Move invocation of the Content lambda after all the MapViewEventHandlers and ViewPointHandlers are set

### DIFF
--- a/toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/Callout.kt
+++ b/toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/Callout.kt
@@ -26,7 +26,6 @@ import androidx.compose.foundation.layout.wrapContentSize
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
-import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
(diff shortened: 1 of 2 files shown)


toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/Callout.kt
```diff
@@ -55,27 +60,58 @@ public class MapViewScope(internal val mapView: MapView)
...
-                .background(Color.White)
-                .border(
-                    border = BorderStroke(2.dp, Color.LightGray),
-                    shape = MaterialTheme.shapes.medium
-                )
-        )
-        {
-            this.content()
+    // Convert the given location to a screen coordinate
+    var calloutScreenCoordinate: ScreenCoordinate by remember {
```
sorenoid, line 88: possibly this could key on `location` then it doesn't need to be assigned in the LaunchedEffect.
```suggestion
    var calloutScreenCoordinate: ScreenCoordinate by remember(location) {
```

sorenoid (approved):
looks good. Good to get the supporting logic out of the MapView body. One suggestion to consider.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/Callout.kt
puneet-pdx: I think this is not the key but the initial value for the calloutScreenCoordinate. we are using the changes to the state of calloutScreenCoordinate to trigger recomposition.

shubham7109 (approved):

PR #432, opened 2024-05-16
Automate authenticator oauth tests

hud10837:
Adds tests for OAuth sign in. This covers the manual test cases 1 and 2. Like the `ServerTrustTests`, the test logic is generally implemented in a single util function that is called by each test along with the corresponding test action (eg. enter credentials and sign in)

I found that because we are calling `handleArcGISAuthenticationChallenge` directly, the tests were catching any `OperationCancelledException`s that were thrown by that method. It seems like this was an oversight so I have changed the implementation and it seems to be working (at least, the tests are passing)

Currently the `signIn` tests will fail because they don't have working credentials until we come up with a solution for test credentials in the toolkit.

~~This also uses the microapp's oauth configuration as well as the sample repo's oauth configuration in order to avoid a launcher disambiguation dialog in the tests. This is something we should confirm with @shubham7109 is okay or we should create new redirect urls and clientIds.~~

The tests use two new OAuth clientId/redirectUrls.


toolkit/authentication/build.gradle.kts
```diff
@@ -49,6 +49,10 @@ android {
     buildFeatures {
         compose = true
     }
+    packagingOptions {
+        exclude("META-INF/LICENSE-notice.md")
+        exclude("META-INF/LICENSE.md")
+    }
```
hud10837, line 55: I couldn't compile after adding uiautomator and mockk without this, but I'm not sure if that's just a local issue

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/OAuthDefaultConfigurationTests.kt
hud10837, line 1: Test Case 1

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/OAuthUserLauncherTests.kt
hud10837, line 1: Test Case 2

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/AuthenticatorState.kt
```diff
@@ -179,7 +180,13 @@ private class AuthenticatorStateImpl(
...
                         // to remove the OAuth prompt when this value changes.
                         _pendingOAuthUserSignIn.value = null
-                    }.getOrThrow()
+                    }.getOrElse {
+                        if (it is OperationCancelledException) {
+                            return ArcGISAuthenticationChallengeResponse.Cancel
+                        } else {
+                            return ArcGISAuthenticationChallengeResponse.ContinueAndFailWithError(it)
+                        }
+                    }
```
hud10837, line 189: We should discuss this. Because the tests use `handleArcGISAuthenticationChallenge()` directly, any exceptions that get thrown are thrown into the test. Normally the caller of `handleArcGISAuthenticationChallenge` will handle the exceptions but it seems to me that we should be handling it here anyway.

gunt0001: I don't think this change is needed. Any exception thrown, would be handled in the SDK's `AuthenticationChallengeQueue`. In the test code, since we bypass that code, we should just catch the exception there.

gunt0001 (commented):
@hud10837 looks good, a few comments.

hud10837:
Thanks @gunt0001 I believe I have addressed all of your comments

toolkit/authentication/build.gradle.kts
gunt0001: I think that change is correct.

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/OAuthDefaultConfigurationTests.kt
```diff
@@ -0,0 +1,162 @@
...
+     *
+     * Note: This function automatically redirects to the app with a successful response without user input,
+     * and fakes the token validation response.
+     *
+     * @since 200.5.0
+     */
+    @Test
+    fun successfulSignIn() = runTest {
+        val response = testOAuthChallengeWithStateRestoration {
+            InstrumentationRegistry.getInstrumentation().context.startActivity(getSuccessfulRedirectIntent("kotlin-authentication-test-1://auth"))
```
gunt0001, line 65: Could you add a comment here, something along the lines of...
```suggestion
            // When the OAuth sign in dialog shows up, we simulate successful sign in by launching an intent with the redirect URl, which otherwise would be sent from the Portal server, but would require valid credentials
            InstrumentationRegistry.getInstrumentation().context.startActivity(getSuccessfulRedirectIntent("kotlin-authentication-test-1://auth"))
```

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/OAuthDefaultConfigurationTests.kt
```diff
@@ -0,0 +1,162 @@
...
+     *
+     * @since 200.5.0
+     */
+    @OptIn(ExperimentalCoroutinesApi::class)
+    fun TestScope.testOAuthChallengeWithStateRestoration(
+        userInputOnDialog: UiDevice.() -> Unit,
+    ): Deferred<Result<ArcGISAuthenticationChallengeResponse>> {
+        ArcGISEnvironment.configureArcGISHttpClient {
+            interceptTokenRequests()
+        }
```
gunt0001, line 115: I suggest to move this part of setting up an interceptor out of this function and call this individually as part of each test function. I don't think it belongs into this function here.

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/OAuthDefaultConfigurationTests.kt
```diff
@@ -0,0 +1,162 @@
...
+        val authenticatorState = AuthenticatorState().apply {
+            oAuthUserConfiguration = OAuthUserConfiguration(
+                "https://arcgis.com",
+                // This client ID is for demo purposes only. For use of the Authenticator in your own app,
+                //            // create your own client ID. For more info see: https://developers.arcgis.com/documentation/mapping-apis-and-services/security/tutorials/register-your-application/
```
gunt0001, line 120: Comment needs formating
```suggestion
                // create your own client ID. For more info see: https://developers.arcgis.com/documentation/mapping-apis-and-services/security/tutorials/register-your-application/
```

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/OAuthDefaultConfigurationTests.kt
```diff
@@ -0,0 +1,162 @@
...
+                "kotlin-authentication-test-1://auth"
+            )
+        }
+        composeTestRule.setContent {
+            Authenticator(
+                authenticatorState = authenticatorState,
+                modifier = Modifier.testTag("Authenticator")
+            )
+        }
+        // isse the OAuth challenge
```
gunt0001, line 131:
```suggestion
        // issue the OAuth challenge
```

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/OAuthDefaultConfigurationTests.kt
```diff
@@ -0,0 +1,162 @@
+package com.arcgismaps.toolkit.authentication
```
gunt0001, line 1: missing copyright

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/OAuthUserLauncherTests.kt
```diff
@@ -0,0 +1,202 @@
+package com.arcgismaps.toolkit.authentication
```
gunt0001, line 1: missing copyright

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/OAuthUserLauncherTests.kt
```diff
@@ -0,0 +1,202 @@
...
+     *
+     * Note: This function automatically redirects to the app with a successful response without user input,
+     * and fakes the token validation response.
+     *
+     * @since 200.5.0
+     */
+    @Test
+    fun successfulSignIn() = runTest {
+        val response = testOAuthChallengeWithStateRestoration {
+            InstrumentationRegistry.getInstrumentation().context.startActivity(getSuccessfulRedirectIntent("kotlin-authentication-test-2://auth"))
```
gunt0001, line 70: add comment here similar to above

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/OAuthUserLauncherTests.kt
```diff
@@ -0,0 +1,202 @@
...
+     *
+     * @since 200.5.0
+     */
+    @OptIn(ExperimentalCoroutinesApi::class)
+    fun TestScope.testOAuthChallengeWithStateRestoration(
+        userInputOnDialog: UiDevice.() -> Unit,
+    ): Deferred<Result<ArcGISAuthenticationChallengeResponse>> {
+        ArcGISEnvironment.configureArcGISHttpClient {
+            interceptTokenRequests()
+        }
```
gunt0001, line 120: As above, I suggest to move the logic to set up the interceptor out of this function and call in each individual test that needs it.

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/OAuthUserLauncherTests.kt
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * Places an [Authenticator] in the composition and issues an OAuth challenge.
+     * Once the browser is launched, [userInputOnDialog] will be called to simulate user input.
+     * Also, the device will be rotated to ensure that the Authenticator can handle configuration changes
+     * before calling [userInputOnDialog].
+     *
+     * @since 200.5.0
+     */
+    @OptIn(ExperimentalCoroutinesApi::class)
+    fun TestScope.testOAuthChallengeWithStateRestoration(
```
gunt0001, line 115: Is this a duplicate of the function in `OAuthDefaultConfigurationTests`?

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/OAuthUserLauncherTests.kt
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * @since 200.5.0
+ *
+ */
+class OAuthUserLauncherTestViewModel : ViewModel() {
+    val authenticatorState = AuthenticatorState().apply {
+        oAuthUserConfiguration = OAuthUserConfiguration(
+            "https://arcgis.com/",
+            // This client ID is for demo purposes only. For use of the Authenticator in your own app,
+            //            // create your own client ID. For more info see: https://developers.arcgis.com/documentation/mapping-apis-and-services/security/tutorials/register-your-application/
```
gunt0001, line 197:
```suggestion
            // create your own client ID. For more info see: https://developers.arcgis.com/documentation/mapping-apis-and-services/security/tutorials/register-your-application/
```

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/TestAuthenticationUtils.kt
```diff
@@ -0,0 +1,77 @@
+package com.arcgismaps.toolkit.authentication
```
gunt0001, line 1: missing copyright

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/TestAuthenticationUtils.kt
```diff
@@ -0,0 +1,77 @@
...
+    every { requestUrl } returns "https://arcgis.com"
+    every { cause } returns Throwable()
+}
+
+/**
+ * Returns an intent that simulates a successful redirect from the OAuth server.
+ *
+ * @since 200.5.0
+ */
+fun getSuccessfulRedirectIntent(redirectUrl: String): Intent {
```
gunt0001, line 27: nitpick naming suggestion:
```suggestion
fun createSuccessfulRedirectIntent(redirectUrl: String): Intent {
```

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/TestAuthenticationUtils.kt
```diff
@@ -0,0 +1,77 @@
...
+        flags = Intent.FLAG_ACTIVITY_CLEAR_TOP
+    }
+}
+
+/**
+ * Intercepts requests to validate a token and returns a fake response that validates the token.
+ *
+ * @since 200.5.0
+ */
+fun ArcGISHttpClient.Builder.interceptTokenRequests() {
```
gunt0001, line 39: nitpick naming suggestion:
```suggestion
fun ArcGISHttpClient.Builder.setupTokenRequestInterceptor() {
```

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/TestAuthenticationUtils.kt
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+}
+
+/**
+ * Returns a fake response that validates the token.
+ *
+ * @since 200.5.0
+ */
+fun Request.getFakeTokenResponse(): Response =
```
gunt0001, line 54: nitpick naming suggestion:
```suggestion
fun Request.createFakeTokenResponse(): Response =
```

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/TestAuthenticationUtils.kt
```diff
@@ -0,0 +1,77 @@
...
+        addHeader("content-type", "text/plain;charset=utf-8")
+        addHeader("date", "Wed, 22 May 2024 15:04:50 GMT")
+        addHeader("expires", "0")
+        addHeader("pragma", "no-cache")
+        addHeader("response-status-code", "200")
+        addHeader("strict-transport-security", "max-age=31536000")
+        addHeader("transfer-encoding", "chunked")
+        addHeader("vary", "X-Esri-Authorization")
+        addHeader("x-content-type-options", "nosniff")
+    }.build()
```
gunt0001, line 77: missing newline

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/UiDeviceExtensions.kt
```diff
@@ -0,0 +1,80 @@
+package com.arcgismaps.toolkit.authentication
```
gunt0001, line 1: missing copyright info

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/UiDeviceExtensions.kt
```diff
@@ -0,0 +1,80 @@
...
+import androidx.test.uiautomator.Until
+import org.junit.Assert
+
+
+/**
+ * Waits for the View with the matching package to be visible. Throws an error if the view can't be
+ * found.
+ *
+ * @param packageId the view's package Id to wait for.
+ * @since 200.4.0
```
gunt0001, line 19: Needs updating also in other places in this file...
```suggestion
 * @since 200.5.0
```

gunt0001 (approved):
Approving so this can move to second review once comments have been addressed.

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/UiDeviceExtensions.kt
```diff
@@ -0,0 +1,96 @@
...
+        )
+    }
+}
+
+/**
+ * Enters the [username] and [password] in the OAuth login page.
+ *
+ * @since 200.5.0
+ */
+fun UiDevice.enterCredentialsOnBrowser(username: String, password: String, activity: Activity) {
```
alan-edi, line 55: Seems this function, and the functions below that it calls, isn't used, but I guess you expect to use it later when we come up with a solution for test credentials in the toolkit?

alan-edi (approved):
@hud10837 Looks good, just a few minor comments to consider

alan-edi:
Oh, one other thing - there are no tests for the following sub-cases, but I guess you have good reasons:
- 1.2 Incorrect Credentials - maybe do it later when we come up with a solution for test credentials in the toolkit?
- 1.5 Pressing the Close Button - is it practicable to automate that?
- 1.6 Press the Home Button - again, is it practicable?

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/UiDeviceExtensions.kt
hud10837: I will remove them for now, can always return to this PR to copy them.

hud10837:
@alan-edi thanks for the review. Regarding those test cases:

- 1.2 Incorrect Credentials - I'm not sure there's any value in testing this (also true for the manual tests) because there is no interaction with the `Authenticator` when you enter invalid credentials -- everything continues to happen in the browser, and if you enter the wrong credentials too many times, it keeps the browser open and locks you out for 15 minutes, so there is no "invalid credentials" redirect back to the app.
- 1.5 Pressing the Close Button - It's possible to automate this, but a bit difficult, and the behavior should be identical to the back button. But I am happy to add this test case if you think it's a good idea
- 1.6 Press the Home Button - The goal of this in the manual tests was to test the activity being destroyed and put in the background, then recreated, which also happens during configuration changes. I think this is very difficult to automate though, so I think that testing only device rotation is sufficient (as it should be the same behavior anyway) What do you think?

alan-edi:
@hud10837 I'm happy with what you say re test cases 1.2, 1.5 and 1.6, thanks

hud10837:
Thanks @alan-edi @gunt0001 !

PR #434, opened 2024-05-16
`Forms` Consume new attachments api

kaushikrw:
### Summary of changes

- Uses the latest daily with new api changes.
- Uses the `attachmentChanged` event to handle deletion and renaming of attachments.
- Changed the type of state holder for the `AttachmentElementState.attachments` from a `Map` to a `List`. This is because the `Map` does not preserve insertion order without an additional overhead. Using a list preserves the insertion order but increases the time complexity for deletion and rename to _O(n)_.
- The `AttachmentElementState.Saver` does to store load statuses of its `FormAttachmentState`. Instead, this can be inferred by looking at the `FormAttachment` objects when building the state list.
- `FormAttachmentState` specifies a smaller size for the thumbnail to save on memory, which increased the UI responsiveness quite a bit.


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
```diff
@@ -94,92 +99,130 @@ internal class AttachmentElementState(
...
-        formElement.addAttachment(name, contentType, data)
-        evaluateExpressions()
-        // refresh the list of attachments
-        loadAttachments()
-        // load the new attachment
-        attachments.entries.last().value.loadWithParentScope()
-        // scroll to the new attachment after a delay to allow the recomposition to complete
-        delay(100)
-        lazyListState.scrollToItem(attachments.count())
+        formElement.addAttachment(name, contentType, data).onSuccess { formAttachment ->
```
sorenoid, line 171: why is the addition to the state list done here and not in response to the added change event like rename and delete?

sorenoid (commented):
I have a question. Looks good, but I don't understand why the add operation is done differently than the delete and rename operation.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
kaushikrw: @sorenoid fixed!

sorenoid (approved):

PR #444, opened 2024-05-27
`Forms` : Optimize thumbnail loading for Attachments

kaushikrw:
### Issue

https://devtopia.esri.com/runtime/apollo/issues/691

### Summary of changes

- Uses platform methods to generate thumbnails which offers correct orientation compared to the sdk method. This also generates thumbnails for videos.
- The actual attachment and thumbnail files are now prefixed with the `attachment.id` so that the file names are unique. The `id` parameter also participates in `hashCode` and `equals` for the `FormAttachmentState`.
- Optimized how the thumbnails are loaded:
    - Instead of loading all thumbnails into memory by storing them as a state object, they are instead serialized into a file.
    - Coil's `AsyncImage` is then used to load these thumbnails from file offloading the overhead to coil which is more optimized.
    - This reduces the memory footprint of the component.

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
@@ -19,6 +19,10 @@ package com.arcgismaps.toolkit.featureforms.internal.components.attachment
 import android.Manifest
 import android.content.Context
 import android.content.pm.PackageManager
+import android.graphics.Bitmap.CompressFormat
+import android.media.ThumbnailUtils
+import android.os.Build
+import android.util.Size
 import androidx.compose.foundation.lazy.LazyListState
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.outlined.AudioFile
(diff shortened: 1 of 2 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
```diff
@@ -282,15 +286,20 @@ internal class FormAttachmentState(
...
-
     /**
-     * The thumbnail of the attachment. This is `null` until [load] is called.
+     * A unique ID for the attachment.
      */
-    val thumbnail: State<ImageBitmap?> = _thumbnail
+    val id by lazy {
+        formAttachment?.attachment?.let {
+            "${it.id}_${it.name}"
+        } ?: name
```
sorenoid, line 301: is name actually unique in the null case? I guess it should never happen?

sorenoid (approved):
LGTM

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
kaushikrw: I think it should not be null, at least after a successful load. But it should probably be a getter instead. I will update it.

kaushikrw:
thanks @sorenoid

PR #448, opened 2024-05-29
Authenticator: automate username password tests

hud10837:
Converts manual tests for the `UsernamePasswordAuthenticator` into automated tests.

Like the `OAuth` tests, there is one test here that fakes responses from the server in order to verify that ArcGISToken authentication works. This test for ArcGIS challenges is not present in the manual tests (but it should have been). The other tests use `NetworkAuthenticationChallengeHandler` in the same manner as the `ServerTrustTests`, ie. they don't do actual network calls but just ensure that the `Authenticator` UI works as expected.


colinanderson (approved):
@hud10837 I don't understand it all 100% but it looks ok.

hud10837:
Thanks @colinanderson

eri9000 (approved):
LGTM 👍🏼

PR #449, opened 2024-05-29
popup media element and media image thumbnail support

sorenoid:
This PR provides a MediaPopupElement composable and support for PopupMedia of type Image (Charts will be supported next).

Coil is used for image loading though it is abstracted to support asynchronous loading of charts through core, so the use of coil is a bit indirect relative to what might be expected.

Images are displayed at 276x225 with title and caption overlaid.


eri9000 (approved):

puneet-pdx (approved):
@sorenoid Just a comment. Looks good overall.

PR #453, opened 2024-05-30
`Forms` : Consume latest api

kaushikrw:
https://devtopia.esri.com/runtime/apollo/issues/707

## Summary

- `FeatureForm.defaultAttachmentsElement` is used to show an `AttachmentsFormElement` if the feature supports attachments. This is to support attachments outside the form authoring experience.
-  Updated add, delete and rename attachment to be non-suspending functions.
- Rename now happens through the `FormAttachment.name` property and hence the `FormAttachmentState.name` is now an observable property backed by a `MutableState`.
- Loading an attachment now doesn't need to write the data into a new file. Instead, the `FormAttachment.filePath` is used directly to view the file. To support this, the `cacheDir` is used in the `FileProvider` since this is where core stores the loaded attachements.
- Rename and Delete context menu can now be triggered with a long press without having to load the attachment.
- Discarding and submitting attachments now works with the use of the new `FeatureForm.finishEditsAsync` call.
- Loading progress animation is now visible when trying to add an attachment via a file picker or the gallery.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -54,4 +54,4 @@ ignoreBuildNumber=false
 # these versions define the dependency of the ArcGIS Maps SDK for Kotlin dependency
 # and are generally not overridden at the command line unless a special build is requested.
 sdkVersionNumber=200.5.0
-sdkBuildNumber=4244
+sdkBuildNumber=4253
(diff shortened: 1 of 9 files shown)


sorenoid (approved):
👍

kaushikrw:
thanks @sorenoid

PR #461, opened 2024-06-10
`Popup:` Fixes TextPopupElement composable shifting in size when coming back into visibility

eri9000:
### Description:
- When a TextPopupElement is scrolled out of view and back into view, its width is recalculated multiple times and it can cause it to push other elements out of view.

### DIFF
--- a/toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/Popup.kt
+++ b/toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/Popup.kt
@@ -153,6 +153,7 @@ private fun PopupBody(popupState: PopupState, onFileClicked: (ViewableFile?) -> 
             val element = entry.popupElement
             when (element) {
                 is TextPopupElement -> {
+                    // a contentType is needed to reuse the TextPopupElement composable inside a LazyColumn
                     item(contentType = TextPopupElement::class.java) {
                         TextPopupElement(
                             entry.state as TextElementState
--- a/toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/Popup.kt
+++ b/toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/Popup.kt
@@ -18,6 +18,7 @@
 
 package com.arcgismaps.toolkit.popup
 
+import android.webkit.WebView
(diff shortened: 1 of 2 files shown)


toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/textelement/TextPopupElement.kt
```diff
@@ -81,7 +81,9 @@ private fun HTML(content: String) {
             val completeHtml = "$header$headStyle<html>${content.trim()}</html>"
             loadDataWithBaseURL(null, completeHtml, "text/html", "UTF-8", null)
         }
-    })
+    },
```
puneet-pdx, line 84: Worth adding a comment here to explain the empty implementation. Something on the lines which states, 
```
By default, AndroidView does not automatically pool or reuse Views. If placed inside of a reusable container (including inside a LazyRow or LazyColumn), the View instances will always be discarded and recreated if the composition hierarchy containing the AndroidView changes, even if its group structure did not change and the View could have conceivably been reused.
If onReset is null or not specified, the AndroidView instance will not support reuse, and the View instance will always be discarded whenever the AndroidView is moved or removed from the composition hierarchy.
```

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/Popup.kt
```diff
@@ -150,35 +150,42 @@ private fun PopupBody(popupState: PopupState, onFileClicked: (ViewableFile?) ->
...
+                    item(contentType = FieldsPopupElement::class.java) {
                         FieldsPopupElement(
                             entry.state as FieldsElementState,
                         )
                     }
+                }
 
-                    is MediaPopupElement -> {
+                is MediaPopupElement -> {
+                    item(contentType = MediaPopupElement::class.java) {
```
puneet-pdx, line 180: Add comment for contentType addition? 
```
contentType is needed for reuse of AndroidView inside a LazyColumn
```

puneet-pdx (approved):
@eri9000 Looks good. Just some suggestions to add doc for these additions, won't hold the PR for those.

shubham7109 (approved):
Looks good 👍

PR #466, opened 2024-06-12
`Callout` - Support SceneView

puneet-pdx:
Related to issue: # https://devtopia.esri.com/runtime/kotlin/issues/4002

### Description:

Changes to show Callout on the SceneView

### Summary of changes:

- Add content lambda to SceneView
- Define GeoViewScope and add the Callout implementation to GeoViewScope
- Move GeoViewScope to separate file
- Add MapViewScope to MapView.kt and SceneViewScope to SceneView.kt
- Add MicroApp to show Callout on the SceneView

### Pre-merge Checklist

  - [ ] link:
  - [ ] Yes
  - [x] No - includes micro app


toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,499 @@
...
+        }
+
+        if (!isGeoViewReady.value) {
+            return
+        }
+
+        // Convert the given location to a screen coordinate
+        var leaderScreenCoordinate: ScreenCoordinate? by remember {
+            mutableStateOf(
+                getLeaderScreenCoordinate(geoView, calloutParams.location!!, calloutParams.offset, calloutParams.rotateOffsetWithGeoView)
```
gunt0001, line 98: I'm seeing several places where non-null assert `!!` is being used on `calloutParams`, which doesn't seem right. In the case of `calloutParams.location` returning null, we should fall back to some default value. I assume for the other cases the `!!` should also be avoided.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,499 @@
...
+        LaunchedEffect(calloutParams.location) {
+            // Used to update screen coordinate when new location point is used
+            leaderScreenCoordinate = getLeaderScreenCoordinate(geoView, calloutParams.location!!, calloutParams.offset, calloutParams.rotateOffsetWithGeoView)
+            // Used to update screen coordinate when viewpoint is changed
+            geoView.viewpointChanged.collect {
+                leaderScreenCoordinate = getLeaderScreenCoordinate(geoView, calloutParams.location!!, calloutParams.offset, calloutParams.rotateOffsetWithGeoView)
+            }
+        }
+
+        val localDensity = LocalDensity.current
```
gunt0001, line 111: Is there a need for declaring a local variable here? I think you can just call `LocalDensity.current` inline where needed.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,499 @@
...
+        // Get the default shape, color & size properties for Callout
+        val properties = CalloutProperties()
```
gunt0001, line 113: You are creating a new instance of `CalloutProperties` with every recomposition, which is inefficent. Seems like you do this only to get hold of some default properties. Instead, you can declare the defaults as static properties, such as:
```kotlin
        DefaultCalloutProperties.cornerRadius
        DefaultCalloutProperties.strokeBorderWidth
        ...
```

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,499 @@
...
+        leaderScreenCoordinate?.let {
+            CalloutSubComposeLayout(
+                leaderScreenCoordinate = it,
+                maxSize = calloutContentMaxSize(
+                    geoView = geoView,
+                    density = LocalDensity.current,
+                    displayMetrics = LocalContext.current.resources.displayMetrics
+                )) {
+                Box(
+                    modifier = calloutParams.modifier!!
```
gunt0001, line 123: Non-null assert `!!` should not be used here.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,499 @@
...
+        calloutParams = CalloutParams()
+    }
+
+    /**
+     * Creates a Callout at the specified geographical location on the GeoView.
+     *
+     * @since 200.5.0
+     */
+    @Composable
+    internal fun Callout() {
```
gunt0001, line 74: I wonder if the `content` lambda parameter should be passed to this function rather than stashing it away in `CalloutParams`. It doesn't seem intuitive that way.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,499 @@
...
+     * @since 200.5.0
+     */
+    @Composable
+    private fun CalloutSubComposeLayout(
+        modifier: Modifier = Modifier,
+        leaderScreenCoordinate: ScreenCoordinate,
+        maxSize: DpSize,
+        calloutContainer: @Composable () -> Unit
+    ) {
+        val configuration = LocalDensity.current
```
gunt0001, line 199: As above, is there a need for this local variable?

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,499 @@
...
+            layout(layoutWidth, layoutHeight) {
+                calloutContainerPlaceable.place(calloutOffsetX, calloutOffsetY)
+            }
+        }
+    }
+
+    /**
+     * UI default properties for the [Callout] component.
+     */
+    private data class CalloutProperties(
```
gunt0001, line 239: Seems like the way this is used, it should be an `object`?
```suggestion
    private object DefaultCalloutProperties(
```

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/MapView.kt
```diff
@@ -523,6 +523,13 @@ public inline fun rememberLocationDisplay(
     }
 }
 
+/**
+ * The receiver class of the [MapView] content lambda.
+ *
+ * @since 200.5.0
+ */
+public class MapViewScope(_mapView: MapView?) : GeoViewScope(_mapView, "MapView")
```
gunt0001, line 531: Constructor must be internal, parameter should not start with `_`.
Why is the parameter nullable?
```suggestion
public class MapViewScope internal constructor(mapView: MapView) : GeoViewScope(mapView, "MapView")
```

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,499 @@
...
+import kotlin.math.roundToInt
+import kotlin.math.sin
+
+/**
+ * The receiver class of the MapView/SceneView content lambda.
+ *
+ * @since 200.5.0
+ *
+ */
+public sealed class GeoViewScope(private var _geoView: GeoView?, className: String) {
```
gunt0001, line 55: Better mark the constructor as protected. Why is `geoView` nullable and mutable?
```suggestion
public sealed class GeoViewScope protected constructor(private val geoView: GeoView, className: String) {
```

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/SceneView.kt
```diff
@@ -452,6 +465,12 @@ private fun ViewpointHandler(
     }
 }
 
+/**
+ * The receiver class of the [SceneView] content lambda.
+ *
+ * @since 200.5.0
+ */
+public class SceneViewScope(private var _sceneView: SceneView?) : GeoViewScope(_sceneView, "SceneView")
```
gunt0001, line 473: Same comments as above for MapViewScope constructor.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/SceneView.kt
```diff
@@ -156,47 +160,56 @@ public fun SceneView(
...
+                        addAll(imageOverlays)
+                    }
                 }
-            }
-        })
+            })
+
+        val sceneViewScope = remember(sceneView) { SceneViewScope(sceneView) }
+        content?.let {
+            sceneViewScope.reset()
```
gunt0001, line 209: What is this reset call all about?

microapps/SceneViewCalloutApp/app/build.gradle.kts
```diff
@@ -0,0 +1,95 @@
...
+plugins {
+    id("com.android.application")
+    id("org.jetbrains.kotlin.android")
+    id("com.google.android.libraries.mapsplatform.secrets-gradle-plugin")
+}
+
+secrets {
+    // this file doesn't contain secrets, it just provides defaults which can be committed into git.
+    defaultPropertiesFileName = "secrets.defaults.properties"
+}
```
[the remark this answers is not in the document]
puneet-pdx: This is being carried over from the template-app project. All other micro apps have it. I would rather create an issue for it and remove it from all the micro-apps together, rather than removing it just here. Let me know if you are ok with going that route.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
puneet-pdx: The constructor of sealed class is protected by default, so didn't specify it explicitly here. If you still think we should be explicit I can change it. 
_Constructors of sealed classes can have one of two [visibilities](https://kotlinlang.org/docs/visibility-modifiers.html): protected (by default) or private:_ 

I don't think it has to be mutable, I have updated it. We made it nullable so we can throw the geoview not initialized error if the user tries to use the callout outside of the geoview.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,499 @@
...
+
+    private val nullGeoViewErrorMessage: String =
+        "$className not initialized. Make sure to use the composable within the MapView or SceneView."
+
+    internal var calloutParams: CalloutParams = CalloutParams()
+    private val geoView: GeoView
+        get() = _geoView ?: error(nullGeoViewErrorMessage)
+
+    internal fun reset() {
```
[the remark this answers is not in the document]
puneet-pdx: I have added doc for it.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
puneet-pdx: We are holding all the callout properties in the CalloutParams, content being one of them. I am not sure if just passing one of them to this method will make sense. Let discuss maybe ?

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,499 @@
...
+     *
+     * @since 200.5.0
+     */
+    @Composable
+    internal fun Callout() {
+
+        val isGeoViewReady = remember { mutableStateOf(false) }
+        // We don't want to start drawing the Callout until the GeoView is ready. We only collect
+        // the drawStatus till the first time GeoView is done drawing. The transformWhile operator
+        // will stop collecting when isGeoViewReady.value becomes false.
```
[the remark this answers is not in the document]
puneet-pdx: In the case when the user specifies the Callout with the MapView, we encounter a race condition where the Callout code is executed before the MapView is done drawing. In that case the Callout does not show up on the MapView as core does not return the location of the callout on the screen. This logic is to handle that scenario.
This only executed once when the mapView draws for the first time and does not collect after that.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
puneet-pdx: This is something we hadn't ironed out yet as the implementation was still evolving. I have updated it.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,499 @@
...
+        }
+
+        // Convert the given location to a screen coordinate
+        var leaderScreenCoordinate: ScreenCoordinate? by remember {
+            mutableStateOf(
+                getLeaderScreenCoordinate(geoView, calloutParams.location!!, calloutParams.offset, calloutParams.rotateOffsetWithGeoView)
+            )
+        }
+
+        LaunchedEffect(calloutParams.location) {
```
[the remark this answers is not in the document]
puneet-pdx: This part is not reachable on purpose till the one in line 80 is active.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
puneet-pdx: updated

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
puneet-pdx: updated

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
puneet-pdx: I guess not, updated

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
puneet-pdx: updated

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/MapView.kt
puneet-pdx: This is just some oversight on my part, from trying to port over the changes in the inheritance hierarchy from the existing implementation. I have updated it.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/SceneView.kt
puneet-pdx: It is to reset the CalloutParameters on recompositions. If we don't explicitly reset the CalloutParameters then when a recomposition is triggered and CalloutParameters are holding on to the previous values the Callout will not draw as we restrict the drawing of multiple callouts by determining if the values on CalloutParameters are already set or not

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/SceneView.kt
puneet-pdx: updated

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/Callout.kt
```diff
@@ -82,7 +43,7 @@ import kotlin.math.sin
  * @since 200.5.0
  */
 @Composable
-public fun MapViewScope.Callout(
+public fun GeoViewScope.Callout(
```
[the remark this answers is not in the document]
puneet-pdx: well you can do it but it will not be very compose-y. Composables are generally independent functions defined in their own file.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,499 @@
...
+/**
+ * The receiver class of the MapView/SceneView content lambda.
+ *
+ * @since 200.5.0
+ *
+ */
+public sealed class GeoViewScope(private var _geoView: GeoView?, className: String) {
+
+    private val nullGeoViewErrorMessage: String =
+        "$className not initialized. Make sure to use the composable within the MapView or SceneView."
```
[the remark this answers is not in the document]
puneet-pdx: Updated. No longer needed

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
puneet-pdx: Updated.

microapps/SceneViewCalloutApp/app/build.gradle.kts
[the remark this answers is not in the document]
gunt0001: As discussed, lets keep it. It's in the app template and other micro apps use secrets, such as FeatureForms.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/Callout.kt
```diff
@@ -82,462 +43,15 @@ import kotlin.math.sin
...
-        arcTo(
-            rect = topLeftCorner,
-            startAngleDegrees = 180f,
-            sweepAngleDegrees = 90f,
-            forceMoveTo = false
-        )
-        // Close the path to complete the shape
-        close()
+    if (!this.isCalloutBeingDisplayed) {
+        isCalloutBeingDisplayed = true
```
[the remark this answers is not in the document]
puneet-pdx: I cannot be private as it is being used in MapView and SceneView to reset the previous state of the Callout.
https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/pull/466/files#diff-87ef0131b55a25ad4995e440b69be9a2acf84f8a00918bcc48e0bb9de308c35cR202

https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/pull/466/files#diff-3e43173786e6ec207ba12b41a06ba17cfa918493e900f0e50fad6f99b2e7c8dbR209

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,504 @@
...
+ * @since 200.5.0
+ */
+public sealed class GeoViewScope protected constructor(private val geoView: GeoView) {
+
+    /**
+     * Used to restrict only one Callout to be displayed at a time.
+     *
+     * @since 200.5.0
+     */
+    internal var isCalloutBeingDisplayed: Boolean = false
```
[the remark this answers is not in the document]
puneet-pdx: updated

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,504 @@
...
+        LaunchedEffect(location) {
+            geoView.drawStatus.transformWhile { drawStatus ->
+                emit(drawStatus)
+                !isGeoViewReady.value
+            }.collect {
+                if (it == DrawStatus.Completed) {
+                    isGeoViewReady.value = true
+                }
+            }
+        }
```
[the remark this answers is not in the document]
puneet-pdx: Updated

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,504 @@
...
+        val isGeoViewReady = remember { mutableStateOf(false) }
+        // We don't want to start drawing the Callout until the GeoView is ready. We only collect
+        // the drawStatus till the first time GeoView is done drawing. The transformWhile operator
+        // will stop collecting when isGeoViewReady.value becomes false.
+        LaunchedEffect(location) {
+            geoView.drawStatus.transformWhile { drawStatus ->
+                emit(drawStatus)
+                !isGeoViewReady.value
+            }.collect {
+                if (it == DrawStatus.Completed) {
```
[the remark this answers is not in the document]
puneet-pdx: yes, that works. I have update it. Now we will start drawing the callout as we get the first InProgress status.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -0,0 +1,504 @@
...
+        var leaderScreenCoordinate: ScreenCoordinate? by remember {
+            mutableStateOf(
+                getLeaderScreenCoordinate(geoView, location, offset, rotateOffsetWithGeoView)
+            )
+        }
+
+        LaunchedEffect(location) {
+            // Used to update screen coordinate when new location point is used
+            leaderScreenCoordinate =
+                getLeaderScreenCoordinate(geoView, location, offset, rotateOffsetWithGeoView)
```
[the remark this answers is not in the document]
puneet-pdx: No it is not. It is needed when the user taps on the screen to show callout at a different location.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -59,126 +58,135 @@ import com.arcgismaps.mapping.view.SceneView
...
+            geoView.drawStatus.transformWhile<DrawStatus, DrawStatus> { drawStatus ->
+                if (drawStatus == DrawStatus.InProgress) {
+                    onGeoViewReady(true)
+                    isGeoViewReady = true
                 }
+                !isGeoViewReady
+            }.collect {
             }
         }
+    }
```
gunt0001, line 133: Although this is working it could be improved syntactically. Since you call it "await", I would expect it to be a suspend function but it is not. As discussed, using the "remember pattern" would be more standard here. I suggest to implement this function as an extension function on `GeoView` (placed here in the GeoViewScope.kt file):
```kotlin
@Composable
internal fun GeoView.rememberIsReady(): State<Boolean> {
    var isGeoViewReady = remember { mutableStateOf(false) }
    LaunchedEffect(Unit) {
        this@rememberIsReady.drawStatus.first()
        isGeoViewReady.value = true
    }
    return isGeoViewReady
}
``` 

Then on the call side it would be used something like this:
```kotlin
val isSceneViewReady = sceneView.rememberIsReady()
if (isSceneViewReady.value) {
   content?.let {
      sceneViewScope.isCalloutBeingDisplayed.set(false)
      sceneViewScope.it()
   }
}
```

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/MapView.kt
```diff
@@ -395,6 +404,7 @@ private fun MapViewEventHandler(
         }
         launch {
             mapView.drawStatus.collect { drawStatus ->
+//                Log.e("mapView.drawStatus", "in mapView drawStatus: $drawStatus")
```
gunt0001, line 407: remove?

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/SceneView.kt
```diff
@@ -156,47 +160,63 @@ public fun SceneView(
...
-            it.spaceEffect = spaceEffect
-            it.sunTime = sunTime
-            it.sunLighting = sunLighting
-            it.ambientLightColor = com.arcgismaps.Color(ambientLightColor.toArgb())
-            it.isAttributionBarVisible = isAttributionBarVisible
-            if (it.graphicsOverlays != graphicsOverlays) {
-                it.graphicsOverlays.apply {
-                    clear()
-                    addAll(graphicsOverlays)
+    Box(modifier = modifier.clipToBounds()) {
```
gunt0001, line 170: Why is this Box necessary? Needs some commenting. (same in MaoView())

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/SceneView.kt
```diff
@@ -156,47 +160,63 @@ public fun SceneView(
...
-                    addAll(imageOverlays)
+                if (it.imageOverlays != imageOverlays) {
+                    it.imageOverlays.apply {
+                        clear()
+                        addAll(imageOverlays)
+                    }
                 }
+            })
+
+        val sceneViewScope = remember(sceneView) { SceneViewScope(sceneView) }
```
gunt0001, line 207: I believe there is no need to pass a key to remember here:
```suggestion
        val sceneViewScope = remember { SceneViewScope(sceneView) }
```

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/MapView.kt
```diff
@@ -198,9 +200,16 @@ public fun MapView(
             })
 
         val mapViewScope = remember(mapView) { MapViewScope(mapView) }
```
gunt0001, line 202: As with SceneView, the key shouldn't be necessary:
```suggestion
        val mapViewScope = remember { MapViewScope(mapView) }
```

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/SceneView.kt
puneet-pdx: updated

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/MapView.kt
puneet-pdx: updated

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -541,3 +525,22 @@ private fun calloutPath(
...
+ * This function is used to wait for the GeoView to be ready to return positive values
+ * for operations like locationToScreen. We determine that by waiting for the first drawStatus
+ * message when the map/scene is rendered on the GeoView.
+ * For the MapView/SceneView's content parameter like the Callout we don't want to start drawing
+ * the Callout until the GeoView is ready.
+ *
+ * @since 200.5.0
+ */
+@Composable
+public fun GeoView.rememberIsReady(): State<Boolean> {
```
gunt0001, line 539: I think we should make it internal for now, unless we know of a public use case:
```suggestion
internal fun GeoView.rememberIsReady(): State<Boolean> {
```

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/MapView.kt
```diff
@@ -173,9 +173,13 @@ public fun MapView(
     val mapView = remember { MapView(context) }
     val layoutDirection = LocalLayoutDirection.current
 
+    // The MapView is wrapped in a Box to ensure that the Callout is drawn on top of the MapView and
+    // that the Callout is clipped to its bounds
```
gunt0001, line 177: I assume this is required so it works with the `SubcomposeLayout`? If so. might be good to mention SubcomposeLayout here.

gunt0001 (approved):

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/MapView.kt
puneet-pdx: Not really. We use the SubcomposeLayout to determine the size of the content composable being passed to the Callout, so we can anchor the Callout at the right location on the map.

sorenoid (approved):
👍

puneet-pdx:
Thanks! @sorenoid @gunt0001

PR #473, opened 2024-06-20
`Forms` : Attachments horizontal scrollbar

kaushikrw:
Related to issue: #[apollo/670](https://devtopia.esri.com/runtime/apollo/issues/670)

### Description: 

Adds a horizontal scrollbar to the list of attachments.

### Summary of changes:

- Created a customizable `horizontalScrollbar` modifier which works with a `LazyRow` to show a horizontal scroll bar.
- If there are more attachments and the list is scrollable, the scroll bar becomes visible under the list. If the list is not scrollable, the scroll bar disappears.

### Pre-merge Checklist

  - [ ] link:
  - [ ] Yes
  - [ ] No


sorenoid (approved):
this looks great. There seems to be a small change in scroll when scrolled all the way to the right and the device is rotated and then rotated back to original orientation. Not a big deal! Just thought I would point it out.

kaushikrw:
> this looks great. There seems to be a small change in scroll when scrolled all the way to the right and the device is rotated and then rotated back to original orientation. Not a big deal! Just thought I would point it out.

Thanks @sorenoid. I think that could be from the viewport size changing, I will make a note.

PR #514, opened 2024-07-15
`Forms` : fix thumbnails mismatch

kaushikrw:
Related to issue: #[apollo/781](https://devtopia.esri.com/runtime/apollo/issues/781)

### Description:

Fixes an issue where thumbnails for a `FormAttachment` are re-used. This PR refactors the thumbnail creation logic to fix the issue.

### Summary of changes:

- Thumbnails are no longer saved to a file. Instead they are loaded directly into the state object `FormAttachmentState`. 
- This is because the `Attachment.id` property is not unique when the attachment has just been added and is not yet synced with the service. Hence there is no unique property available to name the thumbnail file.

### Pre-merge Checklist

  - [ ] link:
  - [ ] Yes
  - [ ] No

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
@@ -19,7 +19,7 @@ package com.arcgismaps.toolkit.featureforms.internal.components.attachment
 import android.Manifest
 import android.content.Context
 import android.content.pm.PackageManager
-import android.graphics.Bitmap.CompressFormat
+import android.graphics.Bitmap
 import android.media.ThumbnailUtils
 import android.os.Build
 import android.util.Size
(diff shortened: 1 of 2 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
sorenoid, line 1: Is it not possible to inject something unique into the FormAttachmentState at creation time, like a UUID, then use that as the name of the file on disk?

sorenoid (commented):
Just a question for now.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
kaushikrw: That UUID would need be the same every time to re-fetch the thumbnail on the next launch of the Form. Otherwise, it would end up creating the thumbnail many times as the form is closed and re-opened.

sorenoid (approved):

PR #523, opened 2024-07-19
Disable UsernamePassword Login button when fields are empty

eri9000:
Related to issue: #

### Description:
- Improvements to the Username/Login prompt 

### Summary of changes:
- Disables the "Login" button when the text is empty
- Disallows submitting empty username/password with keyboard action button 
- Adds test to verify behavior 
 
### Pre-merge Checklist

  - [ ] link: https://runtime-kotlin.esri.com/view/vTest/job/vtest/job/toolkit/53
  - [ ] Yes
  - [ ] No

### DIFF
--- a/toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/UsernamePasswordTests.kt
+++ b/toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/UsernamePasswordTests.kt
@@ -5,12 +5,16 @@ import androidx.activity.ComponentActivity
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.assertIsEnabled
 import androidx.compose.ui.test.assertIsNotEnabled
+import androidx.compose.ui.test.hasImeAction
 import androidx.compose.ui.test.junit4.AndroidComposeTestRule
 import androidx.compose.ui.test.junit4.StateRestorationTester
 import androidx.compose.ui.test.junit4.createAndroidComposeRule
 import androidx.compose.ui.test.onNodeWithText
 import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.performImeAction
+import androidx.compose.ui.test.performTextClearance
 import androidx.compose.ui.test.performTextInput
+import androidx.compose.ui.text.input.ImeAction
 import androidx.test.platform.app.InstrumentationRegistry
 import com.arcgismaps.ArcGISEnvironment
 import com.arcgismaps.httpcore.authentication.ArcGISAuthenticationChallengeResponse
@@ -18,13 +22,18 @@ import com.arcgismaps.httpcore.authentication.NetworkAuthenticationChallenge
 import com.arcgismaps.httpcore.authentication.NetworkAuthenticationChallengeResponse
 import com.arcgismaps.httpcore.authentication.NetworkAuthenticationType
 import com.arcgismaps.httpcore.authentication.TokenCredential
+import io.mockk.Runs
 import io.mockk.coEvery
+import io.mockk.every
+import io.mockk.just
 import io.mockk.mockk
 import io.mockk.mockkObject
 import io.mockk.unmockkAll
+import io.mockk.verify
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.async
+import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.test.TestScope
 import kotlinx.coroutines.test.advanceUntilIdle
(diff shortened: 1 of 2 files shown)


XuemingRocks (commented):
Overall looks good. However, I can still repro with the following workflow:

1. Input username
2. Tap on password field. The Send button at the bottom right corner of the keyboard is enabled.
3. Tap on the Send button.
image

eri9000:
@XuemingRocks Good catch, I have pushed a fix for that. Have another look.

XuemingRocks (approved):
LGTM!

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/UsernamePasswordAuthenticator.kt
```diff
@@ -113,11 +113,13 @@ public fun UsernamePasswordAuthenticator(
...
-                    usernameFieldText,
-                    passwordFieldText
-                )
-                passwordFieldText = ""
+                if (usernameFieldText.isNotEmpty() && passwordFieldText.isNotEmpty()) {
+                    usernamePasswordChallenge.continueWithCredentials(
+                        usernameFieldText,
+                        passwordFieldText
+                    )
+                    passwordFieldText = ""
```
gunt0001, line 121: Why is the password text being reset here? Could add a comment.

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/UsernamePasswordAuthenticator.kt
```diff
@@ -136,7 +138,6 @@ public fun UsernamePasswordAuthenticator(
                     keyboardType = KeyboardType.Email,
                     imeAction = ImeAction.Next
                 ),
-                keyboardActions = keyboardActions,
```
gunt0001, line 139: Why this change?

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/UsernamePasswordTests.kt
```diff
@@ -84,6 +86,31 @@ class UsernamePasswordTests {
         assert(response is NetworkAuthenticationChallengeResponse.ContinueWithCredential)
     }
 
+    /**
+     * Given a Dialog Authenticator
```
gunt0001, line 90: Maybe better:
```suggestion
     * Given a UserNamePasswordAuthenticator
```

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/UsernamePasswordTests.kt
```diff
@@ -84,6 +86,31 @@ class UsernamePasswordTests {
...
+     * When the username or password fields are empty
+     * Then the login button should be disabled
+     * And when the username and password fields are both filled
+     * Then the login button should be enabled
+     *
+     * @since 200.5.0
+     */
+    @Test
+    fun loginButtonEnabledState() = runTest {
+        val response = testUsernamePasswordChallengeWithStateRestoration {
```
hud10837, line 100: Does this need to test the full workflow? I think the test would be faster and more focused by just using `composeTestRule.setContent { UsernamePasswordAuthenticator(UsernamePasswordChallenge(...), modifier = Modifier.testTag("testUsernamePasswordAuthenticator")) } and performing the below verifications direclty there rather than actually logging in.

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/UsernamePasswordAuthenticator.kt
hud10837: This was already being done, to erase the password when submitting as many forms do so if the credentials were wrong then you will have to re-enter the password. I suppose this is a design consideration that we never fully discussed.

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/UsernamePasswordAuthenticator.kt
hud10837: I am assuming it's so that you can no longer press send on the keyboard in order to submit the username and password, but I would assume that this case would be handled by the check added to `submitUsernamePassword`, I'm not sure why Xueming was seeing the submission work out. If necessary I think better to add an additional check to the `onSend` in the keyboard actions.

hud10837 (commented):
Good change, just a few thoughts.

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/UsernamePasswordAuthenticator.kt
eri9000: This was discussed and we agreed it was ok to remove since this `OutlinedTextField` sets the keyboard `imeAction` to `Next` and the `keyboardActions` object modifies `onSend` so it does not directly utilize this anyway.

gunt0001 (approved):
@eri9000 - great test coverage enhancements 👍

hud10837 (approved):
LGTM 👍

PR #524, opened 2024-07-19
Callout: Automated tests

shubham7109:
Related to issue: [#4118](https://devtopia.esri.com/runtime/kotlin/issues/4118)

### Description:
PR to add Callout automated test cases.

### Summary of changes:
Added test cases:
- TestCalloutVisibility
- AttemptToDisplayMultipleCallouts
- TestSwitchingBetweenMultipleCallouts
- TestCalloutResetViaStateChanges
- TestCalloutTheming

### Pre-merge Checklist

[Updated Aug 12th]
  - [x] link: https://runtime-kotlin.esri.com/view/vTest/job/vtest/job/toolkit/59/Test_20Report_20-_20Integration/


toolkit/geoview-compose/src/androidTest/AndroidManifest.xml
```diff
@@ -0,0 +1,21 @@
...
+        <activity
+            android:name="com.arcgismaps.toolkit.geoviewcompose.callout.CalloutTestActivity"
+            android:exported="true"
+            android:launchMode="singleTop"
+            android:taskAffinity="">
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+                <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
+        </activity>
```
hud10837, line 19: is this still necessary?

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/callout/CalloutTestActivity.kt
```diff
@@ -0,0 +1,222 @@
...
+                arcGISMap = ArcGISMap(SpatialReference.wgs84()).apply {
+                    initialViewpoint = Viewpoint(39.8, -98.6, 10e7)
+                },
+                content = {
+                    val calloutLocation = mapViewModel.calloutLocation.collectAsState().value
+                    Callout(location = calloutLocation) {
+                        Text(text = "Hello World")
+                    }
+                    // update recomposition counter
+                    calloutRecompositionCount++
```
hud10837, line 177: When I run this test it always fails due to never being idle. I think what is happening is this line is causing recomposition of the MapView or the whole screen, which therefore is preventing the ComposeTestRule from ever recognizing the composition as idle and therefore it never stops waiting for the callout to display.

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/callout/CalloutTests.kt
```diff
@@ -0,0 +1,262 @@
...
+     *
+     * @since 200.5.0
+     */
+    @Test
+    fun testCalloutResetViaStateChanges() = runTest {
+        val calloutScenarios = CalloutScenarios()
+        composeTestRule.setContent {
+            CompositionLocalProvider {
+                val mapViewModel = compositionLocalOf { MapViewModel() }
+                calloutScenarios.ResetViaStateChanges(mapViewModel.current)
```
hud10837, line 182: Can't you do this with a normal viewmodel function?

```kotlin
val mapViewModel: MapViewModel = viewModel()
calloutScenarios.ResetViaStateChanges(mapViewModel)
```

hud10837: I also think during tests you can safely create the viewmodel outside the `setContent` scope but I'm not sure if that's recommended.

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/callout/CalloutTests.kt
```diff
@@ -0,0 +1,262 @@
...
+
+    /**
+     * GIVEN a Callout which is displayed on the MapView
+     * WHEN state changes are collected from the MapViewScope
+     * THEN verify Callout has been recomposed triggering a reset.
+     *
+     * @since 200.5.0
+     */
+    @Test
+    fun testCalloutResetViaStateChanges() = runTest {
```
hud10837, line 177: This test fails to me, see my comment in the corresponding scenario. I think it is infinitely recomposing.

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/callout/CalloutTestActivity.kt
hud10837, line 1: Should this file be renamed

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/callout/CalloutTests.kt
```diff
@@ -0,0 +1,262 @@
...
+            matcher = hasContentDescription(calloutContainerLabel),
+            timeoutMillis = timeoutMillis
+        )
+
+        // get the current recomposition count for the initial Callout
+        val initialRecompositionCount = composeTestRule
+            .onNodeWithTag(calloutScenarios.recompositionCounterLabel)
+            .fetchSemanticsNode()
+            .config
+            .text
```
hud10837, line 197: Once I commented out the line setting the recompositions to see if it would pass the test, this line fails for me with this message:

```
You cannot retrieve a semantics property directly - use one of the SemanticsConfiguration.getOr* methods instead
```
Maybe better to use a viewmodel to handle this logic.

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/callout/CalloutTests.kt
```diff
@@ -0,0 +1,262 @@
...
+
+        // get the current recomposition count after performing state change
+        val currentRecompositionCount = composeTestRule
+            .onNodeWithTag(calloutScenarios.recompositionCounterLabel)
+            .fetchSemanticsNode()
+            .config
+            .text
+
+        // check to see if Callout was recomposed via a Reset state change in MapViewScope
+        assert(initialRecompositionCount != currentRecompositionCount)
```
hud10837, line 217: We should use truth asserts, [it's in the toml](https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/blob/b18bc6080006d6a1e5d67e6709c97b7ed2c8661d/gradle/libs.versions.toml#L124) as part of the `unitTest` group, but maybe we should also add it to the composeTest group

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/callout/CalloutTests.kt
```diff
@@ -0,0 +1,262 @@
...
+        val calloutContainerNodeInteraction = composeTestRule
+            .onNodeWithContentDescription(calloutContainerLabel)
+
+        val calloutBitmap = calloutContainerNodeInteraction.captureToImage().asAndroidBitmap()
+        // test if the Callout contains a Red background Color
+        assert(calloutBitmap.containsColor(Color.Red.hashCode()))
+
+        val calloutContainerNode = calloutContainerNodeInteraction.fetchSemanticsNode()
+        // test Callout Size excluding leader height and border width
+        assert(calloutContainerNode.size == IntSize(442, 416))
```
hud10837, line 249: This line fails on my device with different width and height values. Is there any other way we can test this? Do we need to test the exact size?

hud10837 (dismissed):
Good job getting these working, I think overall the tests look good, a few changes to make as I'm still running into some failures

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/callout/CalloutTests.kt
shubham7109: From the time I spent testing this, I haven't found the ViewModel work outside of the setContent scope, can you help recommend a set of changes that would make ViewModel be able to work with the isolated Composable functions?

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/callout/CalloutTestActivity.kt
shubham7109: Yes, I wasn't able to determine if this Composable test function here is incorrect. Or if the ViewModel needs to be configured correctly. Could you help recommend a set of changes to get this to run smoothly?

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/callout/CalloutTests.kt
shubham7109: I see it too, I can't tell why the recompositions are happening as the `ResetViaStateChanges()` composable function would run when plugged in a regular activity. Could this be due to the ViewModel being passed in?

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/callout/CalloutTests.kt
shubham7109: I have configured this test to pass on our `Pixel 7` daily test devices. Using another device may change the DPI causing this value to be different. If you feel it isn't a necessary test, I may consider removing this check.

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/callout/CalloutTests.kt
shubham7109: I see it too, good catch! I think it would be better here if the `ResetViaStateChanges` reports the current recomposition count to the test case via an `onCalloutRecomposed` lambda callback. Rather than reading the value of a Text composable.

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/callout/CalloutTests.kt
hud10837: It's not an exact comparison but [here](https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/blob/b18bc6080006d6a1e5d67e6709c97b7ed2c8661d/toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/OAuthDefaultConfigurationTests.kt#L138) I create an `AuthenticatorState` and it's able to stay alive through the test, but in your case if it needs to be a viewmodel then I think the snippet I wrote should work? Quick tests indicated it did.

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/callout/CalloutTests.kt
hud10837: This still fails for me, and it also fails when I try to only hold the count in the test itself rather than in the composable function

```kotlin
androidx.compose.ui.test.junit4.android.ComposeNotIdleException: Idling resource timed out: possibly due to compose being busy.
IdlingResourceRegistry has the following idling resources registered:
- [busy] ComposeIdlingResource is busy due to pending recompositions.
- Note: Timeout on pending recomposition means that there are most likely infinite re-compositions happening in the tested code.
- Debug: hadRecomposerChanges = true, hadSnapshotChanges = false, hadAwaitersOnMainClock = false
```

shubham7109:
Thanks for the feedback @hud10837. I was able to circumvent the `Idling resource timed out` issue which was due to infinite recompositions of the `calloutRecompositionCount`, moving it to a helper function in the ViewModel helped resolve the issue. 

Also added the `truth` library to the `composeTest` group.

Ran a new vTest and looks to be working well now: https://runtime-kotlin.esri.com/view/vTest/job/vtest/job/toolkit/59/Test_20Report_20-_20Integration/

Ready for your review!

sorenoid (approved):
very nice

puneet-pdx (approved):
@shubham7109 looks good 👍🏽

PR #527, opened 2024-07-24
Callout: Fix to observe GeoElement changes 

shubham7109:
Related to issue: [#4343](https://devtopia.esri.com/runtime/kotlin/issues/4343)

### Description:
PR to add a remember key on the underlying GeoElement Callout parameter.   

### Summary of changes:

MapView Callout App:
- Updated the FeatureScreen to enable testing Callout to observe GeoElement changes.
- Removed viewpoint animation from FeatureScreen as it overshadows the Callout leaderLocation animation
- Isolated the clearing of the geoElement on its own function (rather than `clearTapLocationAndGeoElement` on every identified GeoElement), to avoid overshadowing Callout observing GeoElement changes
- GeoElement will now only be cleared on the close IconButton click.  
- Removed GraphicsScreen from microapp as it is not 200.5 priority. 

### Testing

Use the FeatureScreen Callout microapp with "Use tapped location" unchecked.  
- **Note**: The `identifyLayers` may take some time to identify a geoElement on tap, so the Callout may reposition a few seconds after tap.


microapps/MapViewCalloutApp/app/src/main/java/com/arcgismaps/toolkit/mapviewcalloutapp/screens/FeatureScreen.kt
```diff
@@ -100,14 +100,12 @@ fun FeatureScreen(viewModel: MapViewModel) {
             graphicsOverlays = remember { listOf(viewModel.tapLocationGraphicsOverlay) },
             onSingleTapConfirmed = { singleTapConfirmedEvent ->
                 viewModel.apply {
-                    // clears the tapped location and reset selected geoelement
-                    clearTapLocationAndGeoElement()
+                    // clears the tapped location and graphic
```
sorenoid, line 103: it seems like it is necessary to call `viewModel.clearSelectedGeoElement()` here. The old Callout is not getting dismissed.

shubham7109: The aim is to not dismiss the Callout between new GeoElements being identified. This  `onSingleTapConfirmed` now is able to update the content and leaderLocation of the Callout as GeoElement changes. 

Previously, the bug was unnoticed as we would dismiss the Callout between identify operations.

sorenoid (approved):
changes look good

eri9000 (approved):
LGTM 👍🏼

shubham7109:
Thanks @sorenoid and @eri9000  for all the help with testing this fix!

PR #537, opened 2024-08-07
drive publication of final toolkit build from FINAL_BUILD 

sorenoid:
environment variable

Related to issue: https://devtopia.esri.com/runtime/kotlin/issues/4399

### Description:

when FINAL_BUILD environment variable is set (to any value)
* publish the toolkit artifacts and the BOM with the no build number (e.g. 200.6.0)
* pick up the daily build of the SDK as a dependency (e.g. 200.6.0-4133, not 200.6.0) 

The build number from `buildnum.txt` is already used. It is passed in to the gradle command line from the jenkins environment in `ci/publish.sh`. The variables in gradle.properties just provide gradle properties to be passed at the command line.

### Summary of changes:

- check for `FINAL_BUILD` in the environment in the ci script `ci/publish.sh`. If present, pass in appropriate gradle properties to ignore build numbers and use the daily build for the sdk dependency.
- change the name of the `ignoreBuildNumber` gradle property to `finalBuild`.

### Pre-merge Checklist

testing: tested using `publishToMavenLocal` for both FINAL_BUILD and non FINAL_BUILD scenarios

  - [x] link: https://runtime-kotlin.esri.com/view/vTest/job/vtest/job/toolkit/57/
  - [ ] Yes
  - [ ] No

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -38,14 +38,13 @@ buildscript {
         // before any dependent subproject uses its symbols to configure a dokka task.
         classpath(libs.dokka.versioning)
     }
+    val finalBuild: Boolean = (project.properties["finalBuild"] ?: "false")
+        .run { this == "true" }
 
-    if (!project.hasProperty("finalBuild") || project.properties["finalBuild"] != "true") {
-        project.extra.set("finalBuild", "false")
-    } else {
-        project.logger.warn("release candidate build requested")
-    }
-
-    if (!project.hasProperty("versionNumber") || !project.hasProperty("buildNumber")) {
+    if (finalBuild) {
+        check(project.hasProperty("versionNumber"))
+        project.logger.info("release candidate build requested version ${project.properties["versionNumber"]}")
+    } else if (!project.hasProperty("versionNumber") && !project.hasProperty("buildNum")) {
         // both version number and build number must be set
         java.util.Properties().run {
             try {
@@ -70,8 +69,6 @@ buildscript {
                 project.extra.set("buildNumber", "SNAPSHOT")
             }
         }
-    } else {
-        project.logger.info("version and build number set from properties to ${project.properties["versionNumber"]}-${project.properties["buildNumber"]}")
     }
 }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -38,6 +38,41 @@ buildscript {
         // before any dependent subproject uses its symbols to configure a dokka task.
         classpath(libs.dokka.versioning)
     }
+
+    if (!project.hasProperty("finalBuild") || project.properties["finalBuild"] != "true") {
+        project.extra.set("finalBuild", "false")
+    } else {
+        project.logger.warn("release candidate build requested")
+    }
+
+    if (!project.hasProperty("versionNumber") || !project.hasProperty("buildNumber")) {
+        // both version number and build number must be set
+        java.util.Properties().run {
+            try {
+                File(project.projectDir, "../buildnum/buildnum.txt")
+                    .reader()
+                    .use {
+                        load(it)
+                    }
+                this["BUILDVER"]?.let {
+                    project.extra.set("versionNumber", it)
+                }
+                this["BUILDNUM"]?.let {
+                    project.extra.set("buildNumber", it)
+                }
+                check(project.hasProperty("versionNumber"))
+                check(project.hasProperty("buildNumber"))
+                project.logger.info("version and build number set from buildnum.txt to ${project.properties["versionNumber"]}-${project.properties["buildNumber"]}")
+            } catch (t: Throwable) {
+                // The buildnum file is not there. ignore it and log a warning.
+                project.logger.warn("the buildnum.txt file is missing or not readable")
+                project.extra.set("versionNumber", "0.0.0")
+                project.extra.set("buildNumber", "SNAPSHOT")
+            }
+        }
+    } else {
+        project.logger.info("version and build number set from properties to ${project.properties["versionNumber"]}-${project.properties["buildNumber"]}")
+    }
 }
 
 // Path to the centralized folder in root directory where test reports for connected tests end up
@@ -77,3 +112,4 @@ fun getModulesExcept(vararg modulesToExclude: String): List<String> =
             }
             .filter { !modulesToExclude.contains(it) } // exclude specified modules
     }
+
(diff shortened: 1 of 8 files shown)


gradle.properties
```diff
@@ -50,7 +50,7 @@ versionNumber=200.6.0
 buildNumber=0000-snapshot
 #set this flag to `true` to ignore the build number when publishing. This
 # will publish an artifact with a build number like "..:200.2.0" as opposed to "...:200.2.0-3963
```
gunt0001, line 52: Can we add some additional  doc here:
 ```suggestion
# set this flag to `true` to ignore the build number when publishing. This
# will publish an artifact with a build number like "..:200.2.0" as opposed to "...:200.2.0-3963
# Note this flag is overriden at the command line in Jenkins when the final build occurs. See
# _publishReleaseCandidate() in ci/publish.sh
```

File: gradle.properties [hunk 2]
```diff
@@ -50,7 +50,7 @@ versionNumber=200.6.0
...
-ignoreBuildNumber=false
+finalBuild=false
 # these versions define the dependency of the ArcGIS Maps SDK for Kotlin dependency
 # and are generally not overridden at the command line unless a special build is requested.
```
gunt0001, line 55:
```suggestion
# and are generally not overridden at the command line unless a special build is requested,
# such as the final build. See _publishReleaseCandidate() in ci/publish.sh
```

gradle.properties
```diff
@@ -50,7 +50,7 @@ versionNumber=200.6.0
 buildNumber=0000-snapshot
```
gunt0001, line 50: I think we should initialize `buildNumber` and `versionNumber` above in line 49 from `buildnum.txt` instead of hard coding it. This would only be relevant for local builds, since jenkins builds will override these via command line proeprties.

gradle.properties
gunt0001: I think we should initialize `sdkVersionNumber` and `sdkBuildNumber ` below in line 56/57 from `buildnum.txt` instead of hard coding it. This would only be relevant for local builds, since jenkins builds will override these via command line proeprties.

sorenoid: we had previously decided to pin the sdk's build and not use the daily build every day.

gradle.properties
sorenoid: the variable is removed entirely now. It is set either from the command line or procedurally in the top level gradle file.

build.gradle.kts
```diff
@@ -38,6 +38,41 @@ buildscript {
...
+                    project.extra.set("buildNumber", it)
+                }
+                check(project.hasProperty("versionNumber"))
+                check(project.hasProperty("buildNumber"))
+                project.logger.info("version and build number set from buildnum.txt to ${project.properties["versionNumber"]}-${project.properties["buildNumber"]}")
+            } catch (t: Throwable) {
+                // The buildnum file is not there. ignore it and log a warning.
+                project.logger.warn("the buildnum.txt file is missing or not readable")
+                project.extra.set("versionNumber", "0.0.0")
+                project.extra.set("buildNumber", "SNAPSHOT")
```
gunt0001, line 70: Good to have a fallback covered 👍

gunt0001 (approved):
@sorenoid - looks great! I tested the local build by running `./gradlew publish` to check that the version/build numbers get picked up from `buildnum.txt`. I did not test the command line case.

gradle.properties
```diff
@@ -44,13 +44,6 @@ artifactoryGroupId=com.esri
...
 artifactoryPassword=""
-# these numbers will define the artifact version on artifactory
-# and are overridden by the jenkins command line in the daily build
-versionNumber=200.6.0
-buildNumber=0000-snapshot
-#set this flag to `true` to ignore the build number when publishing. This
-# will publish an artifact with a build number like "..:200.2.0" as opposed to "...:200.2.0-3963
-ignoreBuildNumber=false
 # these versions define the dependency of the ArcGIS Maps SDK for Kotlin dependency
 # and are generally not overridden at the command line unless a special build is requested.
```
gunt0001, line 48: Could we still add some additional comment here, along the lines...
```suggestion
# and are generally not overridden at the command line unless a special build is requested.
# A final build before release is such a special build, in which case these SDK version numbers
# are overridden via command line, see _publishReleaseCandidate() in ci/publish.sh.
```

ci/publish.sh
```diff
@@ -67,9 +67,20 @@ function _check_options_and_set_variables() {
...
+    echo "error: Running the publish gradle task failed for the release candidate"
+    exit 1
+  fi
+}
+
 function _publish() {
   _log "Publish the release build to artifactory"
-  if ! ${apps_path}/arcgis-maps-sdk-kotlin-toolkit/ci/run_gradle_task.sh -v -t publish -x "-PartifactoryUsername=${ARTIFACTORY_USR} -PartifactoryPassword=${ARTIFACTORY_PSW} -PversionNumber=${BUILDVER} -PbuildNumber=${BUILDNUM}" ; then
+
+  if [ ! -z "${FINAL_BUILD}" ]; then
```
chri7325, line 81: I think this would be a little simple to always pass in the FINAL_BUILD env variable. It'll be set to false until our actual final build, where it'll be true. This should make the call a little simpler.

sorenoid: @chri7325 just to check: the env var will be the text `true`  or `false`, not `1` or `0` or something else?

chri7325: @sorenoid correct. We will keep it as a boolean false for our entire release and then switch it for the final day https://devtopia.esri.com/runtime/devops/blob/main/sharedlibs/vars/initializeJavaSE.groovy#L34. I'll move this out of the java initialize to the global one.

sorenoid: @chri7325 I was able to simplify the bash part. It caused some changes in the gradle parts, but overall it is an improvement to gradle.

gradle.properties
gunt0001: @sorenoid - I updated my doc suggestion based on your latest changes.

gradle.properties
```diff
@@ -44,14 +44,9 @@ artifactoryGroupId=com.esri
...
+# A final build before release is such a special build, in which case these SDK version numbers
+# are overridden via command line, see _publishReleaseCandidate() in ci/publish.sh.
```
gunt0001, line 50: @sorenoid - this doc needs updating now based on your atest changes:
```suggestion
# A final build before release is such a special build, in which case these SDK version numbers
# are overridden via command line, see sdkVersionNumber in settings.gradle.kts.
```

chri7325 (approved):
Nice!

PR #617, opened 2024-10-21
changes to disable the trace button while a trace is in progress

puneet-pdx:
Related to issue: # https://devtopia.esri.com/runtime/kotlin/issues/4759

### Description:

The user stays on the trace option screen for way too long after a trace has been kicked off and can press the trace button multiple times leading to unexpected behaviors. We need to stop that from happening

### Summary of changes:

- Don't zoom in as a part of trace. Perform zooming when trace is done
- Disable the trace button as soon as Trace() is called and reset it when trace is done
- show progress indicator when trace is in progress and also when identify for starting points is in progress.

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/147/
  - [ ] Yes
  - [ ] No


toolkit/utilitynetworks/src/main/java/com/arcgismaps/toolkit/utilitynetworks/TraceState.kt
```diff
@@ -503,9 +510,14 @@ public class TraceState(
...
                     // If the size of the starting points has changed, then the starting point was added
                     _addStartingPointMode.value = AddStartingPointMode.Stopped
+                    _isIdentifyInProcess.value = false
                     showScreen(TraceNavRoute.TraceOptions)
                 }
             }
+            _isIdentifyInProcess.value = false
+        }
+        result.onFailure {
+            _isIdentifyInProcess.value = false
```
sorenoid, line 520: can this just be done once at the bottom of the function for success and failure?

sorenoid (approved):
LGTM. just one little suggested code cleanup

toolkit/utilitynetworks/src/main/java/com/arcgismaps/toolkit/utilitynetworks/TraceState.kt
puneet-pdx: yes, updated

toolkit/utilitynetworks/src/main/java/com/arcgismaps/toolkit/utilitynetworks/TraceState.kt
```diff
@@ -282,22 +288,29 @@ public class TraceState(
      * @since 200.6.0
      */
     internal suspend fun trace(): Result<Unit> = runCatchingCancellable {
+        _isTraceInProgress.value = true
         // Run a trace
         val traceConfiguration = selectedTraceConfiguration.value
-            ?: throw TraceToolException(TraceError.NO_TRACE_CONFIGURATIONS_FOUND)
```
eri9000, line 287: It seems that we need to set `_isTraceInProgress ` to `true`  at the beginning and set it to `false` at the end, regardless if its a failure or not.
I suggest wrapping this in a `try-finally` with the `finally` being `_isTraceInProgress.value = false`. 
Better yet, create a function that can be re-used in `identifyFeatures()` and only keep track of one progress mutable state

```
private inline fun <T> withProgressIndicator(block: () -> T): T {
    _taskInProgress.value = true
    return try {
        block()
    } finally {
        _taskInProgress.value = false
    }
}
``` 

and you can use it in the function as such: 
```
internal suspend fun trace(): Result<Unit> = runCatchingCancellable {
    withProgressIndicator {
....
```

toolkit/utilitynetworks/src/main/java/com/arcgismaps/toolkit/utilitynetworks/TraceState.kt
```diff
@@ -483,6 +488,7 @@ public class TraceState(
     }
 
     private suspend fun identifyFeatures(mapPoint: Point, screenCoordinate: ScreenCoordinate) {
+        _isIdentifyInProcess.value = true
```
eri9000, line 491: this could be replaces with the suggestion above

eri9000 (commented):
@puneet-pdx Good work,
See my comments.

toolkit/utilitynetworks/src/main/java/com/arcgismaps/toolkit/utilitynetworks/TraceState.kt
```diff
@@ -282,90 +285,98 @@ public class TraceState(
...
+            val utilityTraceParameters =
+                UtilityTraceParameters(
+                    traceConfiguration,
+                    currentTraceStartingPoints.map { it.utilityElement })
 
-        val traceResults = utilityNetwork.trace(utilityTraceParameters).getOrElse {
-            throw it
-        }
+            val traceResults = utilityNetwork.trace(utilityTraceParameters).getOrElse {
+                _isTaskInProgress.value = false
```
eri9000, line 307: i don't think this will be needed since the `withProgressIndicator` will take care of setting it to false

eri9000 (approved):
@puneet-pdx 
Looks good,
just one more comment otherwise LGTM

PR #618, opened 2024-10-22
Compass micro app improvement

gunt0001:
### Description:

Currently the Compass `onClick` event is handled as a trailing lambda in the Compass micro app. It has been pointed out that this makes the code harder to read, since one could assume that this is a "content lambda" instead of a lambda that handles the click event. Ideally, the `onClick` parameter in the `Compass()` function would not be the last parameter, so it cannot be used with trailing lambda, but changing that would break the API. This change makes the micro app code more readable by using a named parameter for the `onClick` lambda.

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/148/
  - [ ] Yes
  - [ ] No


colinanderson (approved):
@gunt0001 looks ok

kaushikrw (approved):
👍

PR #628, opened 2024-10-26
Update Result screen UI when the First element in the list is getting deleted

puneet-pdx:
Related to issue: # https://devtopia.esri.com/runtime/kotlin/issues/4844

### Description:

The Result screen UI does not update when deleting the First result

### Summary of changes:

- We are remembering the selectedTraceRunIndex in the result screen, the value of which does not change when we delete the 0th element cause the UI to not refresh
- Also we were not highlighting the graphic of the next element that becomes the 0th element when the previous element is deleted.

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/162/
  - [ ] Yes
  - [ ] No

### DIFF
--- a/toolkit/utilitynetworks/src/main/java/com/arcgismaps/toolkit/utilitynetworks/ui/TraceResultScreen.kt
+++ b/toolkit/utilitynetworks/src/main/java/com/arcgismaps/toolkit/utilitynetworks/ui/TraceResultScreen.kt
@@ -79,50 +79,58 @@ internal fun TraceResultScreen(
         onBackToNewTrace()
     }
 
-    Surface(modifier = Modifier.fillMaxSize()) {
-        Column {
+    if (traceResults.isNotEmpty()) {
+        Surface(modifier = Modifier.fillMaxSize()) {
+            Column {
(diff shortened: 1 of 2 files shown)


toolkit/utilitynetworks/src/main/java/com/arcgismaps/toolkit/utilitynetworks/ui/TraceResultScreen.kt
```diff
@@ -87,7 +87,11 @@ internal fun TraceResultScreen(
             .fillMaxSize()
             .padding(horizontal = 10.dp)) {
 
-            val selectedTraceRun = remember(selectedTraceRunIndex) { traceResults[selectedTraceRunIndex] }
+            if (traceResults.isEmpty()) {
+                return@Column
+            }
```
sorenoid, line 92: as discussed, better to avoid an early return and specify the composable content for both an if and an else case. approving to keep things moving!

eri9000 (approved):

PR #678, opened 2024-11-22
Floor filter state initialization

puneet-pdx:
Related to issue: # https://devtopia.esri.com/runtime/kotlin/issues/3129

### Description:

- Provide a way to help the user figure out if the FloorFilterState is successfully loaded and is ready to select sites, facilities or levels

### Summary of changes:

- Add initializationStatus `State` that reports on the initialization status of the FloorFilterState. 
- Add InitializationStatus enum
- Deprecate the existing FloorFilterState() fun that takes coroutineScope, replace it with a new version that does not take the coroutineScope
- Remove init() from FloorFilterState and replace it with `public suspend fun initialize(): Result<Unit>`
- add `setSelectedValuesAppliedBeforeInitialization()` which sets the selected values that were applied before initialization by the user
- Refactor FloorFilter to use initializationStatus and to auto initialize FloorFilterState 
- Add `FloorFilterError` to localize errors defined in the component
- Add dependency on material icons to use the error icon 

The design does not specify how to report the errors and we had followed the previous implementation done by the indoors team and swift that did not report errors via UI. I have made changes so that the FloorFilter button will reflect when the FloorFilterState is initializing and when it fails to load.

https://github.com/user-attachments/assets/7e2f79a7-287f-4503-885b-03872eb32fab

floorFilterError

### Pre-merge Checklist

  - [ ] link: ~https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/216/~
  - [x] https://runtime-kotlin.esri.com/view/vTest/job/vtest/job/toolkit/240/
  - [ ] Yes
  - [ ] No


puneet-pdx:
For testing you can use these snippet in the test app
```Kotlin
val floorAwareWebMap by remember {
        mutableStateOf(
            ArcGISMap(
                PortalItem(
                    Portal("https://arcgis.com/"),
//                    "b4b599a43a474d3394" // invalid item id
//                    "471eb0bf37074b1fbb972b1da70fb310" // valid item id, no floor-aware data
                    "b4b599a43a474d33946cf0df526426f5" // valid item id, with floor-aware data
                )
            )
        )
    }
```

for auto setting of the selected values
```Kotlin
      floorFilterState.selectedLevelId = "GFC.WEST.G1"
//    floorFilterState.selectedSiteId = "GFC.RA"
//    floorFilterState.selectedFacilityId = "GCF.EAST.A"
//    floorFilterState.selectedFacilityId = "GFC.RA.L"
```
by using initializationStatus
```Kotlin
    val initializationStatus by floorFilterState.initializationStatus
    when (initializationStatus) {
        is InitializationStatus.FailedToInitialize -> {

        }
        InitializationStatus.Initialized -> {
            floorFilterState.selectedLevelId = "GFC.WEST.G1"
        }
        InitializationStatus.Initializing -> {

        }
        InitializationStatus.NotInitialized -> {

        }
    }
```

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/FloorFilterState.kt
```diff
@@ -212,30 +226,59 @@ private class FloorFilterStateImpl(
...
+        if (_selectedSiteId != null) {
+            selectedSiteId = _selectedSiteId
+        }
+        if (_selectedFacilityId != null) {
+            selectedFacilityId = _selectedFacilityId
+        }
+        if (_selectedLevelId != null) {
+            val temp = _selectedLevelId
+            _selectedLevelId = null
+            selectedLevelId = temp
```
shubham7109, line 281: I'm not sure if I understand why the internal `_selectedLevelId` is being diverted to `null` here. 
```suggestion
            selectedLevelId = _selectedLevelId
```

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/FloorFilter.kt
```diff
@@ -127,25 +140,103 @@ public fun FloorFilter(
...
+                            .width(uiProperties.buttonSize.width.dp),
+                        contentAlignment = Center
+                    ) {
+                        CircularProgressIndicator(modifier = Modifier.size(uiProperties.buttonSize.height.dp * 0.5f))
+                    }
+                }
+
+                is InitializationStatus.FailedToInitialize -> {
+                    val errorMessage = (initializationStatus as InitializationStatus.FailedToInitialize).error.getErrorMessage(context)
+                    Log.e("FloorFilter", errorMessage)
```
shubham7109, line 157:
```suggestion
                    Log.e("FloorFilterException", errorMessage)
```

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/FloorFilter.kt
```diff
@@ -127,25 +140,103 @@ public fun FloorFilter(
...
+                        contentAlignment = Center
+                    ) {
+                        Icon(
+                            imageVector = Icons.Outlined.ErrorOutline,
+                            contentDescription = stringResource(id = R.string.geomodel_has_no_floor_aware_data),
+                            modifier = Modifier.clickable {
+                                Toast.makeText(context, errorMessage, Toast.LENGTH_SHORT).show()
+                            },
+                            tint = MaterialTheme.colorScheme.primary
+                        )
```
shubham7109, line 171: This should either be an `IconButton` or the `Modifier.clickable` should be part of the parent `Box`.

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/FloorFilter.kt
```diff
@@ -127,25 +140,103 @@ public fun FloorFilter(
...
+                            .width(uiProperties.buttonSize.width.dp),
+                        contentAlignment = Center
+                    ) {
+                        Icon(
+                            imageVector = Icons.Outlined.ErrorOutline,
+                            contentDescription = stringResource(id = R.string.geomodel_has_no_floor_aware_data),
+                            modifier = Modifier.clickable {
+                                Toast.makeText(context, errorMessage, Toast.LENGTH_SHORT).show()
+                            },
+                            tint = MaterialTheme.colorScheme.primary
```
shubham7109, line 170: Maybe we should use the error scheme here? 
```suggestion
                            tint = MaterialTheme.colorScheme.error
```

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/FloorFilter.kt
```diff
@@ -127,25 +140,103 @@ public fun FloorFilter(
...
+                isFacilitiesSelectorVisible,
+                uiProperties,
+                onSiteFacilitySelectorVisibilityChanged = { isVisible ->
+                    isSiteAndFacilitySelectorVisible = isVisible
+                },
+                onFacilitiesSelectorVisible = { isVisible ->
+                    isFacilitiesSelectorVisible = isVisible
+                }
+            )
+            return
```
shubham7109, line 223: This was previously `return@Surface`
```suggestion
            return@Column
```

shubham7109 (changes requested):
@puneet-pdx If I were to select a `floorFilterState.selectedLevelId = "GFC.WEST.GB"` on initialize, the floor filter is able to display the floor-map and the levels list. But then if were to open the `SiteFacilityButton` it displays the window with the correct site selected, but it does not show the facility as selected. 


Tested by setting the site/facility/level on initialize and the floor filter can display info on initialize. I added some change requests below:

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/FloorFilter.kt
puneet-pdx: The error is coming from FloorFilter so the TAG should say so

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/FloorFilterState.kt
puneet-pdx: We have a check were we don't set the selectedLevelId to a new value if it is the same as the old one. This logic caches the old value, sets the backing property to null and resets the selectedLevelId for setting the same value again

toolkit/indoors/src/main/java/com/arcgismaps/toolkit/indoors/FloorFilterState.kt
```diff
@@ -60,6 +63,7 @@ import kotlinx.coroutines.launch
  */
 public sealed interface FloorFilterState {
     public val floorManager: StateFlow<FloorManager?>
+    public val initializationStatus: State<InitializationStatus>
```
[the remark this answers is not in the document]
puneet-pdx: I remember having a discussion about not wanting to use SDK types in the toolkit. Also I think initialization makes more sense than load for FloorFilterState. 
Regarding this being a State, I think we are moving more towards Android compose paradigms in the Toolkit. StateFlow in the Kotlin API makes sense.
@sorenoid thoughts ?

puneet-pdx:
>@puneet-pdx If I were to select a floorFilterState.selectedLevelId = "GFC.WEST.GB" on initialize, the floor filter is able to display the floor-map and the levels list. But then if were to open the SiteFacilityButton it displays the window with the correct site selected, but it does not show the facility as selected.

@shubham7109 this is something that is not supported. As if were to do it, it would break the workflow of the UI based component which goes with the linear flow of select a site, then select a facility and then select a level.
Selecting a site resets the previous facility and level that was selected, then the user will select the facility and the level they would want to traverse. So at all times selectedSite, selectedFacility and selectedLevel return the correct values.

Here we are giving the user an option to bypass the usual flow to traverse the facility and to be able to select the starting level/facility/site on the map, the UI updating is not a part of it. 
https://devtopia.esri.com/runtime/common-toolkit/blob/main/designs/FloorFilter/FloorFilter.md
If we were to take the logic out to not reset the facility when site is updated then the Selected facility and site will return stale values, I don't think that will that be acceptable.

shubham7109 (approved):
@puneet-pdx Thanks for the explaination and the initialization fix works well as you have described. Looks good to me 👍

sorenoid:
@puneet-pdx can you please merge in v.next and address lint errors in this vTest? https://runtime-kotlin.esri.com/view/vTest/job/vtest/job/toolkit/236/parsed_console/

puneet-pdx:
@sorenoid updated.
https://runtime-kotlin.esri.com/view/vTest/job/vtest/job/toolkit/240/

sorenoid (approved):

PR #695, opened 2024-12-19
Hud10837/improve tabletop positioning

hud10837:
Related to issue: https://devtopia.esri.com/runtime/kotlin/issues/4764

### Description:

Fixes an error where tapping on the screen to place the tabletop scene would result in a scene that was offset vertically from where we expected it to be. This was caused by an error processing the motion event for the tap. The hit test would be performed using absolute coordinates on the screen, but the scene would be placed according to relative coordinates to the SceneView.

Also cleans up a few key areas to ensure placement is always more accurate.

Note: I was able to verify that when the viewport size of the GLSurfaceView is set, it is correctly smaller than the screen size

### Summary of changes:

- Use `pointerInput` rather than `pointerInteropFilter` to receive touch events on the background surface view
- Upgrade ARCore
- Simplify the translation of ARCore hit poses to Esri `TransformationMatrix` by using `extractTranslation` rather than hardcoding quaternion values when setting the anchor position
- Fix the values used to calculate device orientation

### Pre-merge Checklist

  - [ ] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/260/
  - [ ] Yes
  - [x] No


toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/ArCameraFeed.kt
```diff
@@ -88,9 +89,10 @@ internal fun ArCameraFeed(
 
     AndroidView(factory = { surfaceViewWrapper.glSurfaceView }, modifier = Modifier
         .fillMaxSize()
-        .pointerInteropFilter {
-            cameraFeedRenderer.onClick(it)
-            true
+        .pointerInput(Unit) {
+            detectTapGestures { offset ->
+                cameraFeedRenderer.onClick(offset)
+            }
```
hud10837, line 95: It looks like `MotionEvent.getX()` and `MotionEvent.getY()` can sometimes return relative coordinates, and sometimes return absolute coordinates, depending on the state of the app. More info [here](https://stackoverflow.com/questions/38061734/how-do-i-make-the-coordinates-of-motionevent-match-the-ones-of-a-scaled-canvas) and [here](http://andmobidev.blogspot.com/2010/01/getting-relative-coordinates-from.html)

`detectTapGesture` however is guaranteed to return the coordinates relative to the view it's applied on, see [here](https://developer.android.com/reference/kotlin/androidx/compose/ui/input/pointer/PointerInputScope#(androidx.compose.ui.input.pointer.PointerInputScope).detectTapGestures(kotlin.Function1,kotlin.Function1,kotlin.coroutines.SuspendFunction2,kotlin.Function1)).

Also, we don't actually need a motion event so probably for the best to use the compose solution

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/TableTopSceneView.kt
```diff
@@ -264,10 +254,10 @@ public fun TableTopSceneView(
                                 imageIntrinsics.imageDimensions[0].toFloat(),
                                 imageIntrinsics.imageDimensions[1].toFloat(),
                                 deviceOrientation = when (displayRotation) {
-                                    0 -> DeviceOrientation.Portrait
```
01smito01, line 267: I never saw issues with rotation on older versions of the ARTT app, but this seems to work just fine!

01smito01 (approved):
LGTM. Tested on my phone and the scene placement is defo better than before.

hud10837:
Thanks @alan-edi @01smito01 kicked off another [vtest](https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/269/) to be sure and will merge when it completes

PR #708, opened 2025-01-17
Line ScaleBar: Tie UI and Backend together

puneet-pdx:
Related to issue: # https://devtopia.esri.com/runtime/kotlin/issues/5189

### Description:

Changes to make the Line Scalebar work with new display length and label when the viewpoint changes

### Summary of changes:

- Add scalebarViewModelFactory 
- Scalebar - add logic to initialize the ViewModel and update the scalebar whenever the unitsPerDip, viewpoint and spatialreference change
- Add logic to calculate minimum segment width, add logic to call that in a composable from the viewmodel
- Update test

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/297/
  - [ ] Yes
  - [ ] No


toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
```diff
@@ -26,6 +26,11 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
```
eri9000, line 29: some imports need cleanup

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/internal/ScalebarViewModel.kt
```diff
@@ -132,11 +129,11 @@ internal class ScalebarViewModel(
      *
      * @since 200.7.0
      */
-    fun updateScaleBar(
+    internal fun updateScaleBar(
```
eri9000, line 132: This method does not directly update the scalebar
```suggestion
    internal fun computeScalebarProperties(
```

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
```diff
@@ -85,13 +95,70 @@ public fun Scalebar(
...
+    }
+
+    val isUpdateLabels by scalebarViewModel.isUpdateLabels
+    if (isUpdateLabels) {
+        scalebarViewModel.updateLabels(
+            minSegmentWidth(scalebarViewModel.lineMapLength, labelTypography)
+        )
+    }
+
+    val isScaleBarUpdated by scalebarViewModel.isScaleBarUpdated
```
eri9000, line 126: we could avoid having these flags by returning a list of labels from the method `updateScaleBar`, i would suggest calling updateLabels inside `updateScalebar` and returning this list of labels.

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
```diff
@@ -85,13 +95,70 @@ public fun Scalebar(
...
+    if (isScaleBarUpdated) {
+        val density = LocalDensity.current
+        ShowScalebar(
+            scalebarViewModel.displayLength * density.density,
+            scalebarViewModel.labels,
```
eri9000, line 131: if you get a list of scalebar labels, then you can pass the list to this composable, which will cause a recomposition when the labels change.

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
```diff
@@ -85,13 +95,70 @@ public fun Scalebar(
...
+            style,
+            colorScheme,
+            shapes,
+            modifier
+        )
+    }
+}
+
+@Composable
+private fun ShowScalebar(
```
eri9000, line 141: i don't think this composable is needed as the code could go directly inside `Scalebar`

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/internal/ScalebarViewModel.kt
```diff
@@ -64,16 +69,8 @@ internal class ScalebarViewModel(
      *
      * @since 200.7.0
      */
-    private fun updateLabels() {
+    internal fun updateLabels(minSegmentWidth: Double) {
         val localLabels = mutableListOf<ScalebarLabel>()
-        val minSegmentTestString: String = if (lineMapLength >= 100) {
```
eri9000, line 69: instead of creating a separate `composable` function for this block, can you pass in the `textMeasurer` to this function?
this way you can still refactor this block of code into a function but you can keep it here, with the rest of the business logic

eri9000 (commented):
Great work Puneet.
some suggestions

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
```diff
@@ -85,13 +95,70 @@ public fun Scalebar(
...
+        scalebarViewModel.updateLabels(
+            minSegmentWidth(scalebarViewModel.lineMapLength, labelTypography)
+        )
+    }
+
+    val isScaleBarUpdated by scalebarViewModel.isScaleBarUpdated
+    if (isScaleBarUpdated) {
+        val density = LocalDensity.current
+        ShowScalebar(
+            scalebarViewModel.displayLength * density.density,
```
eri9000, line 130: this could be extracted into an extension function `toPx(dentisy:Density)`. That way we know what the multiplication is doing

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
```diff
@@ -109,7 +176,7 @@ internal fun ScalebarPreview() {
  * Displays a scalebar with single label and endpoint lines.
  *
  * @param modifier The modifier to apply to the layout.
- * @param scaleValue The scale value to display.
+ * @param label The scale value to display.
  * @param maxWidth The width of the scale bar.
```
eri9000, line 180:
```suggestion
 * @param maxWidth The width of the scale bar in pixels.
```

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
```diff
@@ -226,3 +294,31 @@ private fun availableLineDisplayLength(
         }
     }
 }
+
+/**
+ * Returns the minimum segment width required to display the labels without overlapping.
```
eri9000, line 299: is this returning the values in `dp` or `pixels`?

eri9000 (commented):
some followup comments

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
puneet-pdx: updated

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
puneet-pdx: pixels

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
puneet-pdx: we cannot call `updateLabels` from `updateScalebar` as we have to measure the label's maxUnitDisplayWidth before we can call `updateLabels`

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
puneet-pdx: This call will have to be in an if block as we don't want to try to draw the scaleBar if there is no list of labels or if we don't know the displayLength or the displayUnits.
Relying on just the list of labels here gives false impression that we will recompose if the labels list is updated. This is the reason why I put the flag so in the code it reads that we will recompose the scalebar when that flag is set and that there are conditions that need to be met for it to be set.

puneet-pdx: This call will have to be in an if block as we don't want to try to draw the scaleBar if there is no list of labels or if we don't know the displayLength or the displayUnits.
Relying on just the list of labels here gives false impression that we will recompose if the labels list is updated. This is the reason why I put the flag so in the code it reads that we will recompose the scalebar when that flag is set and that there are conditions that need to be met for it to be set.

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
eri9000: Regarding then name, it is common to use noun phrases rather than verbs as composables declare UI not perform actions. I suggest 
`ScalebarView` or simply `Scalebar`

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
eri9000: could you not just pass the result of `minSegmentWidth` to `updateScaleBar` and that way you could just call `UpdateLabels` inside there?

eri9000 (approved):
👍🏼

PR #709, opened 2025-01-20
Hud10837/implement world tracking

hud10837:
Related to issue: #

### Summary of changes:

- 

### Pre-merge Checklist

  - [ ] link:
  - [ ] Yes
  - [ ] No


microapps/ArWorldScaleApp/app/src/main/java/com/arcgismaps/toolkit/arworldscaleapp/screens/MainScreen.kt
```diff
@@ -39,7 +39,6 @@ fun MainScreen() {
     }
     WorldScaleSceneView(
         arcGISScene = arcGISScene,
-        locationDataSource = SystemLocationDataSource(),
```
hud10837, line 42: This will no longer be passed in by the user

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -146,16 +158,15 @@ public fun WorldScaleSceneView(
...
-
     val cameraController = remember { TransformationMatrixCameraController() }
 
-    val coroutineScope = rememberCoroutineScope()
-
-    LaunchedEffect(locationDataSource) {
-        launch {
-            locationDataSource.status.collect {
+    var isLocationDataSourceStarted by remember { mutableStateOf(false) }
+    if (hasCompletedPermissionRequest && context.checkPermissionsGranted()) {
```
hud10837, line 164: `context.checkPermissionsGranted()` will re-check the permissions with the system every time it is called, rather than storing that in a mutable state. I found this was much more predictable and easy to understand than using state, and it should only run once when `hasCompletedPermissionRequest` is true

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -172,65 +183,17 @@ public fun WorldScaleSceneView(
...
-            coroutineScope.launch {
-                locationDataSource.stop()
-            }
-        }
+        )
     }
 
     Box(modifier = modifier) {
-        if (cameraPermissionGranted && arCoreInstalled) {
+        if (isLocationDataSourceStarted && arCoreInstalled) {
```
hud10837, line 196: `isLocationDataSourceStarted` will only be true if permissions have already been granted

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -310,3 +278,206 @@ public fun WorldScaleSceneView(
...
+        }
+
+    if (!hasLaunchedRequest) {
+        SideEffect {
+            hasLaunchedRequest = true
+            launcher.launch(permissionsToRequest.toTypedArray())
+        }
+    }
+
+    return hasCompletedRequest
```
hud10837, line 374: Note that `hasLaunchedRequest` is used internally to this function to ensure the request is only launched once. `hasCompletedRequest` is used for the WorldScaleSceneView to know it's okay to continue with the startup sequence

01smito01 (approved):
Nothing obviously wrong to me, only a few queries & suggestions

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -310,3 +278,210 @@ public fun WorldScaleSceneView(
...
+ * @return A [MutableState] that will be true when the permissions have been requested and the result has been received.
+ * @since 200.7.0
+ */
+@Composable
+private fun requestPermissionsOrFail(
+    context: Context,
+    requestCameraPermissionAutomatically: Boolean,
+    requestLocationPermissionAutomatically: Boolean,
+    initializationStatus: MutableState<WorldScaleSceneViewStatus>,
+    onInitializationStatusChanged: ((WorldScaleSceneViewStatus) -> Unit)?
```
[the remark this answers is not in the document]
hud10837: These are both meant to be set to the callback parameter that's set on the WorldScaleSceneView, and the convention we've used there for the GeoViews and TableTopSceneView is to make them nullable. IIRC this was because by making them nullable we can easily check if they are set to something meaningful, and if not, we can avoid doing calculation (like listening to the state flows of the underlying geoview). I'm not sure how big an improvement that was but given the convention it makes sense to keep it going. In the case of these utility functions, I think it's easier to read the calling code if it just passes through the value from the WSSV argument, rather than having to do a null check inline

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -310,3 +278,210 @@ public fun WorldScaleSceneView(
...
+): MutableState<Boolean> {
+    var hasLaunchedRequest by remember { mutableStateOf(false) }
+    val hasCompletedRequest = remember { mutableStateOf(false) }
```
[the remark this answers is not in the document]
hud10837: It seems to work as a delegated boolean, which looks cleaner, so I've changed that

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -310,3 +278,210 @@ public fun WorldScaleSceneView(
...
+ */
+@Composable
+private fun requestPermissionsOrFail(
+    context: Context,
+    requestCameraPermissionAutomatically: Boolean,
+    requestLocationPermissionAutomatically: Boolean,
+    initializationStatus: MutableState<WorldScaleSceneViewStatus>,
+    onInitializationStatusChanged: ((WorldScaleSceneViewStatus) -> Unit)?
+): Boolean {
+    var hasLaunchedRequest by remember { mutableStateOf(false) }
```
gunt0001, line 298: Does this variable need to be a state object? I don't think we use it to trigger recomposition.
```suggestion
    var hasLaunchedRequest = remember { false }
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -310,3 +278,210 @@ public fun WorldScaleSceneView(
...
+/**
+ * Checks the permissions required for the [WorldScaleSceneView] to function and requests them if necessary.
+ * If [requestCameraPermissionAutomatically] or [requestLocationPermissionAutomatically] are set to false,
+ * the permissions will not be requested automatically and the initialization will fail if the permissions are not granted.
+ *
+ * @return A [MutableState] that will be true when the permissions have been requested and the result has been received.
+ * @since 200.7.0
+ */
+@Composable
+private fun requestPermissionsOrFail(
```
gunt0001, line 291: Since this function returns a remembered state variable, I think we should prefix it with "remember", based on [convention](https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-api-guidelines.md#naming-composable-functions-that-remember-the-objects-they-return). How about:
```suggestion
private fun rememberPermissionsGranted(
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -57,31 +70,33 @@ import com.arcgismaps.mapping.view.SceneViewInteractionOptions
...
+import kotlin.time.Duration.Companion.seconds
 
 @Composable
 public fun WorldScaleSceneView(
     arcGISScene: ArcGISScene,
-    locationDataSource: LocationDataSource,
     modifier: Modifier = Modifier,
     onInitializationStatusChanged: ((WorldScaleSceneViewStatus) -> Unit)? = null,
     requestCameraPermissionAutomatically: Boolean = true,
+    requestLocationPermissionAutomatically: Boolean = true,
```
gunt0001, line 99: I wonder if we should have just a single `requestPermissionsAutomatically` parameter? What's the use case for having separate ones? We can solve this separate from this PR though.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
gunt0001: When `hasCompletedPermissionRequest` is true, doesn't this mean that all required permissions have been granted? Why do we need to then check again with `checkPermissionsGranted()`?

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/LocationDataSourceWrapper.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlinx.coroutines.launch
+
+/**
+ * Wrapper class to manage the lifecycle of a [LocationDataSource].
+ * The [LocationDataSource] will be started when the lifecycle owner is resumed and stopped when the lifecycle owner is paused or destroyed.
+ *
+ * @since 200.7.0
+ */
+internal class LocationDataSourceWrapper(val locationDataSource: LocationDataSource) : DefaultLifecycleObserver {
+    // This ensures that we can stop the location data source on a scope that won't be cancelled when the composition is destroyed
```
gunt0001, line 35: I don't quite understand this comment, lets discuss.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/LocationDataSourceWrapper.kt
```diff
@@ -0,0 +1,55 @@
...
+    private val scope: CoroutineScope = CoroutineScope(Dispatchers.Default)
+
+    override fun onDestroy(owner: LifecycleOwner) {
+        scope.launch {
+            locationDataSource.stop()
+        }
```
gunt0001, line 41: When the wrapper is destroyed, we should just cancel the scope:
```suggestion
        scope.cancel()
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/LocationDataSourceWrapper.kt
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    override fun onPause(owner: LifecycleOwner) {
+        scope.launch {
+            locationDataSource.stop()
+        }
+    }
+
+    fun startLocationDataSource() {
```
gunt0001, line 50: Could do this instead in `onResume`?
```suggestion
    override fun onResume(owner: LifecycleOwner) {
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
hud10837: I changed this to use `LaunchedEffect(Unit)` to launch the request. We can discuss if you think `LaunchedEffect` is appropriate here, but it permits us to use a key even though we are not suspending. The key means we can pass `Unit` and the launch only happens once, which is the behavior this boolean was meant to enable anyhow.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
hud10837: In the version you reviewed, `hasCompletedPermissionRequest` could be true even if all permissions were not granted. However given your question I thought about it more and reworked it so that the mutable state encodes whether all permissions were granted.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/LocationDataSourceWrapper.kt
hud10837: I put this inside the launch, since otherwise I think we defeat the purpose of the wrapper. Let's discuss.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -310,3 +270,161 @@ public fun WorldScaleSceneView(
         }
     }
 }
+
+/**
+ * Checks the permissions required for the [WorldScaleSceneView] to function and requests them if necessary.
+ * If [requestCameraPermissionAutomatically] or [requestLocationPermissionAutomatically] are set to false,
```
gunt0001, line 276: This doc is out of date

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -310,3 +270,161 @@ public fun WorldScaleSceneView(
...
+    onInitializationStatusChanged: ((WorldScaleSceneViewStatus) -> Unit)?
+): State<Boolean> {
+    val allPermissionsGranted = remember { mutableStateOf(false) }
+    val permissionsToRequest = mutableListOf<String>()
+    val launcher =
+        rememberLauncherForActivityResult(
+            ActivityResultContracts.RequestMultiplePermissions()
+        ) { grantedState: Map<String, Boolean> ->
+            // This callback will only be relevant to the permissions that were added to the request.
+            // If the user has set [requestCameraPermissionAutomatically] to false, for example,
```
gunt0001, line 295: This comment is out of date

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -310,3 +270,161 @@ public fun WorldScaleSceneView(
...
+ * @since 200.7.0
+ */
+@Composable
+private fun rememberPermissionsGranted(
+    context: Context,
+    initializationStatus: MutableState<WorldScaleSceneViewStatus>,
+    onInitializationStatusChanged: ((WorldScaleSceneViewStatus) -> Unit)?
+): State<Boolean> {
+    val allPermissionsGranted = remember { mutableStateOf(false) }
+    val permissionsToRequest = mutableListOf<String>()
```
gunt0001, line 289: Where does this list get populated?

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -310,3 +270,161 @@ public fun WorldScaleSceneView(
...
+}
+
+/**
+ * Starts the [locationDataSource] and periodically updates the [cameraController] with new locations
+ *
+ * @since 200.7.0
+ */
+@Composable
+private fun LocationTracker(
+    locationDataSource: LocationDataSource,
```
gunt0001, line 330: I think we can remove this parameter and keep the LDS as an internal remembered variable of this function. The LDS is not used y the client code.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/LocationDataSourceWrapper.kt
```diff
@@ -0,0 +1,60 @@
...
+import kotlinx.coroutines.launch
+
+/**
+ * Wrapper class to manage the lifecycle of a [LocationDataSource].
+ * The [LocationDataSource] will be started when the lifecycle owner is resumed and stopped when the lifecycle owner is paused or destroyed.
+ *
+ * @since 200.7.0
+ */
+internal class LocationDataSourceWrapper(val locationDataSource: LocationDataSource) : DefaultLifecycleObserver {
+    // This ensures that we can stop the location data source on a scope that won't be cancelled when the composition is destroyed
```
gunt0001, line 36:
```suggestion
    // coroutine scope that is tied to the lifecycle of this LocationDataSourceWrapper
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -310,3 +270,161 @@ public fun WorldScaleSceneView(
...
+        }
+    }
+    val lifecycleOwner = LocalLifecycleOwner.current
+    DisposableEffect(locationDataSource) {
+        // This will start the LocationDataSource
+        val wrapper = LocationDataSourceWrapper(locationDataSource)
+        lifecycleOwner.lifecycle.addObserver(wrapper)
+        onDispose {
+            wrapper.onDestroy(lifecycleOwner)
+            lifecycleOwner.lifecycle.removeObserver(wrapper)
```
gunt0001, line 377: I think it' safer to do it in the opposite order:
```suggestion
            lifecycleOwner.lifecycle.removeObserver(wrapper)
            wrapper.onDestroy(lifecycleOwner)
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -310,3 +278,210 @@ public fun WorldScaleSceneView(
...
+                            0.0
+                        )
+                    )
+                    if (!hasSetOriginCamera) {
+                        hasSetOriginCamera = true
+                    }
+
+                    // We have to do this or the error gets bigger and bigger.
+                    cameraController.transformationMatrix =
+                        TransformationMatrix.createIdentityMatrix()
```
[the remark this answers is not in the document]
gunt0001: This change still needs to be made...

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -116,197 +118,212 @@ public fun WorldScaleSceneView(
...
         }
     }
-    DisposableEffect(locationDataSource) {
-        coroutineScope.launch {
-            locationDataSource.start()
-        }
+    val lifecycleOwner = LocalLifecycleOwner.current
+    DisposableEffect(Unit) {
+        // This will start the LocationDataSource
+        val wrapper = LocationDataSourceWrapper(locationDataSource)
```
gunt0001, line 272: This LocationDataSourceWrapper needs to be kept in a remember, otherwise it could be garbage collected. See related comment below on `rememberSystemLocationDataSource`.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -116,197 +118,212 @@ public fun WorldScaleSceneView(
...
+        location.position,
+        distanceUnit = LinearUnit.meters,
+        azimuthUnit = null,
+        curveType = GeodeticCurveType.Geodesic
+    )?.distance ?: return false
+    return distance > WorldScaleParameters.LOCATION_DISTANCE_THRESHOLD_METERS
+}
+
+@Composable
+private fun rememberSystemLocationDataSource(): SystemLocationDataSource {
```
gunt0001, line 316: You could have this return the `LocationDataSourceWrapper` and also do the `DisposableEffect` in this remember function.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -116,197 +114,197 @@ public fun WorldScaleSceneView(
...
+ * Starts the a [SystemLocationDataSource] and periodically updates the [cameraController] with new locations
+ *
+ * @since 200.7.0
+ */
+@Composable
+private fun LocationTracker(
+    cameraController: TransformationMatrixCameraController
+) {
+    val locationDataSource = rememberSystemLocationDataSource().locationDataSource
+    // We should reset the origin camera if the LDS or camera controller changes
```
gunt0001, line 234: Given the location data source won't change, I think this comment needs updating:
```suggestion
    // We should reset the origin camera if the camera controller changes
```

gunt0001 (approved):
@hud10837 - looks good. There are some refinements we can do regarding initialization. For example I noticed that the SceneView may render the globe at full extent (as if looking at it from space) on startup, before we have the initial location. This is something we should be able to avoid. But we can refine that in a separate PR.

PR #717, opened 2025-01-23
changes to use the font specified in the theme

puneet-pdx:
Related to issue: # https://devtopia.esri.com/runtime/kotlin/issues/5234

### Description:

Remove the use of textSize constant and use the text size of the text style mentioned in the theme

### Summary of changes:

- 

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/313/
  - [ ] Yes
  - [ ] No

### DIFF
--- a/toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
+++ b/toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
@@ -117,6 +117,7 @@ public fun Scalebar(
             style,
             colorScheme,
             shapes,
+            labelTypography,
             modifier
         )
     }
@@ -129,6 +130,7 @@ private fun ShowScalebar(
     scalebarStyle: ScalebarStyle,
     colorScheme: ScalebarColors,
     shapes: ScalebarShapes,
+    labelTypography: LabelTypography,
     modifier: Modifier = Modifier
 ) {
     when (scalebarStyle) {
@@ -141,6 +143,7 @@ private fun ShowScalebar(
             maxWidth = maxWidth.toFloat(),
             label = labels[0].label,
             colorScheme = colorScheme,
+            labelTypography = labelTypography
         )
     }
 }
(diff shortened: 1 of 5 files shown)


toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/internal/ScalebarRenderer.kt
```diff
@@ -66,10 +64,11 @@ internal fun LineScalebar(
     maxWidth: Float,
     label: String,
     colorScheme: ScalebarColors,
+    labelTypography: LabelTypography
```
eri9000, line 67: it seems that we only need the property `LabelTypography.labelStyle`, 
I think we should only pass necessary values. 
```suggestion
    labelStyle: TextStyle
```

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/internal/ScalebarRenderer.kt
```diff
@@ -132,10 +132,11 @@ internal fun GraduatedLineScalebar(
     maxWidth: Float,
     tickMarks: List<ScalebarDivision>,
     colorScheme: ScalebarColors,
+    labelTypography: LabelTypography
```
eri9000, line 135: it seems that we only need the property `LabelTypography.labelStyle`, 
I think we should only pass necessary values. 
```suggestion
    labelStyle: TextStyle
```

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
```diff
@@ -141,6 +143,7 @@ private fun ShowScalebar(
             maxWidth = maxWidth.toFloat(),
             label = labels[0].label,
             colorScheme = colorScheme,
+            labelTypography = labelTypography
```
eri9000, line 146: we should only pass what we need, which is the `TextStyle`
```suggestion
            labelStyle = labelTypography.labelStyle
```

eri9000 (commented):
one suggestion regarding the parameters, otherwise LGTM

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
puneet-pdx: How does it even matter here, `labelTypography` just has one property. If we do it here, we will have to do it for `ScalebarColors` and `ScalebarShapes` too, which will unnecessarily increase the parameters being passed.

eri9000 (approved):

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
eri9000: good point 👍🏼

PR #718, opened 2025-01-23
Changes to make graduated line scalebar functional

puneet-pdx:
Related to issue: # https://devtopia.esri.com/runtime/kotlin/issues/5234

### Description:

Changes to make Graduated line scalebar functional

### Summary of changes:

- Updates the logic to determine label size to calcultate the correct number of segments. 
- Update UI to convert offset values in screen pixels
- Add test

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/job/vtest/job/toolkit/316/
  - [ ] Yes
  - [ ] No

### DIFF
--- a/toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
+++ b/toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
@@ -117,6 +117,7 @@ public fun Scalebar(
             style,
             colorScheme,
             shapes,
+            labelTypography,
             modifier
         )
     }
@@ -129,6 +130,7 @@ private fun ShowScalebar(
     scalebarStyle: ScalebarStyle,
     colorScheme: ScalebarColors,
     shapes: ScalebarShapes,
+    labelTypography: LabelTypography,
     modifier: Modifier = Modifier
 ) {
     when (scalebarStyle) {
@@ -141,6 +143,7 @@ private fun ShowScalebar(
             maxWidth = maxWidth.toFloat(),
             label = labels[0].label,
             colorScheme = colorScheme,
+            labelTypography = labelTypography
         )
     }
 }
--- a/toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
+++ b/toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
@@ -198,7 +198,12 @@ internal fun measureAvailableLineDisplayLength(
         ScalebarStyle.GraduatedLine -> {
             // " km" will render wider than " mi"
             val textMeasurer = rememberTextMeasurer()
-            val maxUnitDisplayWidth = textMeasurer.measure(" km", labelTypography.labelStyle).size.width / LocalDensity.current.density
+            val maxUnitDisplayWidth = with(LocalDensity.current) {
+                textMeasurer.measure(
+                    " km",
+                    labelTypography.labelStyle
+                ).size.width.toDp().value
+            }
             maxWidth - (lineWidth.value / 2.0f) - maxUnitDisplayWidth
         }
         ScalebarStyle.Bar,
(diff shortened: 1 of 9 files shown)


toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/Scalebar.kt
```diff
@@ -186,12 +198,12 @@ private fun measureAvailableLineDisplayLength(
         ScalebarStyle.GraduatedLine -> {
             // " km" will render wider than " mi"
             val textMeasurer = rememberTextMeasurer()
-            val maxUnitDisplayWidth = textMeasurer.measure(" km", labelTypography.labelStyle).size.width
-            maxWidth - (lineWidth / 2.0f) - maxUnitDisplayWidth
+            val maxUnitDisplayWidth = textMeasurer.measure(" km", labelTypography.labelStyle).size.width / LocalDensity.current.density
```
eri9000, line 201: would suggest making this an extension function to be more explicit on what this is doing

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/internal/ScalebarRenderer.kt
```diff
@@ -146,9 +147,12 @@ internal fun GraduatedLineScalebar(
             .width(calculateSizeInDp(density, totalWidth))
             .height(calculateSizeInDp(density, totalHeight))
     ) {
+        val tickMarksWithOffsetInScreenSizes = tickMarks.map { tickMark ->
```
eri9000, line 150:
```suggestion
        val tickMarksWithDensityOffsets = tickMarks.map { tickMark ->
```

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/internal/ScalebarUtils.kt
```diff
@@ -125,7 +126,7 @@ internal object ScalebarUtils {
      */
     fun Double.format(): String {
```
eri9000, line 127: doc needs to be updated:
```
     * Formats a double to a string.
     * If the double has no decimal part or the first two decimal places are 0, then it will be displayed as an int.
     * If the second decimal place is zero, it will display the number with one decimal place.
     * Otherwise, it will truncate the number and be displayed with two decimal places.
     *
```

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/internal/ScalebarUtils.kt
```diff
@@ -125,7 +126,7 @@ internal object ScalebarUtils {
...
         return when {
```
eri9000, line 128: This could be simplified by multiplying by the desired decimal places and discarding the rest of the decimals 
```        
return when {
            this % 1.0 == 0.0 || (this * 100).toInt() % 100 == 0  -> this.toInt().toString() // Display as int if very close to an integer
            (this * 10).toInt() % 10 == 0 -> "%.1f".format(this) // Display one decimal place if last digit is zero
            else -> "%.2f".format(this) // Display two decimal places
        }
```

eri9000 (commented):
Looks good, see my comments

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/internal/ScalebarUtils.kt
puneet-pdx: updated

toolkit/scalebar/src/main/java/com/arcgismaps/toolkit/scalebar/internal/ScalebarUtils.kt
```diff
@@ -118,15 +117,15 @@ internal object ScalebarUtils {
...
      * If the second digit after the decimal is zero, it will be displayed with one decimal place.
      * Otherwise, it will be displayed with two decimal places.
      *
      * @since 200.7.0
      */
     fun Double.format(): String {
         return when {
-            (this - this.toInt()).absoluteValue < 1e-12 -> this.toInt().toString() // Display as int if very close to an integer
+            this % 1.0 == 0.0 || (this * 100).toInt() % 100 == 0 -> this.toInt().toString()
             (this * 10) % 1.0 == 0.0 -> "%.1f".format(this) // Display one decimal place if last digit is zero
```
eri9000, line 129: This case will fail with the following: `10.101`
by casing `this` to `Int`, we can discard the rest of the degits after the first decimal place. This allows to check if the 1st decimal place is a 0 more precisely. 
```suggestion
            (this * 100).toInt() % 10 == 0  -> "%.1f".format(this)
```

eri9000 (commented):
One more observation

eri9000 (approved):
👍🏼

puneet-pdx:
@eri9000 thanks, merging

PR #739, opened 2025-02-04
Calibration view

01smito01:
Related to issue: [#5133](https://devtopia.esri.com/runtime/kotlin/issues/5133)

### Description:

Adds the calibration view

### Summary of changes:

- `CalibrationView` composable in `WorldScaleSceneViewScope
- Update microapp to include calibration view

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/372/
  - [ ] Yes
  - [ ] No


toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneViewScope.kt
```diff
@@ -105,4 +184,293 @@ public class WorldScaleSceneViewScope internal constructor(private val sceneView
...
+                        )
+                    }
+                }
+            }
+        }
+    }
+}
+
+internal val LocalColorScheme = compositionLocalOf { DefaultThemeTokens.colorScheme }
+internal val LocalTypography = compositionLocalOf { DefaultThemeTokens.typography }
```
gunt0001, line 328: Could you move these two lines above `CalibrationViewInternal()` since this is where they are used.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneViewScope.kt
```diff
@@ -105,4 +184,293 @@ public class WorldScaleSceneViewScope internal constructor(private val sceneView
...
+                        JoysliderBar(
+                            title = "Heading",
+                            value = heading,
+                            valueFormat = DecimalFormat("+0.#º ; -#.#º"),
+                            minusContentDescription = "Decrease Heading",
+                            plusContentDescription = "Increase Heading",
+                            resetContentDescription = "Reset Heading",
+                            onMinusClick = {
+                                onHeadingChange(-1.0)
+                                heading -= 1
```
gunt0001, line 265: Best to reverse the order here, i.e. first set the value, then invoke the callback. Also applies to the other places but will only mention here.
```suggestion
                                heading -= 1
                                onHeadingChange(-1.0)
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/DefaultThemeTokens.kt
```diff
@@ -0,0 +1,116 @@
...
+        containerColor = ColorTokens.SurfaceVariant,
+        closeButtonContainerColor = ColorTokens.Secondary,
+        closeButtonContentColor = ColorTokens.OnSecondary,
+        buttonContainerColor = ColorTokens.Primary,
+        buttonContentColor = ColorTokens.OnPrimary,
+        sliderTrackColor = ColorTokens.Primary,
+        sliderThumbColor = ColorTokens.OnPrimary
+    )
+
+    val typography = WorldScaleCalibrationViewTypography(
```
gunt0001, line 22:
```suggestion
    val calibrationViewTypography = WorldScaleCalibrationViewTypography(
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/WorldScaleCalibrationViewColorScheme.kt
```diff
@@ -0,0 +1,29 @@
+package com.arcgismaps.toolkit.ar.internal
```
gunt0001, line 1: This class should not be in the internal package, since it is public API.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneViewScope.kt
01smito01: Will need to take `CalibrationViewInternal` out of `WorldScaleSceneViewScope` to do that. Didn't do that so far as there was some reason it wasn't completely trivial, but it shouldn't be much of an effort.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneViewScope.kt
```diff
@@ -106,3 +180,294 @@ public class WorldScaleSceneViewScope internal constructor(private val sceneView
     ): Unit =
         sceneViewScope.Callout(geoElement, modifier, tapLocation, colorScheme, shapes, content)
 }
+
+private var heading by mutableFloatStateOf(0F)
+private var elevation by mutableFloatStateOf(0F)
```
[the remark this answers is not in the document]
01smito01: When the calibration view is closed it is completely removed from the composition, so `remember` and `rememberSaveable` don't help here. I think that was the original reason for having these be global.

I think a workable alternative is to do some state hoisting by making these private properties on `WorldScaleSceneViewScope`, and pass them to `CalibrationViewInternal` from there.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleCalibrationViewDefaults.kt
```diff
@@ -0,0 +1,69 @@
...
+
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.text.TextStyle
+
+public object WorldScaleCalibrationViewDefaults {
+
+    @Composable
+    public fun colorScheme(
```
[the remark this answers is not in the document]
01smito01: `closeButtonContainerColor` and `closeButtonContentColor` don't seem to be used anywhere, so I'll removed them. The close button just uses `buttonContainerColor`/`buttonContentColor`.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneViewScope.kt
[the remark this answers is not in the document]
gunt0001: > I think a workable alternative is to do some state hoisting by making these private properties on WorldScaleSceneViewScope, and pass them to CalibrationViewInternal from there.

Good point. I think these state variable would then need to be defined in `WorldScaleSceneView` and from there be  passed forth and back to the CalibrationView view `WorldScaleSceneViewScope`. I suggest we explore this as part of the other PR that will take care of the persistence. For now, I think moving these variables into `CalibrationViewInternal` is the rigth thing to do. At this point this state doesn't need to be cached when closing the calibration view.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleCalibrationViewDefaults.kt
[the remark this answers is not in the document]
gunt0001: Ah good catch. I think those were leftovers from an earlier iteration. Good to remove...

microapps/ArWorldScaleApp/app/src/main/java/com/arcgismaps/toolkit/arworldscaleapp/screens/MainScreen.kt
```diff
@@ -37,11 +47,29 @@ fun MainScreen() {
...
+    ) {
+        if (displayCalibrationView) {
+            Box(modifier = Modifier.fillMaxSize()) {
+                CalibrationView(
+                    onDismiss = { displayCalibrationView = false },
+                    modifier = Modifier.align(Alignment.BottomCenter),
+                )
+            }
+        } else {
+            FloatingActionButton(onClick = { displayCalibrationView = true }) {
```
hud10837, line 67: FYI -- I moved this to the bottom right in my branch using a Box and modifier.align, but I don't know if it's necessary for this PR which just gets things working

microapps/ArWorldScaleApp/app/src/main/res/values/strings.xml
```diff
@@ -19,4 +19,5 @@
 <resources>
     <string name="app_name">ArWorldScaleApp</string>
     <string name="arcore_not_installed_screen_message">Google Play Services for AR must be installed to run this app.</string>
+    <string name="calibration_view_button_description">Calibration View button</string>
```
hud10837, line 22: Nitpick: I think this should describe what it does? ie. `Show Calibration View`

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneViewScope.kt
```diff
@@ -34,7 +69,47 @@ import com.arcgismaps.toolkit.geoviewcompose.theme.CalloutShapes
...
+     * @since 200.7.0
+     */
+    @Composable
+    public fun CalibrationView(
+        onDismiss: () -> Unit,
+        modifier: Modifier = Modifier,
+        colorScheme: WorldScaleCalibrationViewColorScheme = WorldScaleCalibrationViewDefaults.colorScheme(),
+        typography: WorldScaleCalibrationViewTypography = WorldScaleCalibrationViewDefaults.typography(),
+    ) {
+        WorldScaleCalibrationViewTheme {
```
hud10837, line 98: Are we supposed to explicitly use this theme here? Wouldn't this override the user's theme? If we remove it, I think we can also remove the entire `theme` directory that was added.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneViewScope.kt
```diff
@@ -106,3 +181,294 @@ public class WorldScaleSceneViewScope internal constructor(private val sceneView
...
+            )
+        }
+
+        FilledIconButton(
+            onClick = onResetClick,
+            colors = IconButtonDefaults.filledIconButtonColors(
+                containerColor = LocalColorScheme.current.buttonContainerColor,
+                contentColor = LocalColorScheme.current.buttonContentColor,
+                disabledContainerColor = LocalColorScheme.current.buttonContainerColor,
+                disabledContentColor = LocalColorScheme.current.buttonContentColor
```
hud10837, line 384: Is it right for these to be the same as the non-disabled versions?

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneViewScope.kt
```diff
@@ -106,3 +181,294 @@ public class WorldScaleSceneViewScope internal constructor(private val sceneView
...
+            ),
+            modifier = Modifier.alpha(if (value != 0F) 1F else 0F)
```
hud10837, line 386: Will this still allow the button to be clicked while invisible? If so that may cause an issue for accessibility tools. I think you could potentially use an if statement to conditionally show the button, and if that messes up the positioning, use a spacer, but we could discuss

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/WorldTrackingCameraController.kt
```diff
@@ -87,7 +90,19 @@ internal class WorldTrackingCameraController(private val onLocationDataSourceFai
...
+                deltaRoll = 0.0
+            ).elevate(elevationOffset)
+        )
+
+        totalHeadingOffset += headingOffset
+        totalElevationOffset += elevationOffset
+    }
 
     /**
      * Sets the origin position of the camera to the given location.
```
hud10837, line 108: This function should still be updated to take into account the offsets

toolkit/ar/src/main/res/drawable-anydpi/ic_action_add_heading.xml
hud10837, line 1: Nitpick: rename these to just "add" and "reduce"?

toolkit/ar/src/main/res/drawable/ic_launcher_background.xml
hud10837, line 1: Do we need to add a launcher icon to the module? I think this could be removed

toolkit/ar/src/main/res/drawable/ic_launcher_foreground.xml
hud10837, line 1: As above, I think this could be removed

hud10837 (changes requested):

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneViewScope.kt
01smito01: It's been that way [since prototype](https://devtopia.esri.com/gunt0001/projects/blob/4fd7948262aefcb0b23c159dc1eab4da01f09a67/kotlin-research/WorldScaleCalibrationViewProto/app/src/main/java/com/example/worldscalecalibrationviewproto/CalibrationView.kt#L203), at least. But these buttons are never not enabled.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneViewScope.kt
01smito01: It's still enabled and in the composition even when hidden, yes. I did initially have it entering and exiting the composition but couldn't get it to not affect the positioning of other elements. Will have another go at that.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/WorldTrackingCameraController.kt
hud10837: Yes, when the origin camera is updated using the location, it should still take into account the offsets, because this function is called periodically when the location changes and is significantly different enough from the current origin camera position to reset the origin camera. So if the LDS location gets really far from the current camera position calculated by ARCore, then we will update the origin camera, and we don't want to lose the offsets

hud10837 (approved):
LGTM 👍

PR #764, opened 2025-02-26
Colinanderson/vtest problems

colinanderson:
Related to issue: #

### Summary of changes:

- 

### Pre-merge Checklist

  - [ ] link:
  - [ ] Yes
  - [ ] No


toolkit/utilitynetworks/api/utilitynetworks.api
kaushikrw, line 1: I think for the purposes of this PR, the api files changes are okay. But it seems like some internal compose functions should not be exposed here. Perhaps we can fix it separately.

kaushikrw (approved):

toolkit/utilitynetworks/api/utilitynetworks.api
```diff
@@ -82,7 +82,6 @@ public final class com/arcgismaps/toolkit/utilitynetworks/TraceState {
 
 public final class com/arcgismaps/toolkit/utilitynetworks/ui/ComposableSingletons$TraceResultScreenKt {
 	public static final field INSTANCE Lcom/arcgismaps/toolkit/utilitynetworks/ui/ComposableSingletons$TraceResultScreenKt;
-	public static field lambda-1 Lkotlin/jvm/functions/Function3;
```
gunt0001, line 85: I have no idea why this is not being dumped anymore with version `0.17.0`. I scanned over the [release notes](https://github.com/Kotlin/binary-compatibility-validator/releases) but nothing stood out. Does anyobody know if there is a doc/spec of the api dump? I would like to figure out what this line actually refers to.

colinanderson: Looks like https://github.com/Kotlin/binary-compatibility-validator/blob/master/README.md is the doc.

colinanderson: Maybe we don't need to jump 0.14.0 to 0.17.0 - one of the intermediate updates _might_ fix the problem without introducing changes.

puneet-pdx (approved):
Approving this. `TraceResultScreen` is internal fun, I am not sure why it is showing up in the api file

toolkit/utilitynetworks/api/utilitynetworks.api
colinanderson: Looks like 0.16.0 is the lowest version we can use and it isn't complaining about the api file so looks like we're stuck with that change.

gunt0001: I think we should upgrade to `0.17.0`, otherwise we will see these changes in a future upgrade anyway, might as well deal with it now.

toolkit/utilitynetworks/api/utilitynetworks.api
gunt0001: I agree, I believe at least line 84 refers to the internal compose function `TraceResultScreen`, which we should not dump. @puneet-pdx - thoughts?

gunt0001:
@puneet-pdx 

> TraceResultScreen is internal fun, I am not sure why it is showing up in the api file

See explanation in [our doc.](https://devtopia.esri.com/runtime/kotlin/blob/main/doc/toolkit/contribution_guidelines.md#maintaining-binary-compatibility)

toolkit/utilitynetworks/api/utilitynetworks.api
kaushikrw: I believe it is a known issue in the binary compatibility validator. We had to specify an ignore list like [this](https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/blob/01ce496b7aee7e2313d292984cfaf7b8f31f2f00/toolkit/featureforms/build.gradle.kts#L85) in feature forms.

kaushikrw:
Merging this PR.

PR #765, opened 2025-02-26
Fix init of elevation/heading offsets in CalibrationView

01smito01:
Related to issue: [#5397](https://devtopia.esri.com/runtime/kotlin/issues/5397)

### Description:
Heading and elevation calibration offsets were previously stored in `CalibrationViewInternal`. When the calibration view was dismissed, these values would exit the composition and be lost. They would still be applied to the camera, but reopening the calibration view would incorrectly report these values as zero.

This PR fixes the issue by hoisting these values to `CalibrationState`, a state holder class, so they are not lost when the calibration view is closed.

### Summary of changes:

- Add `CalibrationState` class and associated modifications to `WorldScaleSceneView`, `WorldScaleSceneViewScope`.

### Pre-merge Checklist

  - [ ] link: vTest is [still a bit fried](https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/475/console). Colin's working on it, will kick off a proper vTest once that's resolved.
  - [ ] Yes
  - [ ] No


toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/CalibrationState.kt
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * State holder class for calibration view
+ *
+ * @param onHeadingChange Lambda invoked when the user adjusts heading offset
+ * @param onHeadingReset Lambda invoked when the user resets heading offset
+ * @param onElevationChange Lambda invoked when the user adjusts elevation offset
+ * @param onElevationReset Lambda invoked when the user resets elevation offset
+ */
+@Immutable
```
gunt0001, line 35: Having re-read the [doc](https://developer.android.com/develop/ui/compose/performance/stability#types), I think semantically it makes more sense to mark this with `@Stable`. In compose "stable" means "a type whose properties can change after construction. If and when those properties change during runtime, Compose becomes aware of those changes." - I think that's the case here.
```suggestion
@Stable
```

```
When applied to a class or an interface, Stable indicates that the following must be true:
- The result of equals will always return the same result for the same two instances.
- When a public property of the type changes, composition will be notified.
- All public property types are stable.
```

gunt0001 (approved):
@01smito01 - looks great, some minor comments.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/CalibrationState.kt
```diff
@@ -0,0 +1,74 @@
...
+internal class CalibrationState(private val scope: CoroutineScope) {
+    private val _headingDeltas: MutableSharedFlow<Double> = MutableSharedFlow()
+    val headingDeltas = _headingDeltas.asSharedFlow()
+
+    private val _elevationDeltas: MutableSharedFlow<Double> = MutableSharedFlow()
+    val elevationDeltas = _elevationDeltas.asSharedFlow()
+
+    var totalHeadingOffset by mutableDoubleStateOf(0.0)
+    var totalElevationOffset by mutableDoubleStateOf(0.0)
+
```
hud10837, line 46: I think these could have `private set` to prevent anyone else from editing them

hud10837 (approved):
Looks good, works well with ad-hoc testing

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/CalibrationState.kt
```diff
@@ -0,0 +1,74 @@
...
+import kotlinx.coroutines.launch
+
+
+/**
+ * State holder class for calibration view
+ *
+ * @since 200.7.0
+ */
+@Stable
+internal class CalibrationState(private val scope: CoroutineScope) {
```
gunt0001, line 37: No need for this scope property if you use `tryEmit`, see comments below...
```suggestion
internal class CalibrationState() {
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/CalibrationState.kt
```diff
@@ -0,0 +1,74 @@
...
+    val elevationDeltas = _elevationDeltas.asSharedFlow()
+
+    var totalHeadingOffset by mutableDoubleStateOf(0.0)
+    var totalElevationOffset by mutableDoubleStateOf(0.0)
+
+    fun onHeadingChange(value: Double) {
+        scope.launch {
+            _headingDeltas.emit(value)
+            totalHeadingOffset += value
+        }
```
gunt0001, line 51: You don't need a scope if you use `tryEmit`. 
```suggestion
        _headingDeltas.tryEmit(value)
        totalHeadingOffset += value
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/CalibrationState.kt
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    fun onElevationChange(value: Double) {
+        scope.launch {
+            _elevationDeltas.emit(value)
+            totalElevationOffset += value
+        }
```
gunt0001, line 58: You don't need a scope if you use `tryEmit`:
```suggestion
       _elevationDeltas.tryEmit(value)
       totalElevationOffset += value
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/CalibrationState.kt
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    fun onHeadingReset() {
+        scope.launch {
+            _headingDeltas.emit(-totalHeadingOffset)
+            totalHeadingOffset = 0.0
+        }
```
gunt0001, line 65: You don't need a scope if you use `tryEmit`:
```suggestion
        _headingDeltas.tryEmit(-totalHeadingOffset)
        totalHeadingOffset = 0.0
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/CalibrationState.kt
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    fun onElevationReset() {
+        scope.launch {
+            _elevationDeltas.emit(-totalElevationOffset)
+            totalElevationOffset = 0.0
+        }
```
gunt0001, line 72: You don't need a scope if you use `tryEmit`:
```suggestion
        _elevationDeltas.tryEmit(-totalElevationOffset)
        totalElevationOffset = 0.0
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/WorldTrackingCameraController.kt
```diff
@@ -90,7 +90,7 @@ internal class WorldTrackingCameraController(private val onLocationDataSourceFai
      *
      * @since 200.7.0
      */
-    internal fun updateCamera(headingOffset: Double, elevationOffset: Double) {
+    internal fun updateCamera(headingOffset: Double = 0.0, elevationOffset: Double = 0.0) {
```
gunt0001, line 93: We will never call this function without parameters, thus it seems a code smell that both parameters are defaulted. A cleaner approach would be to break these up in two separate functions:
- updateCameraHeading(headingOffset: Double)
- updateCameraElevation(elevationOffset: Double)

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/WorldTrackingCameraController.kt
```diff
@@ -114,31 +111,13 @@ internal class WorldTrackingCameraController(private val onLocationDataSourceFai
             Camera(
                 location.position.y,
                 location.position.x,
-                if (location.position.hasZ) location.position.z ?: totalElevationOffset else totalElevationOffset,
-                totalHeadingOffset,
+                if (location.position.hasZ) location.position.z ?: calibrationState.totalElevationOffset else calibrationState.totalElevationOffset,
+                calibrationState.totalElevationOffset,
```
gunt0001, line 115: Shouldn't this apply the heading instead of elevation?
```suggestion
                calibrationState.totalHeadingOffset,
```

gunt0001 (commented):
@01smito01 - great refactor, a few comments.

PR #775, opened 2025-03-10
Add clipping distance and fix NaN elevation crash

hud10837:
Related to issue: https://devtopia.esri.com/runtime/kotlin/issues/5223

### Summary of changes:

- Adds clipping distance
- Ensures that NaN elevations are filtered out, which caused a crash on some devices
- Makes the `headings` flow in the `WorldScaleNmeaLocationProvider` an empty flow and always sets the course of the locations in the locations flow to whatever is available in the Nmea location, because these headings are not being used anyway and are not planned to be used as there will be a different mechanism for gathering headings.

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/500/
  - [ ] https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/505/
  - [ ] Yes
  - [x] No


toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/WorldTrackingCameraController.kt
```diff
@@ -260,6 +265,8 @@ internal fun shouldUpdateCamera(
         || location.verticalAccuracy.isNaN()
     ) return false
 
+    // filter out locations with a NaN z value
+    if (location.position.z?.isNaN() == true) return false
```
gunt0001, line 269: Once we use the fused orientation provider, should we remove this check again? At this point we won't care about the z value, just the lat/lon.

gunt0001 (approved):
@hud10837 - looks good, one question

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/WorldTrackingCameraController.kt
hud10837: I don't think I understand. How will the fused orientation provider affect the altitude of the position?

gunt0001: Ah my bad, ignore my comment. I was thinking of "elevation provider" which we decided we won't use at this point.

01smito01 (approved):
LGTM

hud10837:
Thanks both, merging

PR #776, opened 2025-03-10
`FeatureForm` : support for UtilityAssociations

kaushikrw:
Related to issue: #[kotlin/5456](https://devtopia.esri.com/runtime/kotlin/issues/5456)

### Description:

Adds UtilityAssociations support for `FeatureForm`.

### UtilityAssociations UI structure 

-  A new `UtilityAssociationsFormElement` is displayed if it is part of the form info.
    - First step is to call `element.fetchAssociationsFilterResults` which fetches the list of associations organized within a specific structure represented by the `UtilityAssociationsFilterResult` class.
    - Once fetched, this element will display the list of `UtilityAssociationsFilter`s.
    - Each filter will have a list of `UtilityAssociations` that are grouped by layer, sublayer, table or subtable. This is represented by the `UtilityAssociationGroupResult` class.
    - Finally, once you click on a group, this will display the actual `UtilityAssociation`s present in the `UtilityAssociationGroupResult`.

### Navigation Support

- The component allows you to navigate to the `FeatureForm` of an `ArcGISFeature` on the other end of the `UtilityAssociation`. This is achieved by adding a navigation graph within the component.
- Since with compose navigation only `Serializable` or `Parcelable` objects can be saved, for example during configuration changes, this means the `FeatureForm` api object cannot be used here and has to hoisted out of the navigation graph.
- Hence, a `Stack` data structure is created out of the composition that will track all the `FeatureForm`s visited. This stack will closely track the current navigation stack and provide the source of truth for the navigation destinations. 

### API Changes

Since the `Stack` has to be out of the composition, this means providing a State holder class for the `FeatureForm` composable. This state holder can then be hoisted up, like in a `ViewModel`. Hence the proposed API changes -

- A new `FeatureFormState` class which internally creates and holds on to this `Stack` of `FeatureForm` api objects.
- Provides an observable `activeFeatureForm` property that will always point to the current `FeatureForm` instance being edited.
- Deprecate the older `FeatureForm` composable in favor of the new one.
    - Internally, the deprecated  `FeatureForm` composable utilizes the `FeatureFormState`.
    - This composable will also not display any `UtilityAssociationsFormElement`s and will function and look exactly like it used to.
- Deprecated the `ValidationErrorVisibility` class as the updated workflow specifies to show all errors when the form is opened. This setting will possibly be part of the web map spec in a future release.
- `FeatureForm` composable provides an optional `onDimiss` callback. If this is provided, a "close" icon is displayed which when clicked raises this event.
- `FeatureForm` composable provides a boolean `showFormActions`. If this is true (default), A "Discard" and "Save" buttons within an action bar is displayed when there are edits in the form.
- A new `FeatureFormEditingEvent` is introduced which is raised when the user "Saves" or "Discards" edits on the Form. (see ui changes below).

### UI Changes

- `FeatureForm.hasEdits` property is used to show an action bar on top of the form if `showFormActions` is true which enables two options -
    - "Save" - saves the edits using `FeatureForm.finishEditing()` and raises the  `FeatureFormEditingEvent.SavedEdits` if only the result of the operation is successful.
    - "Discard" - discards any edits using `FeatureForm.discarEdits()` and raises the  `FeatureFormEditingEvent.DiscardedEdits`.
- "Close" - if the corresponding `onDimiss` callback is provided, a close icon is shown. If there are active edits on the form as indicated by `FeatureForm.hasEdits`, a dialog is presented to save or discard edits. These actions will have the same behavior as above.
- When attempting to navigate to another `Feature`, the same workflow applies. If there are edits, a dialog is presented to save/discard.
- If the user attempts to save and there are validation errors as indicated by `FeatureForm.validationErrors`, a dialog is presented with this information.

These changes should satisfy the workflows mentioned in the issue [here](https://devtopia.esri.com/runtime/apollo/issues/1118).

### Internal Changes

- Since any internal states that are created for other `FormElement`s are now stored in the `FeatureFormState` class, this means we no longer need any `rememberSaveable` methods on these elements. 

### Test Changes

- Updated the tests that used any of the removed APIs.

### Test Data

I've added some test data in the linked issue.

### Notes

- Tests for UN elements will be added in a separate PR. I will also update the deprecated constructor for the other tests here.
- Readme and doc will be updated in a separate PR.

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/job/vtest/job/toolkit/526/
  - [ ] Yes
  - [ ] No

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/UtilityAssociations.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/UtilityAssociations.kt
@@ -36,6 +36,7 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.painter.Painter
 import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import com.arcgismaps.Guid
 import com.arcgismaps.data.ArcGISFeature
@@ -192,7 +193,7 @@ private fun AssociationItem(
     }
     val trailingText = when {
         terminal != null -> {
-            "Terminal : ${terminal.name}"
+            stringResource(R.string.terminal, terminal.name)
         }
 
         fractionAlongEdge != null -> {
@@ -201,7 +202,7 @@ private fun AssociationItem(
 
         association.associationType is UtilityAssociationType.Containment &&
             associatedFeature.globalId == association.toElement.globalId -> {
-            "Containment Visible : ${association.isContainmentVisible}"
+            stringResource(R.string.containment_visible, association.isContainmentVisible)
         }
 
         else -> ""
--- a/toolkit/featureforms/src/main/res/values/strings.xml
+++ b/toolkit/featureforms/src/main/res/values/strings.xml
@@ -153,6 +153,8 @@
     <string name="the_form_has_validation_errors">The form has validation errors</string>
     <string name="the_form_has_edits">The form has edits</string>
     <string name="do_you_want_to_save_your_changes">Do you want to save your changes?</string>
+    <string name="terminal">Terminal : %1$s</string>
+    <string name="containment_visible">Containment Visible : %1$s</string>
     <plurals name="you_have_errors_that_must_be_fixed_before_saving">
         <item quantity="zero">You have %1$d errors that must be fixed before saving.</item>
         <item quantity="one">You have %1$d error that must be fixed before saving.</item>


toolkit/featureforms/build.gradle.kts
```diff
@@ -23,6 +23,7 @@ plugins {
     id("artifact-deploy")
     id("com.google.android.libraries.mapsplatform.secrets-gradle-plugin")
     alias(libs.plugins.binary.compatibility.validator) apply true
+    alias(libs.plugins.kotlin.serialization) apply true
```
kaushikrw, line 26: Serialization plugin is needed for compose navigation. Specifically, for the route classes used.

toolkit/featureforms/src/androidTest/java/com/arcgismaps/toolkit/featureforms/DateTimeFieldTests.kt
```diff
@@ -99,7 +98,7 @@ class DateTimeFieldTests : FeatureFormTestRunner(
         }
         dateTimeField.assertIsDisplayed()
         val helper = dateTimeField.onChildWithContentDescription("supporting text")
-        val helperMatcher = hasText("Date Entry is Required")
+        val helperMatcher = hasText("Required")
```
kaushikrw, line 101: Updated the correct text to check for.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureFormState.kt
```diff
@@ -0,0 +1,507 @@
...
+            form = this.featureForm,
+            elements = this.featureForm.elements,
+            scope = coroutineScope
+        )
+        // Add the provided state collection to the store.
+        store.addLast(FormStateData(featureForm, states))
+        evaluateExpressions()
+    }
+
+    internal constructor(
```
kaushikrw, line 145: This constructor is to primarily support the deprecate `FeatureForm` composable which creates its own `FormStateCollection`.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureFormState.kt
```diff
@@ -0,0 +1,507 @@
...
+ * @param form the [FeatureForm] to create the states for.
+ * @param scope a [CoroutineScope] to run collectors and calculations on.
+ *
+ * @return returns the [FormStateCollection] created.
+ */
+internal fun createStates(
+    form: FeatureForm,
+    elements: List<FormElement>,
+    scope: CoroutineScope,
+    ignoreList : Set<Class<out FormElement>> = emptySet(),
```
kaushikrw, line 249: This param allows the deprecated `FeatureForm` composable to not display any `FormElement`s such as `UtillityAssociationsFormElement`.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
```diff
@@ -183,213 +325,646 @@ public fun FeatureForm(
...
  * @param typography The [FeatureFormTypography] to use for the FeatureForm.
  *
- * @since 200.4.0
+ * @since 200.7.0
  */
 @Composable
 public fun FeatureForm(
-    featureForm: FeatureForm,
+    featureFormState: FeatureFormState,
+    onDismiss: (() -> Unit)?,
```
kaushikrw, line 365: Should this use a different mechanism to hide the close icon? For ex, use a `Boolean` instead of an optional parameter.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/attachment/AttachmentElementState.kt
```diff
@@ -159,8 +149,6 @@ internal class AttachmentElementState(
         state.loadWithParentScope()
         // scroll to the new attachment after a delay to allow the recomposition to complete
         scope.launch {
-            delay(100)
-            lazyListState.scrollToItem(0)
```
kaushikrw, line 163: Moved this into the composition as I was seeing an exception about doing this `scrollToItem` during the measurement phase.

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapViewModel.kt
```diff
@@ -589,3 +477,33 @@ val List<FeatureEditResult>.errors: List<Throwable>
...
+ */
+fun ArcGISMap.clearSelection() {
+    operationalLayers.forEach { layer ->
+        when (layer) {
+            is SubtypeFeatureLayer -> {
+                layer.clearSelection()
+            }
+            is FeatureLayer -> {
+                layer.clearSelection()
+            }
```
eri9000, line 492: these could be combined

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureFormState.kt
```diff
@@ -0,0 +1,487 @@
...
+import com.arcgismaps.toolkit.featureforms.internal.components.text.TextFieldProperties
+import com.arcgismaps.toolkit.featureforms.internal.components.text.TextFormElementState
+import com.arcgismaps.toolkit.featureforms.internal.components.utilitynetwork.UtilityAssociationsElementState
+import com.arcgismaps.toolkit.featureforms.internal.utils.fieldIsNullable
+import com.arcgismaps.toolkit.featureforms.internal.utils.toMap
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+
+/**
+ * The state object for a [FeatureForm] used by the [FeatureForm] composable. This class is
```
eri9000, line 70: linking to composable is broken as it takes me to the `FeatureForm` class instead
```suggestion
 * The state object for a [FeatureForm] used by the FeatureForm composable. This class is
```

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapScreen.kt
```diff
@@ -232,19 +210,27 @@ fun MapScreen(mapViewModel: MapViewModel = hiltViewModel(), onBackPressed: () ->
...
                 enter = slideInVertically { h -> h },
                 exit = slideOutVertically { h -> h },
                 label = "feature form"
             ) {
-                val isSwitching = uiState is UIState.Switching
                 // remember the form and update it when a new form is opened
-                val rememberedForm = remember(this, isSwitching) {
-                    featureForm!!
+                val rememberedForm = remember(this) {
+                    featureFormState!!
```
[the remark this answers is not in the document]
kaushikrw: That is by design though. Within the `AnimatedVisibility` scope, it would never be null since it is visible only when `featureFormState != null`.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureFormState.kt
```diff
@@ -0,0 +1,487 @@
...
+ *
+ * @param featureForm the [FeatureForm] to create the state for.
+ *
+ * @since 200.7.0
+ */
+@Stable
+public class FeatureFormState private constructor(
+    private val featureForm: FeatureForm
+) {
+    private val store: ArrayDeque<FormStateData> = ArrayDeque()
```
eri9000, line 92: seems like you're only working with the last element, any specific reason you're using a Deque?

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapScreen.kt
```diff
@@ -306,17 +269,23 @@ fun MapScreen(mapViewModel: MapViewModel = hiltViewModel(), onBackPressed: () ->
...
+    if (errors != null) {
+        ErrorDialog(
+            error = errors!!,
+            onContinue = {
+                mapViewModel.clearErrors()
             },
-            onCancel = {
-                showDiscardEditsDialog = false
+            onDismissRequest = {
+                //mapViewModel.rollbackEdits()
```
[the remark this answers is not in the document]
kaushikrw: Removed.

eri9000 (commented):
a few more comments

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapScreen.kt
```diff
@@ -459,11 +429,11 @@ fun FeatureFormSheet(
...
             FeatureForm(
-                featureForm = featureForm,
+                featureFormState = state,
                 modifier = Modifier.fillMaxSize(),
-                validationErrorVisibility = errorVisibility
+                onDismiss = onDismiss,
+                onEditingEvent = onEditingEvent
             )
         }
     }
```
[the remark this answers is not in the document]
kaushikrw: The `FeatureForm` component will now show this dialog. This can be removed.

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/screens/map/MapScreen.kt
```diff
@@ -271,29 +257,6 @@ fun MapScreen(mapViewModel: MapViewModel = hiltViewModel(), onBackPressed: () ->
...
-                    mapViewModel.cancelCommit()
-                },
-                onDismissRequest = {
-                    mapViewModel.rollbackEdits()
-                }
-            )
-        }
-
         // show pick a feature dialog if the layer is a sublayer
         is UIState.SelectFeature -> {
```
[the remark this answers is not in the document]
kaushikrw: Possibly. I think that would be a good idea. It would also mean moving other screens like the `FeatureFormSheet` and `AddFeatureSheet` into separate destinations. If its okay, I will tackle this in a separate PR.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/UtilityAssociationsElement.kt
```diff
@@ -0,0 +1,167 @@
...
+ * @param modifier The [Modifier] to apply to this layout.
+ */
+@Composable
+private fun Filters(
+    filterResults: List<UtilityAssociationsFilterResult>,
+    onClick: (Int) -> Unit,
+    modifier: Modifier = Modifier
+) {
+    Surface(modifier = modifier) {
+        Column {
```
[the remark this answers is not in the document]
kaushikrw: They are not scrollable according to the [design](https://www.figma.com/design/xewoyOCqOjhCpb0OwSk6d8/Utility-Network-Design?node-id=122-1552).

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureFormState.kt
kaushikrw: It is essentially a `Stack` of visited `FeatureForm`s. The `Push` and `Pop`, hence are tied to the actions of the navigation. When you navigate to a new `FeatureForm` via an `UtilityAssociation`, a `Push` operation adds this new form to the stack. Similarly, when you hit "Back", a `Pop` operation removes the current form and brings the previous one to the top.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/utils/Dialog.kt
```diff
@@ -160,13 +163,24 @@ internal sealed class DialogType {
...
+        val body: String
+    ) : DialogType()
 }
 
 /**
  * Shows the appropriate dialogs as requested by the [LocalDialogRequester].
  */
 @Composable
-internal fun FeatureFormDialog(states: FormStateCollection) {
+internal fun FeatureFormDialog(states: FormStateCollection, navHostController: NavHostController) {
```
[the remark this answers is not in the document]
kaushikrw: I would agree but this needed to solve the following problem -

The `DialogType.SaveFeatureDialog` data class has `onSave` and `onDiscard` function types. These are initially defined with a body that uses the current `NavHostController` to perform navigation. However, after an orientation change, the composition has a new `NavHostController` instance but the saved function type is still referencing the old instance. 

Hence, the `FeatureFormDialog` doesn't actually consume the `NavHostController` directly. It merely passes it to any long living lambda's that are not part of the composition, but need to perform navigation actions (like in response to a user click).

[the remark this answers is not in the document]
kaushikrw: I could alternatively move the `SaveFeatureDialog` into the primary composition, which would also solve the issue but would have to be called from Multiple places. Thoughts?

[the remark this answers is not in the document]
kaushikrw: @eri9000 I've updated this so that the `NavHostController` is not passed. Instead, the navigation is now hoisted at the `FeatureFormState` level.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
```diff
@@ -183,213 +324,655 @@ public fun FeatureForm(
...
+                )
+                FeatureFormDialog(states)
+            }
+        }
+    }
+    DisposableEffect(state) {
+        onDispose {
+            state.setNavigationCallback(null)
+            state.setNavigateBack(null)
+        }
```
kaushikrw, line 686: Added `DisposableEffect` to set the navigation events back to null.

eri9000 (commented):
Good work, a few more things.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
eri9000: > use a showCloseIcon: Boolean instead of an optional parameter.

I think that would be a good replacement

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/utils/Dialog.kt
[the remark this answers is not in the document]
eri9000: updated code looks good 👍🏼

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureFormState.kt
```diff
@@ -0,0 +1,487 @@
...
+        store.addLast(FormStateData(form, states))
+        _activeFeatureForm.value = form
+        evaluateExpressions()
+    }
+
+    /**
+     * Pops the current [FeatureForm] from the stack and sets the previous form as the [activeFeatureForm].
+     */
+    internal fun popBackStack(): Boolean {
+        return if (store.size <= 1) {
```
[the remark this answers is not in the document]
kaushikrw: It's actually meant to check there is more than 1 element in the stack.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureFormState.kt
```diff
@@ -0,0 +1,487 @@
...
+                states.add(element, state)
+            }
+
+            else -> {}
+        }
+    }
+    // The Toolkit currently only supports AttachmentsFormElements via the
+    // default attachments element. Once AttachmentsFormElements can be authored
+    // the switch case above should have a case added for AttachmentsFormElement.
+    if (form.defaultAttachmentsElement != null) {
```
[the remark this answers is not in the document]
kaushikrw: The `FeatureForm.defaultAttachmentsElement` is not part of the elements list though. Plus, it needs to be displayed at the bottom of the form.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
kaushikrw: Updated!

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureFormState.kt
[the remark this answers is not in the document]
eri9000: gotcha 👍🏼

toolkit/featureforms/api/featureforms.api
```diff
@@ -1,7 +1,56 @@
+public final class com/arcgismaps/toolkit/featureforms/ComposableSingletons$FeatureFormKt {
```
sorenoid, line 1: should add this to ignored types in build.gradle.kts

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/UtilityAssociations.kt
```diff
@@ -0,0 +1,312 @@
...
+            }
+        }
+
+        else -> {
+            Pair(null, null)
+        }
+    }
+    val trailingText = when {
+        terminal != null -> {
+            "Terminal : ${terminal.name}"
```
sorenoid, line 195: Shouldn't these be localized?

sorenoid (changes requested):
Looks good overall. I have a couple questions.

toolkit/featureforms/api/featureforms.api
kaushikrw: This is the public API though.

kaushikrw:
Changed the base back to `feature-branches/forms`.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/UtilityAssociations.kt
kaushikrw: We don't usually localize property names, do we?

kaushikrw: Looks like we can add localization to these properties.

toolkit/featureforms/api/featureforms.api
sorenoid: In the past we have ignored the "ComposableSingletons" which leak into the public API on behalf of internal composable functions. For example [here](https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/blob/063f96f69693b35705c1d22eccd38c8c14411223/toolkit/featureforms/build.gradle.kts#L94). I am not convinced that was the right thing to do.

sorenoid (approved):
Just one more comment. Happy to approve however you handle it.

toolkit/featureforms/api/featureforms.api
kaushikrw: That's a good point. I am not sure where it is come from too. It does not seem to be from the HIDDEN deprecation level functions. I'll try to narrow it down.

kaushikrw: @sorenoid Looks like its from the `FeatureFormTitle` composable. I don't understand why though as there is nothing special about this function. Nevertheless, this specific composable is removed in the follow up enhancements [PR](https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/pull/783/files#diff-abf396b5ae994999e6ef4dab255146bc3fc8161653c280f3627e6c4064baf320) and it fixes the api dump.

kaushikrw:
Thanks @sorenoid and @eri9000, this was a pretty big one!

PR #808, opened 2025-03-31
retrieve artifactory info from local.properties

akh13771:
Related to issue: 

### Description:
Removing internal Artifactory information from public repo to `local.propreties`

### Summary of changes:
- Removed artifactory URL from `gradle.properties`
- Updated `settings.gradle.kts` to retrieve credentials from command line first then `local.properties`
- Updated `build.gradle.kts`  to retrieve credentials from command line first then `local.properties` then set the credentials as project variables to be used for other tasks in the build

### Pre-merge Checklist

- a vTest Job for this PR has been run to test retrieving credentials from the command line
  - [ ] link: 
- Changes were test as well using `./ci/run_gradle_task.sh -t assembleDebug`
  - [ ] Yes
  - [ ] No

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -39,6 +39,30 @@ buildscript {
         // before any dependent subproject uses its symbols to configure a dokka task.
         classpath(libs.dokka.versioning)
     }
+    val localProperties = java.util.Properties().apply {
+        val localPropertiesFile = file("local.properties")
+        if (localPropertiesFile.exists()) {
+            load(localPropertiesFile.inputStream())
+        }
+    }
+    // Find these in properties passed through command line or read from local.properties and
+    // set them as project properties
+    val artifactoryUrl: String? = project.findProperty("artifactoryUrl") as String?
+        ?: localProperties.getProperty("artifactoryUrl")
+        ?: ""
+    val artifactoryUsername: String? = project.findProperty("artifactoryUsername") as String?
+        ?: localProperties.getProperty("artifactoryUsername")
+        ?: ""
+    val artifactoryPassword: String? = project.findProperty("artifactoryPassword") as String?
+        ?: localProperties.getProperty("artifactoryPassword")
+        ?: ""
+
+    if (artifactoryUrl != "") {
+        project.extra.set("artifactoryUrl", artifactoryUrl)
+        project.extra.set("artifactoryUsername", artifactoryUsername)
+        project.extra.set("artifactoryPassword", artifactoryPassword)
+    }
+
     val finalBuild: Boolean = (project.properties["finalBuild"] ?: "false")
         .run { this == "true" }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -49,7 +49,7 @@ buildscript {
     // set them as project properties
     val artifactoryUrl: String? = project.findProperty("artifactoryUrl") as String?
         ?: localProperties.getProperty("artifactoryUrl")
-        ?: ""
+        ?: "https://esri.jfrog.io/artifactory/arcgis"
     val artifactoryUsername: String? = project.findProperty("artifactoryUsername") as String?
         ?: localProperties.getProperty("artifactoryUsername")
         ?: ""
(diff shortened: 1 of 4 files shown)


build.gradle.kts
```diff
@@ -39,6 +39,30 @@ buildscript {
...
+        ?: ""
+    val artifactoryPassword: String? = project.findProperty("artifactoryPassword") as String?
+        ?: localProperties.getProperty("artifactoryPassword")
+        ?: ""
+
+    if (artifactoryUrl != "") {
+        project.extra.set("artifactoryUrl", artifactoryUrl)
+        project.extra.set("artifactoryUsername", artifactoryUsername)
+        project.extra.set("artifactoryPassword", artifactoryPassword)
+    }
```
gunt0001, line 64: @akh13771 - this still doesn't work when building the project from Android Studio and without any `local.properties` configured. The individual project modules apply the [artefact-deploy plugin](https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/blob/main/buildSrc/src/main/kotlin/deploy/ArtifactPublisher.kt#L46) which requires `artifactoryUrl` to be set. If not set, the plugin will throw an error. The best I could come up with so far is the following, which falls back to use `https://esri.jfrog.io/artifactory/arcgis` if `artifactoryUrl` could not be configured from commandline or local properties. We might want to centralize "https://esri.jfrog.io/artifactory/arcgis", since it's duplicated in `settings.gradle.kts`:

```suggestion
    // Find these in properties passed through command line or read from local.properties and
    // set them as project properties
    var artifactoryUrl: String? = project.findProperty("artifactoryUrl") as String?
        ?: localProperties.getProperty("artifactoryUrl")
        ?: ""
    val artifactoryUsername: String = project.findProperty("artifactoryUsername") as String?
        ?: localProperties.getProperty("artifactoryUsername")
        ?: ""
    val artifactoryPassword: String = project.findProperty("artifactoryPassword") as String?
        ?: localProperties.getProperty("artifactoryPassword")
        ?: ""

    if (artifactoryUrl.isNullOrEmpty()) {
        artifactoryUrl = "https://esri.jfrog.io/artifactory/arcgis"
    }
    project.extra.set("artifactoryUrl", artifactoryUrl)
    project.extra.set("artifactoryUsername", artifactoryUsername)
    project.extra.set("artifactoryPassword", artifactoryPassword)
```

akh13771: @gunt0001 Thanks for testing this, I forgot that the `artifactoryUrl` property was needed by other modules in this project. I went ahead and made the logic changes you suggested so that it will set `artifactoryUrl` to the public jfrog artifactory. Tested this and validated that if artifactoryUrl is not set in local.properties or the command line, it will default to the public instance. I did notice with this logic I implemented, if an empty string was passed in through either local.properties or the command-line, then it sets `artifactoryUrl` as an empty string. If that's fine, we can use the logic I implemented, but if not then the logic you have above will suffice so that even if `artifactoryUrl` is mistakenly set as an empty string, then it use the if block to set it accordingly. Let me know what you think

gunt0001 (approved):
@akh13771 - looks good 👍 

> I did notice with this logic I implemented, if an empty string was passed in through either local.properties or the command-line, then it sets artifactoryUrl as an empty string. If that's fine, we can use the logic I implemented

I think that's fine as is. I would say it's a user error if an empty string was passed.

Chad-Sargent (approved):
LGTM

skoppula018 (approved):
Looks good to me

StevenCoburn (approved):

settings.gradle.kts
```diff
@@ -65,10 +84,14 @@ dependencyResolutionManagement {
                 "https://esri.jfrog.io/artifactory/arcgis"
             )
         }
-        maven {
-            url = java.net.URI(
-                "https://olympus.esri.com/artifactory/arcgisruntime-repo/"
-            )
+        if (artifactoryUrl != "") {
```
chri7325, line 87: It is probably better practice to use ! isBlank() since Kotlin provides that. I see it used below as well.

chri7325 (approved):
Minor suggestion, looks good

settings.gradle.kts
Chad-Sargent: @chri7325  I'm not sure how to best add this change this since this is a merge from a fork. Do you know?

chri7325: @Chad-Sargent you'd need write access to @akh13771 fork. I didn't realize this was a fork. We can probably push through with this merge and have a follow up if necessary directly in the repo so anyone can write to it.

Chad-Sargent: OK sounds good

PR #832, opened 2025-04-17
WorldScaleSceneView: Add hit-test to SingleTapConfirmed

hud10837:
Related to issue: https://devtopia.esri.com/runtime/kotlin/issues/5197

### Description:

Adds the ability to get a map point when tapping on the screen based on a hit test rather than `screenToBaseSurface`. This is done by performing a hit test on the latest frame object form ARCore. In order to ensure it happens in `onSingleTapConfirmed`, we wait for that event to fire from the underlying scene view, hold onto it until the next frame, then perform a hit test and convert the hit result to a map point before firing the `onSingleTapConfirmed` callback on the `WorldScaleSceneView`. Note the hit result may still be null.

Previously the `SingleTapConfirmedEvent` would always have a null `mapPoint` property, because the SceneView always passes null for this.

### Summary of changes:

- Change the microapp to use the `SingleTapConfirmedEvent.mapPoint` instead of calculating the screen to base surface in order to place graphics
- Enhance the `ArSessionWrapper` to accept a parameter for the `PlaneFindingMode`, so that `WorldScaleSceneView` can have a `PlaneFindingMode.HORIZONTAL_AND_VERTICAL` if it needs to do hit tests.
- Enhance `WorldScaleCameraController` with a `getPointFromPose` function. The calculation for converting an ARCore pose to a map point is different for each camera controller because the `WorldTrackingCameraController` bases it on the ARCore world origin, whereas `GeospatialTrackingCameraController` doesn't keep track of the ARCore world origin so it needs to use the Earth object to get a lat/long.
- Perform the buffering logic for `onSingleTapConfirmed` as described above

Please test outside on the ground and on vertical planes. For testing, it may be helpful to modify `WorldScaleCameraController` to pass `true` to the `ArCameraFeed` parameter called `visualizePlanes`. It may also be helpful to set the base surface opacity of the scene in the microapp to 0f.

### Pre-merge Checklist

  - [ ] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/617/
  - [ ] https://runtime-kotlin.esri.com/view/vTest/job/vtest/job/toolkit/665/
  - [ ] Yes
  - [ ] No


toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -244,14 +262,46 @@ public fun WorldScaleSceneView(
...
     )
 
+    var lastSingleTapConfirmedEvent: SingleTapConfirmedEvent? by remember { mutableStateOf(null) }
+    val rememberedSingleTapConfirmedEvent = rememberUpdatedState(lastSingleTapConfirmedEvent)
+
     Box(modifier = modifier) {
         ArCameraFeed(
             session = arSessionWrapper,
             onFrame = { frame, displayRotation, session ->
+                rememberedOnSingleTapConfirmed.value?.let { singleTapConfirmedCallback ->
```
colinanderson, line 279: As there are similarly named things here - `rememberedOnSingleTapConfirmed` and `rememberedSingleTapConfirmedEvent` I'd recommend adding comments explaining what they are for at the point the vals are decalared.

colinanderson (commented):
@hud10837 the code looks ok. Two comments to consider. I'll take another look when the testing setup instructions are available.

colinanderson (approved):
@hud10837 code looks ok and the micro app worked ok when I was at street level rather than up in the office.

microapps/ArWorldScaleApp/app/src/main/java/com/arcgismaps/toolkit/arworldscaleapp/screens/MainScreen.kt
```diff
@@ -83,8 +83,8 @@ import com.arcgismaps.mapping.symbology.Symbol
 import com.arcgismaps.mapping.symbology.SymbolStyle
```
gunt0001, line 83: I'm getting a warning in this file:

image

microapps/ArWorldScaleApp/app/src/main/java/com/arcgismaps/toolkit/arworldscaleapp/screens/MainScreen.kt
```diff
@@ -113,8 +113,15 @@ fun MainScreen() {
         }
     }
     var displayCalibrationView by remember { mutableStateOf(false) }
-    val graphicsOverlays = remember { listOf(GraphicsOverlay()) }
-    val proxy = remember { WorldScaleSceneViewProxy() }
+    // The graphics overlay should have a surface placement of absolute to ensure
+    // that graphics are placed correctly if the user taps on a wall
```
gunt0001, line 117:
```suggestion
    // that graphics are placed correctly in 3D space if the user taps on a real object in the camera feed, such as a wall or tree
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -127,7 +132,10 @@ import java.time.Instant
  * @param onScale lambda invoked when a user performs a pinch gesture on the WorldScaleSceneView.
  * @param onUp lambda invoked when the user removes all their pointers from the WorldScaleSceneView.
  * @param onDown lambda invoked when the user first presses on the WorldScaleSceneView.
- * @param onSingleTapConfirmed lambda invoked when the user taps once on the WorldScaleSceneView.
+ * @param onSingleTapConfirmed lambda invoked when the user taps once on the WorldScaleSceneView. The
+ * [SingleTapConfirmedEvent.mapPoint] passed to this lambda is calculated by performing a hit test against
+ * detected planes in the camera feed. If no planes are detected at the tapped point, the mapPoint
+ * will be null.
```
gunt0001, line 138: 👍

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -214,7 +222,13 @@ public fun WorldScaleSceneView(
     )
     // If PE data could not be configured, we can't position the scene camera accurately
     if (!pedataConfigured) return
-
+    val planeFindingMode = remember(onSingleTapConfirmed) {
+        if (onSingleTapConfirmed != null) {
+            PlaneFindingMode.HORIZONTAL_AND_VERTICAL
+        } else {
+            PlaneFindingMode.DISABLED
```
gunt0001, line 229: fyi - I tested this case and it works 👍

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -244,14 +259,46 @@ public fun WorldScaleSceneView(
...
+                        if (hit == null) {
+                            userSingleTapConfirmedCallback(singleTapConfirmedEvent)
+                        }
+                        else {
+                            // hit is not null, so we can get the map point
+                            // first get the point relative to camera
+                            val mapPoint =
+                                worldScaleCameraController.getPointFromPose(hit.hitPose, session)
+                            userSingleTapConfirmedCallback(singleTapConfirmedEvent.copy(mapPoint = mapPoint))
+                        }
```
gunt0001, line 298: formatting- move `else` one line up, just after the closing bracket of the `if` statement:
```suggestion
                        if (hit == null) {
                            userSingleTapConfirmedCallback(singleTapConfirmedEvent)
                        } else {
                            // hit is not null, so we can get the map point
                            // first get the point relative to camera
                            val mapPoint =
                                worldScaleCameraController.getPointFromPose(hit.hitPose, session)
                            userSingleTapConfirmedCallback(singleTapConfirmedEvent.copy(mapPoint = mapPoint))
                        }
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/WorldScaleSceneView.kt
```diff
@@ -244,14 +259,46 @@ public fun WorldScaleSceneView(
...
+                    }
+                    lastSingleTapConfirmedEvent = null
+                }
```
gunt0001, line 301: I wonder if we should factor all this logic above out into a separate function, something like `handleSingleTapConfirmed()`

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/GeospatialTrackingCameraController.kt
```diff
@@ -163,6 +163,16 @@ internal class GeospatialTrackingCameraController(
         }
     }
 
+    override fun getPointFromPose(pose: Pose, session: Session): Point? {
+        return try {
```
gunt0001, line 167: Could make this an expression...
```suggestion
    override fun getPointFromPose(pose: Pose, session: Session): Point? = try {
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/WorldTrackingCameraController.kt
```diff
@@ -96,6 +97,17 @@ internal class WorldTrackingCameraController(
...
 
+    /**
+     * Converts an ARCore [Pose] to a global [Point] using the Pose's offset from the origin camera.
+     *
+     * @since 200.8.0
+     */
+    override fun getPointFromPose(pose: Pose, session: Session): Point? {
+        val hitPoseTransformationMatrix = pose.transformationMatrix
+        val origin = cameraController.originCamera.value.transformationMatrix
+        return GeometryEngine.projectOrNull(Camera(origin + hitPoseTransformationMatrix).location, WorldScaleParameters.SR_CAMERA)
```
gunt0001, line 108: Is the projection really needed here? Calculating a location from a `Camera` object should already give you the correct special reference for that location.

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/GeospatialTrackingCameraController.kt
gunt0001: How about:
```kotlin
override fun getPointFromPose(pose: Pose, session: Session): Point? = try {
            session.earth?.getGeospatialPose(pose)?.let { geospatialPose ->
                Point(geospatialPose.longitude, geospatialPose.latitude, geospatialPose.altitude, WorldScaleParameters.SR_WGS84_WGS_VERTICAL).let {
                    GeometryEngine.projectOrNull(it, WorldScaleParameters.SR_CAMERA)
                }
            }
        } catch (e: NotTrackingException) {
            null
        }
```

hud10837: I made this change 👍

gunt0001 (approved):
@hud10837 - looks good 👍

microapps/ArWorldScaleApp/app/src/main/java/com/arcgismaps/toolkit/arworldscaleapp/screens/MainScreen.kt
```diff
@@ -239,16 +257,18 @@ fun MainScreen() {
...
                     onSingleTapConfirmed = { singleTapConfirmedEvent ->
-                        proxy.screenToBaseSurface(singleTapConfirmedEvent.screenCoordinate)
+                        singleTapConfirmedEvent.mapPoint
                             ?.let { point ->
-                                graphicsOverlays.first().graphics.add(
-                                    Graphic(
-                                        point,
-                                        treeSymbol.value
+                                graphicsOverlays.first().graphics.apply {
+                                    clear()
```
colinanderson, line 264: Instead of clearing and creating a new graphic could you just update the geometry of the graphic? I suppose then it would be necessary to handle the case where there was no pre-existing graphic.

colinanderson (approved):
@hud10837 looks ok. I added one comment but probably not worth making any change.

microapps/ArWorldScaleApp/app/src/main/java/com/arcgismaps/toolkit/arworldscaleapp/screens/MainScreen.kt
hud10837: I was going to write up an argument not to change this, but it was quicker to change the code 😅

microapps/ArWorldScaleApp/app/src/main/java/com/arcgismaps/toolkit/arworldscaleapp/screens/MainScreen.kt
```diff
@@ -77,25 +75,35 @@ import com.arcgismaps.mapping.Basemap
...
 private const val KEY_PREF_ACCEPTED_PRIVACY_INFO = "ACCEPTED_PRIVACY_INFO"
 
+val graphicSymbol = SimpleMarkerSceneSymbol(
+    style = SimpleMarkerSceneSymbolStyle.Cone,
+    color = Color.red,
+    height = 0.5,
+    width = 0.5,
+    depth = 0.5,
+    anchorPosition = SceneSymbolAnchorPosition.Bottom
+).apply {
```
gunt0001, line 99: I propose making the cone slightly transparent and a little smaller. Also renaming the variable to `coneSymbol` which is more descriptive:
```suggestion
val coneSymbol = SimpleMarkerSceneSymbol(
    style = SimpleMarkerSceneSymbolStyle.Cone,
    color = Color.fromRgba(255, 0, 0, 70),
    height = 0.4,
    width = 0.2,
    depth = 0.2,
    anchorPosition = SceneSymbolAnchorPosition.Bottom
).apply {
```

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/ArSessionWrapper.kt
```diff
@@ -134,8 +137,8 @@ internal fun rememberArSessionWrapper(
             useGeospatial
         )
     }
-    LaunchedEffect(useGeospatial) {
-        arSessionWrapper.resetSession(useGeospatial)
+    LaunchedEffect(useGeospatial, planeFindingMode) {
```
gunt0001, line 140: Would this trigger two recompositions of `LaunchedEffect` if `useGeospatial` and `planeFindingMode` change at the same time?

gunt0001 (approved):
@hud10837 - looks good. Some comments...

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/ArSessionWrapper.kt
hud10837: I wasn't sure of the answer, so I built an app with this code:

```kotlin
var key1 by remember { mutableIntStateOf(0)}
var key2 by remember { mutableIntStateOf(0)}
Column(
    modifier = Modifier.padding(innerPadding)
) {
    Button(
        onClick = { key1++ },
    ) {
        Text(text = "Key 1: $key1")
       }
    Button(
        onClick = { key2++ },
    ) {
        Text(text = "Key 2: $key2")
    }
    Button(
        onClick = {
            key1++
            key2++
        },
    ) {
        Text(text = "Set both")
    }
}

LaunchedEffect(key1, key2) {
    Log.e("MainActivity", "Relaunch with Key1: $key1, Key2: $key2")
}
```

If you set both keys at the same time, the `LaunchedEffect` only runs once.

PR #842, opened 2025-05-05
Alan/basemap gallery 3d basemaps implementation

alan-edi:
Related to issue: https://devtopia.esri.com/runtime/kotlin/issues/5694

### Description:

Implements support for 3D basemaps in the BasemapGallery.

### Summary of changes:

- In the basemap gallery component:
   - In BasemapGalleryItem:
      - New property added: `is3D: Boolean`.
      - New public `constructor(item: Item, is3D: Boolean = false)` added.
      - Public `constructor(item: Item)` deprecated with `DeprecationLevel.HIDDEN` because it's superseded by the one mentioned above.
   - In BasemapGallery:
      - In the Composable BasemapGalleryItem function, put the thumbnail Image inside a Box. Also inside the Box, if `basemapGalleryItem.is3D`, add a Badge containing the text "3D". This is displayed on top of the top left corner of the thumbnail. Note this uses Box rather than BadgedBox for reasons elaborated in the [Design/Prototype issue](https://devtopia.esri.com/runtime/kotlin/issues/5394#issuecomment-5471667).
      - In the BasemapGalleryPreview, flag the first few BasemapGalleryItems as 3D.
   - Having done that I found `apiDump` added a public singleton `ComposableSingletons$BasemapGalleryKt`, so I removed that by adding an `apiValidation` block in `build.gradle.kts`. I copied the technique for this from the [Popup example](https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/blob/01ce496b7aee7e2313d292984cfaf7b8f31f2f00/toolkit/popup/build.gradle.kts#L74) that is linked in the [Toolkit Contribution Guidelines](https://devtopia.esri.com/runtime/kotlin/blob/main/doc/toolkit/contribution_guidelines.md#maintaining-binary-compatibility).
   - Having done that, ran `apiDump` again. In the resultant `basemapgallery.api` file:
      - The constructor that takes only an Item is marked as `synthetic` because it's now `HIDDEN`.
      - The new constructor that takes an Item and a Boolean has been added.
      - A getter function `is3D()` has been added.
   - In the basemap gallery README, just add a sentence saying how the example code could be modified to use SceneView/ArcGISScene.
- In the micro app:
   - Refactor micro app to use SceneView in order to showcase 3D basemaps alongside 2D basemaps.
   - In `ViewModel`, on successful loading of the Portal:
      - First call `portal.fetch3DBasemaps()` and add any resultant basemaps to the list of BasemapGalleryItems. Flag these as 3D basemaps if they contain an ArcGISSceneLayer.
      - Then call `portal.fetchDeveloperBasemaps()` and add any resultant basemaps to the list of BasemapGalleryItems. Pass `is3D = false` for these ones. Note the old code used a different method, `portal.fetchBasemaps()`. The rationale for using `fetchDeveloperBasemaps` is given in the [Design/Prototype issue](https://devtopia.esri.com/runtime/kotlin/issues/5394#issuecomment-5471419).
   - In the micro app README, change map to scene and add a sentence about 3D basemaps being highlighted using a "3D" badge. Also update the screenshot to show this.

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/644/
  - [ ] Yes
  - [ ] No


toolkit/basemapgallery/src/main/java/com/arcgismaps/toolkit/basemapgallery/BasemapGallery.kt
```diff
@@ -87,13 +88,20 @@ internal fun BasemapGalleryItem(
...
+        Box {
+            Image(
+                painter = thumbnail.value,
+                contentDescription = basemapGalleryItem.title,
+                modifier = Modifier
+                    .padding(8.dp)
+                    .clip(RoundedCornerShape(8.dp))
+            )
+            if (basemapGalleryItem.is3D) {
+                Badge {
```
colinanderson, line 100: I'd add this alignment so that the badge was on the top right as I think that is the more common location for badges.
```
Badge(modifier = Modifier.align(Alignment.TopEnd))
```
image
E.g. from a Mac the email badge is like this
image

microapps/BasemapGalleryApp/app/src/main/java/com/arcgismaps/toolkit/basemapgalleryapp/ViewModel.kt
```diff
@@ -65,13 +66,31 @@ class ViewModel(application: Application) : AndroidViewModel(application) {
             portal.load()
```
colinanderson, line 66: I'm seeing that when the app starts the gallery is sometimes empty for a noticeable period. A solution might be to load the styles first rather than last since that is the gallery content that is seen first. I wonder if one of the new portal queries is a lot slower than what was being used before?

Maybe it would be more responsive if basemap styles and portal items fetching were launched separately.

colinanderson: I wonder if the delay is being caused by having to load the basemaps to check if 3D basemaps are really 3D.... is that an necessary step?

colinanderson (commented):
@alan-edi looks ok but I'm seeing slow performance of the gallery items appearing.

microapps/BasemapGalleryApp/app/src/main/java/com/arcgismaps/toolkit/basemapgalleryapp/ViewModel.kt
alan-edi: 1. Loading the styles first makes sense and seems to make a significant difference, so changed to do that.
2. Re loading the basemaps etc, both swift & dotnet choose to do that (although Qt doesn't). I feel inclined to follow the majority unless we feel a significant performance penalty persists.

microapps/BasemapGalleryApp/app/src/main/java/com/arcgismaps/toolkit/basemapgalleryapp/ViewModel.kt
```diff
@@ -65,13 +66,31 @@ class ViewModel(application: Application) : AndroidViewModel(application) {
...
+                        .onSuccess {
+                            it.forEach { basemap ->
+                                basemap.load()
+                                    .onFailure { Log.w("BasemapGallery", "Failed to load basemap") }
+                                    .onSuccess {
+                                        basemap.item?.let { item ->
+                                            val sceneLayers =
+                                                basemap.baseLayers.filterIsInstance<ArcGISSceneLayer>()
+                                            val galleryItem =
+                                                BasemapGalleryItem(item, sceneLayers.size > 0)
```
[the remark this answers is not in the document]
alan-edi: > Or you might be able to simplify to something like

That doesn't really look simpler to my eye.

alan-edi (commented):
@colinanderson I've changed:
- the order of loading
- the badge alignment
- the `isNotEmpty` simplification

Please have another look. In particular, what do you think of the performance now?

alan-edi:
@colinanderson Further changes, all in ViewModel, complete now:
- Don't load the `Portal` because that's done under the covers when basemaps fetched.
- Don't check that 3D basemaps have a ArcGISSceneLayer - just flag all basemaps returned by `fetch3DBasemaps` as 3D.
- Perform each operation that does network requests asynchronously.

microapps/BasemapGalleryApp/app/src/main/java/com/arcgismaps/toolkit/basemapgalleryapp/ViewModel.kt
```diff
@@ -60,44 +61,60 @@ class ViewModel(application: Application) : AndroidViewModel(application) {
...
-            service.load()
-                .onFailure { Log.w("BasemapGallery", "Failed to load basemap styles") }
+            portal2DBasemapsJob.await()
+                .onFailure { Log.w("BasemapGallery", "Failed to fetch 2D basemaps") }
                 .onSuccess {
-                    // for each basemap style info create a gallery item and add it to the list of items
-                    service.info?.stylesInfo?.forEach { basemapStyleInfo ->
-                        val galleryItem = BasemapGalleryItem(basemapStyleInfo)
-                        styleItems.add(galleryItem)
+                    it.forEach { basemap ->
```
colinanderson, line 102: Can't say if it is better but you can do this by using `map` functions
```kotlin
                            portalItems.addAll(
                                it.mapNotNull { basemap -> basemap.item }
                                    .map { item -> BasemapGalleryItem(item, false) })
```
similarly for the basemap styles
```kotlin
                    styleItems.addAll(
                        service.info?.stylesInfo.orEmpty().map {
                            BasemapGalleryItem(it)
                        })
```

Probably just more interesting in seeing how we can handle `null` in Kotlin.

colinanderson (approved):
@alan-edi looks ok to me. I'll leave it up to you if you want to make changes or not based on the comments.

microapps/BasemapGalleryApp/app/src/main/java/com/arcgismaps/toolkit/basemapgalleryapp/ViewModel.kt
```diff
@@ -60,44 +61,60 @@ class ViewModel(application: Application) : AndroidViewModel(application) {
 
     init {
         viewModelScope.launch {
-            // get basemap portal items from a portal
+            // get basemap style info from a basemap style service
+            async {
```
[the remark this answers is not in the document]
alan-edi: Agreed, launch is better

microapps/BasemapGalleryApp/app/src/main/java/com/arcgismaps/toolkit/basemapgalleryapp/ViewModel.kt
alan-edi: Yes, interesting to compare the ways of handling null, but I'm not seeing it as an improvement so leaving as is

alan-edi (commented):
Thanks @colinanderson

microapps/BasemapGalleryApp/app/src/main/java/com/arcgismaps/toolkit/basemapgalleryapp/ViewModel.kt
```diff
@@ -60,44 +61,60 @@ class ViewModel(application: Application) : AndroidViewModel(application) {
...
-                                }
-                            }
+            val portal3DBasemapsJob = async {
+                portal.fetch3DBasemaps()
+            }
+
+            // fetch 2D basemaps from the portal
+            val portal2DBasemapsJob = async {
+                portal.fetchDeveloperBasemaps()
+            }
```
hud10837, line 86: Since you go on to immediately call `await()` sequentially on each of these, I'm not sure that you get much benefit from using `async`. You could just as easily use `launch` for each of these and process the result within the `launch` block.

I think you could also make this more functional style to simplify:

```kotlin
val basemapGalleryItems = portal
                        .fetch3DBasemaps()
                        .getOrElse { TODO() }
                        .mapNotNull { basemap ->  basemap.item }
                        .map { item -> BasemapGalleryItem(item, true) }
portalItems.addAll(basemapGalleryItems)
```

toolkit/basemapgallery/src/main/java/com/arcgismaps/toolkit/basemapgallery/BasemapGalleryItem.kt
```diff
@@ -36,11 +37,13 @@ import com.arcgismaps.portal.LoadableImage
 public class BasemapGalleryItem(
     public val title: String,
     public val tag: Any? = null,
+    public val is3D: Boolean,
```
hud10837, line 40: I think generally we shouldn't add a non-defaulted parameter after a defaulted one. Would it be okay to default this to `false`?

hud10837 (commented):
Looks good, just a few thoughts

toolkit/basemapgallery/src/main/java/com/arcgismaps/toolkit/basemapgallery/BasemapGalleryItem.kt
alan-edi: Good point, defaulting it to `false`

microapps/BasemapGalleryApp/app/src/main/java/com/arcgismaps/toolkit/basemapgalleryapp/ViewModel.kt
alan-edi: I need to display all the 3D basemaps at the top of the gallery, so all of them need added to `portalItems` before adding any of the 2D ones. Using `await` seems the most convenient way to do that.

However, the functional style you show seems to be an improvement, so I've changed to that.

alan-edi (commented):
@hud10837 I've made a couple of changes. Please have another look

hud10837 (approved):
LGTM :+1:

alan-edi:
Thanks @hud10837. New vtest run: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/679/

PR #853, opened 2025-05-21
Hud10837/update tabletop microapp data

hud10837:
Related to issue: https://devtopia.esri.com/runtime/kotlin/issues/5628

### Description:

The existing data used for the tabletop microapp floats slightly above the surface, even in ArcGIS online. This PR changes the data to an integrated mesh scene which performs decently on old hardware, given the clipping distance.

### Summary of changes:

- Change the scene layer in the microapp
- Change the anchor point, clipping distance, and translation factor to match
- Set the "keep screen on" flag on the microapp -- this was bothering me while testing, because the screen kept turning off when not interacting with the app.

### Pre-merge Checklist

  - [ ] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/684/
  - [ ] Yes
  - [x] No

### DIFF
--- a/microapps/ArTabletopApp/app/src/main/java/com/arcgismaps/toolkit/artabletopapp/MainActivity.kt
+++ b/microapps/ArTabletopApp/app/src/main/java/com/arcgismaps/toolkit/artabletopapp/MainActivity.kt
@@ -20,6 +20,7 @@ package com.arcgismaps.toolkit.artabletopapp
 
 import android.os.Bundle
 import android.util.Log
+import android.view.WindowManager
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.compose.foundation.layout.Box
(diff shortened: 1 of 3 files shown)


alan-edi (approved):
@hud10837 The code looks fine. I tried running the app to see if the scene renders on the surface, but I'm unable to verify that. My phone, a Motorola edge 40 neo, has always struggled to find stable planes using this app, and that makes it impossible to tell if the scene is placed accurately :(

microapps/ArTabletopApp/app/src/main/java/com/arcgismaps/toolkit/artabletopapp/MainActivity.kt
```diff
@@ -60,6 +61,7 @@ class MainActivity : ComponentActivity() {
                 }
             }
         }
+        window.addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
```
gunt0001, line 64: We should probably add this to the WorldScale app as well. Could be done as part of this PR.

gunt0001 (approved):
@hud10837 - looks good in my local testing 👍 One minor comment...

PR #856, opened 2025-05-22
Add fix so that we always run the workManagerRepository funs in viewModel scope

puneet-pdx:
Related to issue: # https://devtopia.esri.com/runtime/kotlin/issues/5878

### Description:

Currently on orientation changes the OfflineMapAreas composable's coroutine scope was getting cancelled, leading to a cancellation exception that was bubbling up as downloadFailed error. The cancellation of the coroutine scope was causing the collect call `workManager.getWorkInfoByIdFlow` to get removed as well. Leading to the Status of the download not getting updated when the backgrounded download finished.
To fix that we now rely on the user to provide us with the viewmodelScope in the OfflineMapAreasState. So we always run the workManagerRepository funs in the viewModel scope, and the long running collect operations survive orientation changes.

### Summary of changes:

- Add `CoroutineScope` parameter to the OfflineMapState to hold on to the parent viewModelScope
- Move call to download preplannedMapArea from OfflineMapArea to PrePlannedMapAreas

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/job/vtest/job/toolkit/696/
  - [ ] Yes
  - [ ] No

### DIFF
--- a/toolkit/offline/src/main/java/com/arcgismaps/toolkit/offline/OfflineMapState.kt
+++ b/toolkit/offline/src/main/java/com/arcgismaps/toolkit/offline/OfflineMapState.kt
@@ -52,8 +52,7 @@ internal const val notificationChannelDescription =
  */
 @Stable
 public class OfflineMapState(
-    private val arcGISMap: ArcGISMap,
-    private val viewModelScope: CoroutineScope
+    private val arcGISMap: ArcGISMap
 ) {
     private lateinit var _workManagerRepository: WorkManagerRepository
     private var _mode: OfflineMapMode = OfflineMapMode.Unknown
@@ -113,8 +112,7 @@ public class OfflineMapState(
                         preplannedMapArea = it,
                         offlineMapTask = offlineMapTask,
                         portalItemId = portalItemId,
-                        workManagerRepository = _workManagerRepository,
-                        scope = viewModelScope
+                        workManagerRepository = _workManagerRepository
                     )
                     preplannedMapAreaState.initialize()
                     _preplannedMapAreaStates.add(preplannedMapAreaState)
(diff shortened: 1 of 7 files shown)


sorenoid (changes requested):
nice catch. Some suggestions.

toolkit/offline/src/main/java/com/arcgismaps/toolkit/offline/OfflineMapState.kt
```diff
@@ -51,7 +52,8 @@ internal const val notificationChannelDescription =
  */
 @Stable
 public class OfflineMapState(
-    private val arcGISMap: ArcGISMap
+    private val arcGISMap: ArcGISMap,
+    private val viewModelScope: CoroutineScope
```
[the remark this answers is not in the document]
puneet-pdx: I was thinking it might be easier if we could tie the async work with the Lifetime of the ViewModel that the user will be using. But I see your point about using the name to signify this to the user. 

>Why not just create a coroutine using a Launcher and Dispatchers.IO at the point of launching the worker?

I was concerned about having the coroutine outliving the PreplannedMapAreaState object. I have pushed the change where now I am creating the coroutine with Dispatchers.IO and later cancelling them.

toolkit/offline/src/main/java/com/arcgismaps/toolkit/offline/preplanned/PreplannedMapAreasState.kt
```diff
@@ -87,27 +89,35 @@ internal class PreplannedMapAreaState(
      *
      * @since 200.8.0
      */
-    internal fun downloadPreplannedMapArea() {
+    internal fun downloadPreplannedMapArea() = runCatchingCancellable {
```
shubham7109, line 92: Maybe the job failure be captured in `PreplannedMapAreas` as well?  
```kotlin
if (state.status.allowsDownload) {
    state.downloadPreplannedMapArea().onFailure {
        state.updateStatus(Status.DownloadFailure(it))
        state.disposeScope()
    }
}
```

shubham7109 (approved):
Looks good, download status does persist, just one thought on the scope but seems like the coroutine should be contained in `PreplannedMapAreaState` 👍 .

toolkit/offline/src/main/java/com/arcgismaps/toolkit/offline/preplanned/PreplannedMapAreasState.kt
puneet-pdx: I am thinking catching the exception might not be needed here at all. Let's consider that refactor in this PR where I have cancellation in place ?
https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/pull/859

PR #901, opened 2025-06-23
l10n staging 200.8.0

sorenoid:
Related to issue: https://devtopia.esri.com/runtime/kotlin/issues/5920

### Description:

localization and cleanup of res folder.

### Summary of changes:

- localization
- removal of values-he, values-id folders.
- removal of duplicated values folders inside values folders

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/vTest/job/vtest/job/toolkit/771/
  - [ ] Yes
  - [ ] No


TADraeseke (changes requested):
- I see all the id (indonesia) files have been deleted. Just checking that's expected.

- There also seems to be some nesting values-in/values-id

- values-iw/values-iw -> values-da is an odd diff that stood out to me

Just checking on a few of these things I noticed. Thanks!

sorenoid:
@TADraeseke 

> * I see all the id (indonesia) files have been deleted. Just checking that's expected.

Yes it was intended. `values-id` name was causing lint issues. Java needs translations to be in folders with deprecated locale codes, so while `id` is the up to date code, `in` is what Java needs and this causes a lint error for us.

Here is the lint output

```
Lint found 1 errors, 40 warnings. First failure:
  
  /Users/sor10874/git/github/arcgis-maps-sdk-kotlin-toolkit/toolkit/utilitynetworks/src/main/res/values-id: Error: The locale folder "id" should be called "in" instead; see the java.util.Locale documentation [LocaleFolder]
  
     Explanation for issues of type "LocaleFolder":
     From the java.util.Locale documentation:
     "Note that Java uses several deprecated two-letter codes. The Hebrew ("he")
     language code is rewritten as "iw", Indonesian ("id") as "in", and Yiddish
     ("yi") as "ji". This rewriting happens even if you construct your own
     Locale object, not just for instances returned by the various lookup
     methods.
  
     Because of this, if you add your localized resources in for example
     values-he they will not be used, since the system will look for values-iw
     instead.

```

What I found was that wherever we had a (non-nested) `values-id` there was also a `values-in` containing an identical strings file, so I removed the `values-id` folders.
 
> * There also seems to be some nesting values-in/values-id

I intended to remove all these, they are from a previous translation issue and are not in use. Did you see any that were not removed?

> * values-iw/values-iw -> values-da is an odd diff that stood out to me
> 
I don't see this. Do you mind commenting on the diff where it is?

> Just checking on a few of these things I noticed. Thanks!

shubham7109 (approved):
Seems like `ar/build.gradle.kts` & `authentication/build.gradle.kts` currently disables `MissingTranslation`, I'm assuming this currently due to ongoing effort for more translations.

Approving as changes looks good. 👍

PR #906, opened 2025-06-26
Alan/flyover poi

alan-edi:
Related to issue: https://devtopia.esri.com/runtime/kotlin/issues/6009

### Description:

In the Flyover micro app, adds a drop-down menu that presents a list of a few Points of Interest (POI), these being locations within the city to jump to.

### Summary of changes:

- Moves the Scaffold from MainActivity to MainScreen.
- Adds a data class PointOfInterest. Note this includes a `description` property that's intended for use if/when we add a Callout. The description texts will need to be fleshed out if we go ahead with a Callout, or the entire property be deleted if we drop the Callout.
- In MainScreen, adds a Scaffold that contains a TopBar that contains a DropDownMenu of points of interest. Selecting a POI causes the FlyoverSceneView to zoom to that place. Note the translationFactor is also set. Currently all POIs have a relatively small translationFactor, 10.0, except the "Flying over the city" POI which uses 1000.0.
- Also in MainScreen, reorders some code to hopefully make it read better now that quite a lot has been added. Also adds some comments. 

### Pre-merge Checklist

  - [ ] link:
  - [ ] Yes
  - [ ] No

### DIFF
--- a/microapps/ArFlyoverApp/app/src/main/java/com/arcgismaps/toolkit/arflyoverapp/MainActivity.kt
+++ b/microapps/ArFlyoverApp/app/src/main/java/com/arcgismaps/toolkit/arflyoverapp/MainActivity.kt
@@ -23,14 +23,9 @@ import android.util.Log
 import android.view.WindowManager
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
-import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.ExperimentalMaterial3Api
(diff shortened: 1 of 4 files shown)


alan-edi (commented):
@colinanderson Good suggestions, adopted, please have another look

alan-edi (commented):
@gunt0001 Thanks. I believe I've adopted all your suggestions. In the case of the rememberPointsOfInterest() function, please check it looks OK. It works but I'm not sure I've implemented it in the best way.

gunt0001 (approved):
@alan-edi - looks good, one remaining comment.

PR #909, opened 2025-06-28
`FeatureForm` : Provide navigation blocking 

kaushikrw:
Related to issue: #[apollo/1239](https://devtopia.esri.com/runtime/apollo/issues/1239)

### Description:

Provides a way for developers to block or disable navigation to another associated feature. This is important to support the workflows where geometry changes can be made outside the context of a feature form. In such cases, the app can disable navigation when there are geometry changes and re-enable it after the changes have been persisted. There might be other use cases as well for a developer or an app to block navigation.

This mechanism is the outcome of the research from this [issue](https://devtopia.esri.com/runtime/apollo/issues/1272).

### Summary of changes:

- Added a parameter `isNavigationEnabled : Boolean` which defaults to true. When it is true, all navigation between features is allowed. When false is set, the navigation is disabled.
- When the navigation is disabled, the list of associations is greyed out. "Back" navigation will also be disabled along with the "back arrow" icon greyed out if there is an associated feature to go back to.
- Added a menu option in the micro-app, in the map screen where this navigation can be toggled to enable/disable.

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/job/vtest/job/toolkit/788/
  - [x] Yes
  - [ ] No

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
@@ -302,9 +302,10 @@ public fun FeatureForm(
  * @param showFormActions Indicates if the form actions (save and discard buttons) should be displayed.
  * Default is true.
  * @param isNavigationEnabled Indicates if the navigation is enabled for the form when there are
- * [UtilityAssociationsFormElement]s present. If true, the user can navigate to associated features
+ * [UtilityAssociationsFormElement]s present. When true, the user can navigate to associated features
  * and back. If false, this navigation is disabled. If there are geometry edits on the feature,
  * this flag can be set to false to prevent navigation until the geometry edits are saved or discarded.
+ * Default is true.
  * @param onBarcodeButtonClick A callback that is invoked when the barcode accessory is clicked.
  * The callback is invoked with the [FieldFormElement] that has the barcode accessory. If null, the
  * default barcode scanner is used.


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/UtilityAssociations.kt
```diff
@@ -246,11 +257,12 @@ private fun AssociationItem(
                 Text(
                     text = title,
                     modifier = Modifier.padding(
-                        top = 6.dp
+                        top = if (trailingText.isNotEmpty()) 6.dp else 0.dp,
```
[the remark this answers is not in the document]
kaushikrw: Not related but I noticed the spacing was off if there is no trailing text.

sorenoid (approved):

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
```diff
@@ -301,6 +301,10 @@ public fun FeatureForm(
  * callback will be invoked when the close icon is clicked. Default is true.
  * @param showFormActions Indicates if the form actions (save and discard buttons) should be displayed.
  * Default is true.
+ * @param isNavigationEnabled Indicates if the navigation is enabled for the form when there are
```
shubham7109, line 304: Should the default is true be mentioned in the doc here (similar to `showFormActions`)? Not needed if this doc is already reviewed.

shubham7109 (approved):
Tested the changes here, and PR looks good. Just one clarification on the doc:

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureForm.kt
kaushikrw: Updated!

kaushikrw:
Thanks @sorenoid @shubham7109, merging.

PR #959, opened 2025-07-28
`FeatureForm`: Micro-app - Update how thumbnails are rendered

kaushikrw:
Related to issue: #[kotlin/6287](https://devtopia.esri.com/runtime/kotlin/issues/6287)

### Description:

This PR updates the mechanism of how the thumbnails for `PortalItem`s are loaded and displayed in the  `MapListScreen` of the micro-app. The new mechanism uses coil3 and its features to efficiently load and render a `LoadableImage`.

### Summary of changes:

- Introduced a new `LoadableImageFetcher` class which translates a `LoadableImage` into a `FetchResult` that coil3 can understand. 
    - An instance of a `LoadableImageFetcher` created from a `LoadableImage` will load the loadable and return the result as a `BitmapImage`.
    - A `Keyer` provides a unique key for each `LoadableImage` so that coil can cache the images and retrieve them from the cache using the key (which is the `url`).
- This allows thumbnails to be loaded in parallel, rather than alongside the portal items, which helps reduce the time required to fetch all portal items when the user logs in or refreshes the list.
- Removed the older implementation of the `AsyncImage`.
- This also fixes an issue in the micro-app where the thumbnails being rendered were mismatched. 

### Pre-merge Checklist

  - [ ] link:
  - [ ] Yes
  - [ ] No


sorenoid (approved):

TADraeseke (changes requested):
Hi Kaushik! Just one question. Thanks!

microapps/FeatureFormsApp/app/src/main/java/com/arcgismaps/toolkit/featureformsapp/utils/LoadableImageFetcher.kt
```diff
@@ -0,0 +1,73 @@
...
+    private val loadable: LoadableImage
+) : Fetcher {
+
+    /**
+     * Fetches the image from the [LoadableImage] source and returns it as an [ImageFetchResult].
+     *
+     * @return An [ImageFetchResult] containing the loaded image as a bitmap, or null if the image could not be loaded.
+     */
+    override suspend fun fetch(): ImageFetchResult? {
+        loadable.retryLoad()
```
TADraeseke, line 41: Why is this a retryLoad() instead of just load()

kaushikrw: That is due to the behavior of a `Loadable`. If the initial `load()` fails for whatever reason, any subsequent `load()` calls will fail immediately. So, using `retryLoad()` is essentially the same as doing this -

```
when(loadable.loadStatus) {
    is LoadStatus.NotLoaded -> load()
    is LoadStatus.FailedToLoad -> retryLoad()
    else -> {}
}
```

TADraeseke: Can it get caught in a loop if it repeatedly fails to load? Could a counter for failed attempts be helpful to break out?

kaushikrw: That's a valid point, but it can never enter a loop. The task is created and initiated when the thumbnail or the `PortalItem` comes into view. If it fails, it is re-created and retried only when it comes back into view.

TADraeseke: Ahh got it, thanks for the clarification!

TADraeseke (approved):
LGTM!

kaushikrw:
Thanks @TADraeseke and @sorenoid! Merging now!

PR #969, opened 2025-08-14
Introduce network mocking to the toolkit

kaushikrw:
Related to issue: #[kotlin/6416](https://devtopia.esri.com/runtime/kotlin/issues/6416)

### Description:

This PR introduces the framework for mocking network responses for any tests in the toolkit.

### Summary of changes:

- Added dependency on the `mockingjay` library which is used for mocking network data.
- Introduced a new gradle plugin `GrantTestPermission` which will grant the required permissions to the test app. This is used for granting elevated permissions such as `MANAGE_EXTERNAL_STORAGE` which is necessary for reading the mocked response files stored in the root of the storage on the test device.
- Enabled mocking for the `FeatureFormTests`.
- Fixed the failing group element tests.

Note : This is dependent on the companion sdk PR in the linked issue.

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/job/vtest/job/toolkit/994/
  - [x] Yes
  - [ ] No


gunt0001 (approved):
@kaushikrw - looks good. I suggest to merge this straight into `v.next` though. Once on `v.next`, @rom14514 can continue the work to consolidate the test utils in a shared library etc. At some point this would also be merged into v.next and if you are still working on this feature branch at that point, you can rebase.

Would that work?

toolkit/featureforms/src/androidTest/java/com/arcgismaps/toolkit/featureforms/BarcodeTests.kt
```diff
@@ -41,11 +39,6 @@ class BarcodeTests : FeatureFormTestRunner(
     @get:Rule
     val composeTestRule = createComposeRule()
 
-    // Grant camera permission for barcode scanning
-    @get:Rule
-    val runtimePermissionRule: GrantPermissionRule =
-        GrantPermissionRule.grant(Manifest.permission.CAMERA)
-
```
hud10837, line 48: If I am understanding this correctly, then the `GrantTestPermissions` plugin doesn't actually need to grant the camera permission before running the tests, because the test can request it. I have some questions about this though:

1. ~~Could we grant the storage permissions using `GrantPermissionRule`? If so, could we add that as a `BeforeAfterAction` once we have that available to the toolkit (soon)?~~ I investigated this and it's not possible
2. If we can't do the above, could we keep the Camera permission granted at this level? This would be helpful because it would mean we could share the grant permissions task from the SDK plugin without modifications
3. When we use `GrantPermissionRule`, we don't need to worry about the manifest? I'm guessing that is the case because I don't see it in the docs and you are just now adding the permissions to the manifest

kaushikrw:
> @kaushikrw - looks good. I suggest to merge this straight into `v.next` though. 
> 
> Would that work?

I may have branched off from the feature branch.. so this might also include changes from the feature branch. Is that okay? Alternatively, I can create another PR to merge the `feature-branches/forms` into `v.next` right after this is first merged into `feature-branches/forms`.

hud10837 (approved):
Looks good, approving so this could move forward but I do have questions to guide the follow-up work

toolkit/featureforms/src/androidTest/java/com/arcgismaps/toolkit/featureforms/BarcodeTests.kt
kaushikrw: > If we can't do the above, could we keep the Camera permission granted at this level? This would be helpful because it would mean we could share the grant permissions task from the SDK plugin without modifications.

Yeah, the camera permission can stay within the test. I just moved it out because the having all the permissions granted by the plugin made sense.

> When we use GrantPermissionRule, we don't need to worry about the manifest? I'm guessing that is the case because I don't see it in the docs and you are just now adding the permissions to the manifest

I believe so according to the [doc](https://developer.android.com/reference/androidx/test/rule/GrantPermissionRule). But the manage storage permission definitely need the manifest.

gunt0001:
> I may have branched off from the feature branch.. so this might also include changes from the feature branch. Is that okay? Alternatively, I can create another PR to merge the feature-branches/forms into v.next right after this is first merged into feature-branches/forms.

Ok, in that case, I think the options are:
1. Go ahead and merge this PR into your feature branch. We will make changes independently on `v.next` to introduce the shared test util lib and the shared gradle plugin for data syncing. In this case you will likely need to resolve conflicts once you merge your feature branch into v.next.
2. Do what you suggest above - `Alternatively, I can create another PR to merge the feature-branches/forms into v.next right after this is first merged into feature-branches/forms.`

Whatever works best for you...

kaushikrw:
Thanks @gunt0001 and @hud10837. I will merge this and follow up with a [PR](https://devtopia.esri.com/runtime/kotlin/issues/6733) into v.next.

PR #973, opened 2025-08-27
Pause ar session before closing it

rom14514:
Related to issue: #

### Description:

Pause ar session before closing it. This approash is recommended by [oficcial documentation](https://developers.google.com/ar/reference/java/com/google/ar/core/Session#close-).

### Summary of changes:

- add `session?.pause()` before `session?.close()`

### Pre-merge Checklist

  - [ ] link:
  - [ ] Yes
  - [ ] No

### DIFF
--- a/toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/ArSessionWrapper.kt
+++ b/toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/ArSessionWrapper.kt
@@ -30,6 +30,7 @@ import com.google.ar.core.Config
 import com.google.ar.core.Config.PlaneFindingMode
 import com.google.ar.core.Session
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.launch
 import java.util.concurrent.locks.Lock
 import java.util.concurrent.locks.ReentrantLock
@@ -54,12 +55,19 @@ internal class ArSessionWrapper(
     private var shouldInitializeDisplay = true
 
     override fun onDestroy(owner: LifecycleOwner) {
+        var sessionRef: Session? = null
         withLock { session, _ ->
             this.session = null
-            session?.pause()
-            owner.lifecycleScope.launch(Dispatchers.Default) {
-                session?.close()
-            }
+            sessionRef = session
+        }
+        // Call pause() before close(), even if running on the same thread.
+        // This ensures proper cleanup before closing the session.
+        sessionRef?.pause()
+        // Close the session on a background thread as recommended in the official docs:
+        // https://developers.google.com/ar/reference/java/com/google/ar/core/Session#close()
+        // Using NonCancellable so the close operation runs even if the coroutine scope is cancelled.
+        owner.lifecycleScope.launch(Dispatchers.Default + NonCancellable) {
+            sessionRef?.close()
         }
     }
 
--- a/toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/ArSessionWrapper.kt
+++ b/toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/ArSessionWrapper.kt
@@ -53,6 +53,7 @@ internal class ArSessionWrapper(
     override fun onDestroy(owner: LifecycleOwner) {
         withLock { session, _ ->
             this.session = null
+            session?.pause()
             session?.close()
         }
     }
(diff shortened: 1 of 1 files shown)


toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/ArSessionWrapper.kt
```diff
@@ -53,6 +53,7 @@ internal class ArSessionWrapper(
     override fun onDestroy(owner: LifecycleOwner) {
         withLock { session, _ ->
             this.session = null
+            session?.pause()
             session?.close()
```
hud10837: the official doc also says to do this on a background thread. Maybe we should launch this on `Dispatchers.Default` as well
```suggestion
            withContext(Dispatchers.Default) {
                session?.close()
            }
```

Edit: my bad, we are not on a coroutine context, so this will need further thinking...

hud10837 (approved):
LGTM 👍

toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/ArSessionWrapper.kt
```diff
@@ -53,7 +56,10 @@ internal class ArSessionWrapper(
     override fun onDestroy(owner: LifecycleOwner) {
         withLock { session, _ ->
             this.session = null
-            session?.close()
+            session?.pause()
+            owner.lifecycleScope.launch(Dispatchers.Default) {
+                session?.close()
+            }
```
sorenoid: I don't think it is the worst thing in the world to close the session on the main thread in `onDestroy`, it's being destroyed and it is off the screen. But to launch something inside a lock effectively removes it from the lock.

If `close()` needs to be done under lock, I would do it like below. Note the `NonCancellable` added to the coroutine context to ensure it runs before everything is torn down. Please check what is the essential critical section in the previous code. Was it just setting the session to null? Is `close()` now not critical because of the call to `pause()`? Only if `close()` is still critical should the code below be used. Otherwise, just move the launch out of the call to  withLock.

```kotlin
    override fun onDestroy(owner: LifecycleOwner) {
        lock.lock()
        val sessionRef = this.session
        this.session = null
        sessionRef?.pause()
        owner.lifecycleScope.launch(Dispatchers.Default + NonCancellable) {
            try {
                sessionRef?.close()
            } finally {
                lock.unlock()
            }
        }
    }

**Review by sorenoid — CHANGES_REQUESTED** [review]
I have some suggestions for the launched coroutine from within the locked critical section.

File: toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/ArSessionWrapper.kt (hunk 2 above)
Comment (rom14514) [reply toolkit/ar/src/main/java/com/arcgismaps/toolkit/ar/internal/ArSessionWrapper.kt]: This came up while researching a different issue.
While looking into that, I checked the docs for `Session.close()` and noticed it recommends calling it on a separate thread:

> This method will take several seconds to complete. To prevent blocking the main thread, call [pause()](https://developers.google.com/ar/reference/java/com/google/ar/core/Session#pause()) on the main thread, and then call [close()](https://developers.google.com/ar/reference/java/com/google/ar/core/Session#close()) on a background thread.

Given that `close()` can take several seconds, it seems better to run it off the non-main thread (even though most samples call it on the main thread, likely for simplicity). But I completely missed the `withLock` part. I believe we can call `session?.close()` outside of `withLock` like this:

```kotlin
    override fun onDestroy(owner: LifecycleOwner) {
        var sessionRef: Session? = null
        withLock { session, _ ->
            this.session = null
            sessionRef = session
        }
        sessionRef?.pause()
        owner.lifecycleScope.launch(Dispatchers.Default + NonCancellable) {
            sessionRef?.close()
        }
    }
```

Also, I am not sure if we need `NonCancellable` since `Session.close()` is non-suspending function.

sorenoid: Not sure about the `NonCancellable`. It may be necessary so the coroutine is scheduled at all.

sorenoid: otherwise LGTM. thanks

rom14514:
@sorenoid Thanks for the feedback! I kept `NonCancellable` as you suggested so the coroutine will start even if the scope is cancelled. The documentation for `NonCancellable` includes the following warning:

> WARNING: This object is not designed to be used with [launch](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/launch.html), [async](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/async.html), and other coroutine builders. if you write launch(NonCancellable) { ... } then not only the newly launched job will not be cancelled when the parent is cancelled, the whole parent-child relation between parent and child is severed. The parent will not wait for the child's completion, nor will be cancelled when the child crashed.

In our case, this behavior is intentional — we specifically want the coroutine to run even if the scope is cancelled, and we do not rely on any parent–child relationship — so I believe the warning can be safely ignored.

sorenoid (approved):

gunt0001:
Ad-hoc tested by destroying and creating a WorldScaleSceneView by adding/removing repeatedly to/from compostion.

PR #978, opened 2025-09-05
`FeatureForm` : Add delete UNA implementation 

kaushikrw:
Related to issue: #[apollo/1223](https://devtopia.esri.com/runtime/apollo/issues/1223)

### Description:

Updates the "delete association" user story implementation.

### Summary of changes:

- Copied in the m3 `ModalBottomSheet` as it is experimental.
- The `UtilityAssociationDetails` is no longer its own navigation destination but is presented in a `ModalBottomSheet`.
- Updated the impl of the `UtilityAssociationsElementState` to support deleting associations and providing a binding to the UI.
- Moved the call site of the `FeatureFormDialog` to the root level so that it does not need to be called for every screen.

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/919/ (1 test failing for unrelated reasons)
  - [ ] Yes
  - [ ] No


puneet-pdx:
@kaushikrw the swipe to remove gesture only works for the first time. If the user does swipe to delete gesture and then cancels the delete from the dialog, then the swipe to delete does not work for the subsequent time.


https://github.com/user-attachments/assets/aaba7bad-4261-40df-886a-d912a1640c33

puneet-pdx (commented):
@kaushikrw looks good overall. See my comment about swipe to delete not working as expected

kaushikrw:
Thanks @puneet-pdx.

> @kaushikrw the swipe to remove gesture only works for the first time. If the user does swipe to delete gesture and then cancels the delete from the dialog, then the swipe to delete does not work for the subsequent time.

Looking at this, it seems like it could either be a bug in m3 or intended behavior. I will try to get to the bottom of this.

puneet-pdx (approved):
@kaushikrw LGTM

PR #981, opened 2025-09-10
Add support for UNAssociationsPopupElement to Popup

puneet-pdx:
Related to issue: # https://devtopia.esri.com/runtime/apollo/issues/1407
 

### Description:

Add support for UNAssociations in Popups. 

### Summary of changes:

- Add `PopupState`, Add new Popup Composable that takes the State object
- Update all internal popup elements to not have remember functions and just initialize them using constructors and hold them in the PopupState object
- Add UtilityAssociationElement(from FeatureForms)
- Add support for Navigation to enable browsing from one Popup to another based on an association element(from FeatureForms)
- Add PopupScreen
- Add ContentAwareTopBar(from FeatureForms)
- Deprecate stateful PopupComposable
- Update MicroApp to use the new Composable

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/937/
  - [x] https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/946/
  - [ ] Yes
  - [ ] No


microapps/PopupApp/app/src/main/java/com/arcgismaps/toolkit/popupapp/screens/mapscreen/MapViewModel.kt
```diff
@@ -106,45 +87,8 @@ class MapViewModel(
...
-                val savedStateHandle = extras.createSavedStateHandle()
-
-                return MapViewModel(
-                    savedStateHandle,
-                    application
-                ) as T
-            }
-        }
+    fun updatePopupState(popup: Popup?) {
+        _popupState.value = PopupState(popup!!, viewModelScope)
```
kaushikrw: This needs a null check for `popup` before you apply the `!!` operator.

toolkit/popup/api/popup.api
```diff
@@ -1,4 +1,69 @@
...
+
+public final class com/arcgismaps/toolkit/popup/PopupState {
+	public static final field $stable I
+	public fun <init> (Lcom/arcgismaps/mapping/popup/Popup;)V
+	public fun <init> (Lcom/arcgismaps/mapping/popup/Popup;Lkotlinx/coroutines/CoroutineScope;)V
+	public final fun getActivePopup ()Lcom/arcgismaps/mapping/popup/Popup;
+	public final fun getPopup ()Lcom/arcgismaps/mapping/popup/Popup;
+}
+
+public final class com/arcgismaps/toolkit/popup/internal/element/fieldselement/FieldsElementState$Creator : android/os/Parcelable$Creator {
```
kaushikrw: There are many classes in this api dump that are internal and need to be added to the list of `ignoredClasses` in `build.gradle.kts`.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/Popup.kt
```diff
@@ -91,26 +65,63 @@ private data class PopupState(@Stable val popup: Popup)
...
+ * @param modifier The [Modifier] to be applied to layout corresponding to the content of this
+ * Popup.
+ * @param onDismiss Callback that is invoked when the user clicks the close icon in the top app bar.
+ * @param showCloseIcon Flag to indicate if the close icon should be shown in the top app bar. If true, the [onDismiss]
+ * callback will be invoked when the close icon is clicked. Default is true.
+ * @param isNavigationEnabled Indicates if the navigation is enabled for the popup when there are
+ * [UtilityAssociationsPopupElement]s present. When true, the user can navigate to associated features
+ * and back. If false, this navigation is disabled. Default is true
+ *
+ * @since 200.9.0
```
kaushikrw:
```suggestion
 * @since 300.0.0
```
I don't think there is a 200.9

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/Popup.kt
```diff
@@ -91,26 +65,63 @@ private data class PopupState(@Stable val popup: Popup)
  *
  * @since 200.5.0
  */
+@Deprecated(
+    message = "Maintained for binary compatibility. Use the overload that uses the PopupState object.",
+    level = DeprecationLevel.HIDDEN
```
kaushikrw: Any particular reason this is hidden? Imo this would be confusing to anyone who upgrades to v300 and is not able to see this function. Not sure what the user story for Popup is but imo having this as a warning with later removal or a full removal in 300.0 might be a better approach.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/Popup.kt
```diff
@@ -91,26 +65,63 @@ private data class PopupState(@Stable val popup: Popup)
...
+            scope,
+            // Ignore the UtilityAssociationsFormElement as it is not supported with this API
+            ignoreList = setOf(
+                UtilityAssociationsPopupElement::class.java
+            )
+        )
+    }
+
+    if (stateData.getActivePopupStateData().initialEvaluation.value) {
+        Popup(stateData, modifier, showCloseIcon = false)
```
kaushikrw: It would be nice to have a progress/loading indicator here until the expressions are evaluated. A simple jump without a transition might feel jarring.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/Popup.kt
```diff
@@ -91,26 +65,63 @@ private data class PopupState(@Stable val popup: Popup)
...
-private val attachments: MutableList<PopupAttachment> = mutableListOf()
-
 @Composable
-private fun Popup(popupState: PopupState, modifier: Modifier = Modifier) {
+public fun Popup(
+    popupState: PopupState,
+    modifier: Modifier = Modifier,
+    onDismiss: () -> Unit = {},
+    showCloseIcon: Boolean = true,
+    isNavigationEnabled : Boolean = true
```
kaushikrw:
```suggestion
    isNavigationEnabled : Boolean = true
```
For `FeatureForm`, this was meant as a way for a developer to disable the navigation in case there are geometry edits on the feature (since a form isn't concerned with geometry). Is there a similar need/requirement for a developer to disable navigation for popups?

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/PopupState.kt
```diff
@@ -0,0 +1,358 @@
...
+import com.arcgismaps.toolkit.popup.internal.element.state.mutablePopupElementStateCollection
+import com.arcgismaps.toolkit.popup.internal.element.textelement.TextElementState
+import com.arcgismaps.toolkit.popup.internal.element.utilityassociationselement.UtilityAssociationsElementState
+import com.arcgismaps.toolkit.popup.internal.navigation.NavigationRoute
+import com.arcgismaps.toolkit.popup.internal.navigation.lifecycleIsResumed
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart
+import kotlinx.coroutines.launch
+
+public class PopupState(@Stable public val popup: Popup) {
```
kaushikrw: Doc is missing for the `PopupState` class.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/PopupState.kt
```diff
@@ -0,0 +1,358 @@
...
+internal data class PopupStateData(
+    val popup: Popup,
+) {
+    internal lateinit var stateCollection: PopupElementStateCollection
+        private set
+    /**
+     * Indicates if the evaluateExpression function for the [popup] has been run.
+     */
+    internal var initialEvaluation : MutableState<Boolean> = mutableStateOf(false)
+        private set
```
kaushikrw:
```suggestion
    var initialEvaluation : Boolean by mutableStateOf(false)
        private set
```
This needs to be a var of type `Boolean` for the private setter to work. Otherwise any consumer can simply set a value directly like so -
```kotlin
initialEvaluation.value = false
```

I also don't think there is a need to mark these properties and methods as internal since the class is already marked internal.

Similarly for other properties below too.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/PopupState.kt
```diff
@@ -0,0 +1,358 @@
...
+
+internal fun ArcGISFeature.toPopup(): Popup {
+    val popupDefinition = when {
+        this.featureTable?.popupDefinition != null -> {
+            this.featureTable?.popupDefinition
+        }
+
+        this.getFeatureSubtype() != null && this.featureTable is ArcGISFeatureTable -> {
+            (this.featureTable as ArcGISFeatureTable).subtypeSubtables
+                .firstOrNull { it.name == this.getFeatureSubtype()?.name }
```
kaushikrw:
```suggestion
                .firstOrNull { it.subtype == this.getFeatureSubtype()}
```
The `FeatureSubtype` can be checked directly here.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/fieldselement/FieldsElementState.kt
```diff
@@ -15,42 +15,30 @@
...
-    val fieldsToFormattedValuesMap = fieldNames.zip(element.formattedValues).toMap()
-    FieldsElementState(
+    constructor(element: FieldsPopupElement, popup: Popup) : this(
         title = element.title,
         description = element.description,
-        fieldsToFormattedValuesMap,
-        id = PopupElementState.createId()
+        fieldsToFormattedValues = element.fields.map { it.label }
+            .zip(element.formattedValues)
+            .toMap(),
```
kaushikrw:
```suggestion
        fieldsToFormattedValues = element.fields
            .mapIndexed { index, field -> field.label to element.formattedValues[index] }
            .toMap(),
```

The intermediate collection from `zip` can be avoided by using `mapIndexed` instead.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/media/MediaElementState.kt
```diff
@@ -55,31 +45,48 @@ import java.util.UUID
...
+    val element: MediaPopupElement,
+    val popup: Popup,
     override val id : Int = createId()
 ) : PopupElementState() {
 
-    constructor(
-        mediaPopupElement: MediaPopupElement,
+    lateinit var description: String
+    lateinit var title: String
+    lateinit var media: List<PopupMediaState>
```
kaushikrw: I'm wondering why these properties are mutable.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/media/MediaElementState.kt
```diff
@@ -55,31 +45,48 @@ import java.util.UUID
...
+
+    /**
+     * Indicates if the evaluateExpression function for the [popup] has been run.
+     */
+    internal var isInitialized : MutableState<Boolean> = mutableStateOf(false)
+        private set
+
+    fun init(
```
kaushikrw: Similarly, this indicates this `MediaElementState` can be initialized multiple times. Any particular reason why this approach compared to the typical constructor pattern?

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/media/MediaPopupElement.kt
```diff
@@ -39,13 +42,27 @@ internal fun MediaPopupElement(
...
+    }
+    if (state.isInitialized.value) {
+        MediaPopupElement(
+            title = state.title,
+            description = state.description,
+            stateId = state.id,
+            media = state.media,
+            onClickedMedia = onClickedMedia
+        )
+    }
```
kaushikrw, line 63: This initialization approach appears somewhat complex. Since the `init` can be called many times, this makes the state mutable plus the scope provided is associated with the composition, and several refresh or fetch operations depend on it, it might be helpful to consider simplifying this process.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/utilityassociationselement/UtilityAssociationDetails.kt
```diff
@@ -0,0 +1,245 @@
...
+internal fun UtilityAssociationDetails(
+    state: UtilityAssociationsElementState,
+    modifier: Modifier = Modifier
+) {
+    val associationResult = state.selectedAssociationResult ?: return
+    val filter = state.selectedFilterResult?.filter ?: return
+    val association = associationResult.association
+    var showConfirmationDialog by remember {
+        mutableStateOf(false)
+    }
```
kaushikrw:
```suggestion
```
Since Popups are read only the confirmation dialog stuff can be removed.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/utilityassociationselement/UtilityAssociationGroupResult.kt
```diff
@@ -0,0 +1,329 @@
...
+@Composable
+private fun AssociationItem(
+    title: String,
+    association: UtilityAssociation,
+    associatedFeature: ArcGISFeature,
+    enabled: Boolean,
+    onClick: () -> Unit,
+    onDetailsClick: () -> Unit,
+    modifier: Modifier = Modifier
+) {
```
kaushikrw, line 213: Some of the UI and rules for this has been updated. It might be a good idea to revisit this before release and update it accordingly.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/utilityassociationselement/UtilityAssociationsElementState.kt
```diff
@@ -0,0 +1,120 @@
...
+ * @param scope The [CoroutineScope] to launch coroutines from.
+ */
+internal class UtilityAssociationsElementState(
+    element: UtilityAssociationsPopupElement,
+    scope: CoroutineScope
+) : PopupElementState() {
+    override val id : Int = element.hashCode()
+    val label : String = element.title
+    val description: String = element.description
+    //    val isVisible : StateFlow<Boolean> = element.isVisible
```
kaushikrw:
```suggestion
```

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/utilityassociationselement/UtilityAssociations.kt
```diff
@@ -0,0 +1,329 @@
...
+            align = Alignment.Top
+        ),
+        shape = RoundedCornerShape(15.dp),
+        color = MaterialTheme.colorScheme.surfaceContainer
+    ) {
+        LazyColumn(
+            modifier = Modifier.clip(shape = RoundedCornerShape(15.dp)),
+            state = lazyListState
+        ) {
+            groupResult.associationResults.forEachIndexed { index, info ->
```
kaushikrw:
```suggestion
            groupResult.associationResults.forEachIndexed { index, info ->
```
IIRC popup spec defines a [displayCount](https://runtime-zip.esri.com/userContent/runtime/kotlin/doc_kotlin/300.0.0/last_good/html/arcgis-maps-kotlin/com.arcgismaps.mapping.popup/-utility-associations-popup-element/display-count.html) for the number of associations to show. I would check with Swift to see how they've handles this.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/navigation/NavigationAction.kt
```diff
@@ -0,0 +1,53 @@
...
+package com.arcgismaps.toolkit.popup.internal.navigation
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+/**
+ * Indicates the action to take when navigating.
+ */
+@Parcelize
+internal sealed class NavigationAction(val value: Int) : Parcelable {
```
kaushikrw: I'm not sure Popup needs this `NavigationAction`. For forms, this was meant as a way to cache user actions across configuration changes and dialogs. Since popup is read only a navigation action does not need a confirmation like in forms (which is where they are cached).

kaushikrw (changes requested):
@puneet-pdx I wasn't able to load any map to test in the micro-app but here is some initial feedback.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/PopupState.kt
puneet-pdx: added

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/Popup.kt
puneet-pdx: This check is not needed here and I have removed it. We are displaying progress/loading indicator in the PopupScreen.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/Popup.kt
puneet-pdx: I vaguely remember Soren mentioning it(I am not 100%). I have changed it to warning for now. Soren can chime in later.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/Popup.kt
puneet-pdx: yes, updated

microapps/PopupApp/app/src/main/java/com/arcgismaps/toolkit/popupapp/screens/mapscreen/MapViewModel.kt
puneet-pdx: updated

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/Popup.kt
puneet-pdx: No, I checked the design, there is no requirement. I will remove support for it.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/navigation/NavigationAction.kt
puneet-pdx: Ok. I have removed it.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/utilityassociationselement/UtilityAssociationsElementState.kt
puneet-pdx: updated

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/utilityassociationselement/UtilityAssociationGroupResult.kt
puneet-pdx: Yes, the code on `v.next` is missing support for viewing details for an association and onDetailsClick is never invoked.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/utilityassociationselement/UtilityAssociationDetails.kt
puneet-pdx: updated

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/utilityassociationselement/UtilityAssociations.kt
puneet-pdx: I have added support for taking in account the display count.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/PopupState.kt
puneet-pdx: updated

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/fieldselement/FieldsElementState.kt
puneet-pdx: updated

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/PopupState.kt
puneet-pdx: Updated. We should make similar change for FormStateData as well 

https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/blob/feature-branches/forms/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/FeatureFormState.kt#L331

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/media/MediaElementState.kt
puneet-pdx: I have updated them, only media is mutable now as it needs to be created from inside a composable

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/media/MediaElementState.kt
puneet-pdx: I have updated it to not be an init function by a create function for creating media

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/media/MediaPopupElement.kt
puneet-pdx: I have refactored this a bit, to prevent creation of media on recompositions. Lets discuss...

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/media/MediaElementState.kt
```diff
@@ -55,31 +47,46 @@ import java.util.UUID
...
+    val scope: CoroutineScope,
     override val id : Int = createId()
 ) : PopupElementState() {
 
-    constructor(
-        mediaPopupElement: MediaPopupElement,
-        scope: CoroutineScope,
+    val description: String = element.description
+    val title: String = element.title
+    lateinit var media: List<PopupMediaState>
```
kaushikrw: Any particular reason this is late initialized? `isPopupMediaCreated` should be enough to infer if the media creation is completed, no?

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/media/MediaPopupElement.kt
kaushikrw: This looks good now!

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/PopupState.kt
```diff
@@ -0,0 +1,376 @@
...
+                attachments = attachments,
+                coroutineScope = coroutineScope,
+                ignoreList = ignoreList
+            )
+            popupStateData.setStates(states)
+        }
+        store.addLast(popupStateData)
+    }
+
+    public constructor(popup: Popup, scope: CoroutineScope) : this(popup) {
```
kaushikrw, line 125: Missing doc for the public ctor.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/screens/ContentAwareTopBar.kt
```diff
@@ -0,0 +1,225 @@
...
+@Composable
+private fun PopupTitle(
+    title: String,
+    subTitle: String,
+    showCloseIcon: Boolean,
+    showBackIcon: Boolean,
+    onBackPressed: () -> Unit,
+    onClose: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
```
kaushikrw, line 172: One thing I noticed while running the app is, the popup shows up in an expandable group with an empty title. So I was wondering why the expandable group since there is no group element in popups.

kaushikrw (commented):
Looking good! Just a few more comments below.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/media/MediaElementState.kt
puneet-pdx: Hmm. I see your point. I have changed it to be an empty list on initialization.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/PopupState.kt
puneet-pdx: added

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/screens/ContentAwareTopBar.kt
puneet-pdx: All the element types are displayed as an expandable cards with title being the title of the element. This is a requirement coming from the design
_A stretch goal is to have the individual view elements be "collapsible". When collapsed, a popup element view just displays the title of the element._
https://devtopia.esri.com/runtime/common-toolkit/blob/main/designs/Popup/Popup.md#miscellaneous

puneet-pdx:
@kaushikrw I have addressed your comments here and also enabled the close button in the micro app. I am looking into updating `UNAssociationsScreen` with the latest changes.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/screens/ContentAwareTopBar.kt
kaushikrw: Ah I see. Perhaps it is a design discussion then, because the collapsed elements do have a title. I will leave that up to you..

microapps/PopupApp/app/src/main/java/com/arcgismaps/toolkit/popupapp/screens/mapscreen/MainScreen.kt
```diff
@@ -63,13 +68,32 @@ fun MainScreen(viewModel: MapViewModel) {
             skipHiddenState = false
         )
     )
+    LaunchedEffect(scaffoldState.bottomSheetState.currentValue) {
+        if (scaffoldState.bottomSheetState.currentValue == SheetValue.Hidden) {
+            unselectFeature(viewModel.geoElement, viewModel.layer)
+            viewModel.updatePopupState(null)
+            viewModel.setGeoElement(null)
+        }
+    }
```
kaushikrw, line 75: These sequence of steps are already happening in the `Popup.onDismiss` event so it seems redundant to do the same here.

kaushikrw (approved):
Looks good! I just have one comment below.

microapps/PopupApp/app/src/main/java/com/arcgismaps/toolkit/popupapp/screens/mapscreen/MainScreen.kt
puneet-pdx: The `Popup.onDismiss` does not get invoked if the user manually closes the bottomsheet. If that happens then without this there will be no bottomsheet on the screen and there will be a feature selected on the screen. This makes sure that the behavior is consistent for both the workflows.

microapps/PopupApp/app/src/main/java/com/arcgismaps/toolkit/popupapp/screens/mapscreen/MainScreen.kt
```diff
@@ -63,13 +68,28 @@ fun MainScreen(viewModel: MapViewModel) {
...
+
     BottomSheetScaffold(
         sheetContent = {
-            if (viewModel.popup != null) {
-                Popup(viewModel.popup!!, Modifier.animateContentSize())
+            if (viewModel.popupState != null) {
+                Popup(
+                    viewModel.popupState!!,
```
sorenoid: could use smart casting here rather than the !! operator.

```suggestion
            val state = viewModel.popupState
            if (state != null) {
                Popup(
                    state,
```

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/navigation/PopupNavHost.kt
```diff
@@ -0,0 +1,150 @@
...
+
+        composable<NavigationRoute.UNAssociationDetailView> { backStackEntry ->
+            val route = backStackEntry.toRoute<NavigationRoute.UNAssociationDetailView>()
+            val popupStateData = remember(backStackEntry) { state.getActivePopupStateData() }
+            // Get the selected UtilityAssociationsElementState from the state collection
+            val utilityAssociationsElementState = popupStateData.stateCollection[route.stateId]
+                // guard against null value
+                as? UtilityAssociationsElementState ?: return@composable
+            // Display the association details
+            UtilityAssociationDetails(
```
sorenoid: It seems like this function should not be called if the result and filter are null, rather than passing in the state and returning early from a composable function. Could the parameters be 

```kotlin
val result = utilityAssociationsElementState.state.selectedAssociationResult
val filter = state.selectedFilterResult?.filter
if (result != null && filter != null) {
    UtilityAssociationDetails(
        result,
        filter,
        modifier = Modifier.fillMaxSize()
    )
}
```

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/utilityassociationselement/UtilityAssociationDetails.kt
```diff
@@ -0,0 +1,183 @@
...
+ * @param state The [UtilityAssociationsElementState] of the element.
+ * @param modifier The [Modifier] to apply to this layout.
+ */
+@Composable
+internal fun UtilityAssociationDetails(
+    state: UtilityAssociationsElementState,
+    modifier: Modifier = Modifier
+) {
+    val associationResult = state.selectedAssociationResult ?: return
+    val filter = state.selectedFilterResult?.filter ?: return
```
sorenoid: see comment on this function at the call site

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/utilityassociationselement/UtilityAssociationDetails.kt
```diff
@@ -0,0 +1,183 @@
...
+                    onValueChange = {},
+                    modifier = Modifier
+                        .padding(horizontal = 20.dp, vertical = 5.dp)
+                        .fillMaxWidth()
+                )
+            }
+        }
+        associationResult.getFractionAlongEdge()?.let { fraction ->
+            FractionAlongEdgeControl(
+                fraction = associationResult.getFractionAlongEdge()!!.toFloat(),
```
sorenoid:
```suggestion
                fraction = fraction.toFloat(),
```

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/utilityassociationselement/UtilityAssociationsElementState.kt
```diff
@@ -0,0 +1,119 @@
...
+) : PopupElementState() {
+    override val id : Int = element.hashCode()
+    val label : String = element.title
+    val description: String = element.description
+    val displayCount: Int = element.displayCount
+
+    private var _loading: MutableState<Boolean> = mutableStateOf(true)
+
+    private var _filters: MutableState<List<UtilityAssociationsFilterResult>> =
+        mutableStateOf(emptyList())
```
sorenoid: nitpick: normally a backing var is directly above the val that uses it.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/PopupState.kt
```diff
@@ -0,0 +1,383 @@
...
+                is UtilityAssociationsPopupElement -> {
+                    states.add(
+                        element,
+                        UtilityAssociationsElementState(element, coroutineScope)
+                    )
+                }
+
+                else -> {
+                    // TODO remove for release
+                    println("encountered element of type ${element::class.java}")
```
sorenoid: don't forget to remove this!

sorenoid (changes requested):
Wow! that's quite an undertaking. I will test it out tomorrow. For now here is my review of the code.

sorenoid (approved):
Great work!

puneet-pdx:
Thanks @kaushikrw and @sorenoid for the reviews! Merging this.

PR #997, opened 2025-10-13
Add fix for popup crash

puneet-pdx:
Related to issue: # https://devtopia.esri.com/runtime/kotlin/issues/6482

### Description:

The popup component currently crashes when you tap on a feature on the map to load its popup. The MediaElementState currently throws an exception when the Image downloaded is null, leading to the crash. The configuration on the server changed recently where it has stopped returning images if there is no user agent in the header of the request.

Swift is using `ArcGISEnvironment.urlSession.data(from: url)` to download their images which probably applies the `user-agent` to the requests.

### Summary of changes:

- Add user agent string to the header of the image request
- Add default image drawable to use if the image returned from the server is null
- update annotation

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/995/
  - [ ] Yes
  - [ ] No

### DIFF
--- a/toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/media/MediaElementState.kt
+++ b/toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/media/MediaElementState.kt
@@ -45,6 +45,10 @@ import kotlinx.coroutines.withContext
 import java.io.File
 import java.util.UUID
 
+internal object PopupConstants {
+    const val USER_AGENT = "ArcGIS Maps SDK for Kotlin Toolkit"
+}
+
 /**
  * Represents the state of an [MediaPopupElement]
  */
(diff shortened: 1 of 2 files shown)


toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/element/media/MediaElementState.kt
```diff
@@ -42,6 +45,10 @@ import kotlinx.coroutines.withContext
 import java.io.File
 import java.util.UUID
 
+internal object PopupConstants {
+    const val USER_AGENT = "ArcGIS Maps SDK for Kotlin Toolkit"
```
puneet-pdx: PS - This is only being used for coil requests to fetch media images and no where else.

sorenoid (approved):

rchintapalli (approved):

puneet-pdx:
Thanks @sorenoid @rchintapalli merging...

PR #1010, opened 2025-10-20
Guard SuspendableCancellableCoroutine to only be resumed once

eri9000:
Related to issue: https://devtopia.esri.com/runtime/kotlin/issues/6742

### Summary of changes:
- Guards against `resume` being more than once
- Adds test cases  

### Pre-merge Checklist

  - [ ] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/1007/ -> ClientCertificateAuthenticationTest are passing ✅ 
  - [ ] Yes
  - [ ] No

### DIFF
--- a/toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/AuthenticatorState.kt
+++ b/toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/AuthenticatorState.kt
@@ -40,6 +40,7 @@ import com.arcgismaps.httpcore.authentication.OAuthUserSignIn
 import com.arcgismaps.httpcore.authentication.PasswordCredential
 import com.arcgismaps.httpcore.authentication.ServerTrust
 import com.arcgismaps.httpcore.authentication.TokenCredential
+import java.util.concurrent.atomic.AtomicBoolean
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.channels.awaitClose
 import kotlinx.coroutines.channels.trySendBlocking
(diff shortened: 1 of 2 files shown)


toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/AuthenticatorState.kt
```diff
@@ -386,24 +387,33 @@ private class AuthenticatorStateImpl(
...
+            /**
+             * Local function to resume the continuation only once.
+             * @param value the alias to resume with, or null if cancelling.
+             */
+            fun resumeOnce(value: String?) {
+                if (hasResumed.getAndSet(true)) return
                 _pendingClientCertificateChallenge.value = null
-                continuation.resume(null) { _, _, _ -> }
+                continuation.resume(value) { _, _, _ -> }
             }
```
sorenoid, line 400: I see that `hasResumed` and the local function are referenced from `aliasCallback`, which is in turn referenced by the instance of `ClientCertificateChallenge` set as the value of the member property `_pendingClientCertificateChallenge`. All of those things are subject to garbage collection if another challenge is set on `_pendingClientCertificateChallenge`. 

I also see that the only way to set the pending challenge is through this function so it all hangs together, just a little perilous. 

I think it all works as is. I might suggest a small wrapper class like CoreTaskWrapper in the SDK to provide an await function which runs the lambda, a method called `resumeOnce` and an atomic boolean property `hasResumed`.

sorenoid (commented):
I think this is ok. It seems a bit precarious to introduce a local function in a lambda which might get garbage collected. It would be less risky to introduce a small wrapper class like CoreTaskWrapper to ensure the new local variables don't get garbage collected early, or if the thing that now references them (_pendingClientCertificateChallenge) changes. But I am not as familiar with this code as others and defer approving or requesting changes.

eri9000:
Work was done in other PR https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/pull/1011

PR #1018, opened 2025-10-30
`FeatureForm` : Add paging to searching feature candidates

kaushikrw:
Related to issue: [#apollo/1478](https://devtopia.esri.com/runtime/apollo/issues/1478)

### Description:

Adds paging to searching feature candidates when there are more features to fetch.

### Summary of changes:

- Updated the view model to keep fetching next pages until some search results are available.
- Updated the corresponding screen to show a loading indicator while this fetch is in progress.

### Pre-merge Checklist

  - [ ] link:
  - [ ] Yes
  - [ ] No

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/AddAssociationFromSourceViewModel.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/AddAssociationFromSourceViewModel.kt
@@ -231,6 +231,12 @@ internal class AddAssociationFromSourceViewModel(
     val isSearchingForCandidates: Boolean
         get() = _isSearchingForCandidates.value
 
+    /**
+     * Indicates whether the ViewModel is currently fetching the next page of feature candidates.
+     * This is used to prevent multiple simultaneous fetches.
+     */
+    private var _isFetchingNextPage = false
+
     init {
         viewModelScope.launch {
             // Whenever the selected source or asset type changes, reload the feature candidates
@@ -536,6 +542,38 @@ internal class AddAssociationFromSourceViewModel(
         setAssociatedFeaturesFilterText("")
     }
 
+    /**
+     * Loads more feature candidates from the selected source and asset type if there are more
+     * candidates to fetch. This method will not fetch more candidates if a fetch is already in
+     * progress or if there is an active search filter.
+     *
+     * This is not thread-safe as it is intended to be called from the ui thread only.
+     */
+    suspend fun loadMoreFeatureCandidates() {
+        // Only fetch the next page if not already fetching and there's no active search filter
+        if (_isFetchingNextPage.not() && featureSourcesFilterText.isEmpty()) {
+            val source = _selectedSource.value ?: return
+            val assetType = _selectedAssetType.value ?: return
+            val nextQueryParams = _featureCandidatesUiState.value.nextQueryParams ?: return
+            _isFetchingNextPage = true
+            source.queryFeatures(
+                assetType = assetType,
+                parameters = nextQueryParams
+            ).onSuccess { result ->
+                // Append the newly fetched candidates to the existing list
+                val updatedCandidates =
+                    _featureCandidatesUiState.value.candidates + result.candidates
+                // Update the candidates list with the updated list
+                _featureCandidatesUiState.value = FeatureCandidatesUiState(
+                    isLoading = false,
+                    candidates = updatedCandidates,
+                    nextQueryParams = result.nextQueryParams
+                )
+            }
+            _isFetchingNextPage = false
+        }
+    }
+
     companion object {
         fun Factory(
             featureForm: FeatureForm,
@@ -586,8 +624,7 @@ internal data class NewAssociationOptions(
  * @param isLoading Indicates whether the feature candidates are currently being loaded.
  * @param error An optional [Throwable] representing any error that occurred during loading.
  * @param candidates A list of loaded [UtilityAssociationFeatureCandidate] objects.
- * @param nextQueryParams Optional [QueryParameters] for fetching the next set of candidates,
- * if any.
+ * @param nextQueryParams Optional [QueryParameters] for fetching the next set of candidates, if any.
  */
 internal data class FeatureCandidatesUiState(
     val isLoading: Boolean = false,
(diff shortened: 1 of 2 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/AddAssociationFromSourceViewModel.kt
```diff
@@ -540,11 +582,18 @@ internal data class NewAssociationOptions(
 /**
  * Represents the UI state for loading feature candidates, including loading status,
  * any errors encountered, and the list of loaded candidates.
+ *
+ * @param isLoading Indicates whether the feature candidates are currently being loaded.
+ * @param error An optional [Throwable] representing any error that occurred during loading.
+ * @param candidates A list of loaded [UtilityAssociationFeatureCandidate] objects.
+ * @param nextQueryParams Optional [QueryParameters] for fetching the next set of candidates,
+ * if any.
```
puneet-pdx: this can just go on line 589

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/AddAssociationFromSourceViewModel.kt
```diff
@@ -174,12 +175,43 @@ internal class AddAssociationFromSourceViewModel(
...
-    }.map { (text, _) ->
+    }.map { (text, candidatesUiState) ->
         return@map if (text.isNotEmpty()) {
-            val filteredList = filterFeatureCandidates(text)
+            val candidateList = candidatesUiState.candidates.toMutableList()
+            var nextQueryParams = candidatesUiState.nextQueryParams
+            var filteredList = filterFeatureCandidates(text)
+            // If the filtered list is empty and there are more candidates to fetch, keep fetching
+            // until we either get results or run out of candidates
+            while (filteredList.isEmpty() && nextQueryParams != null) {
```
puneet-pdx, line 185: should this logic be changed to 
```suggestion
            while (nextQueryParams != null) {
```
as we don't want to stop looking till `nextQueryParams` is not null.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/AddAssociationFromSourceViewModel.kt
```diff
@@ -174,12 +175,43 @@ internal class AddAssociationFromSourceViewModel(
...
         associatedFeaturesFilterText to _featureCandidatesUiState.value
-    }.map { (text, _) ->
+    }.map { (text, candidatesUiState) ->
         return@map if (text.isNotEmpty()) {
-            val filteredList = filterFeatureCandidates(text)
+            val candidateList = candidatesUiState.candidates.toMutableList()
+            var nextQueryParams = candidatesUiState.nextQueryParams
+            var filteredList = filterFeatureCandidates(text)
+            // If the filtered list is empty and there are more candidates to fetch, keep fetching
+            // until we either get results or run out of candidates
```
puneet-pdx: this statement is not correct, there is not an or but an and operator in play here

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/AddAssociationFromSourceViewModel.kt
puneet-pdx: also can we just use `candidatesUiState.nextQueryParams` here instead of a local variable `nextQueryParams`
that way we will not have to update both the _featuerCandidateUiState and the nextQueryParams on line 198

puneet-pdx (commented):
@kaushikrw some minor comments

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/AddAssociationFromSourceViewModel.kt
kaushikrw: > as we don't want to stop looking till nextQueryParams is not null.

This is not true. We only want to fetch pages until we have some results. We do not want to fetch all the pages if there is no need to do so.

> also can we just use candidatesUiState.nextQueryParams here instead of a local variable nextQueryParams
that way we will not have to update both the _featuerCandidateUiState and the nextQueryParams on line 198

Since we're using the `map` operator , the `candidatesUiState` becomes outdated after we fetch a new page and update the `_featureCandidatesUiState` state with the new result. If we were to use `mapLatest` it would cancel the transformation and start a new one, but it is experimental. Hence we need to store an updated `nextQueryParams` for subsequent fetches.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/AddAssociationFromSourceViewModel.kt
kaushikrw: > there is not an or but an and operator in play here

That is true but this is talking about the exit condition. Loop continues only while both conditions are true. It stops when either a match is found (filteredList no longer empty) or there are no more pages (nextQueryParams is null).

Does this sound clearer?

> // Keep fetching additional pages while no candidates match the filter and a next page (nextQueryParams) is 
// available. Stop when matches appear or there are no more pages.

kaushikrw:
@puneet-pdx updated the PR to also fetch new pages (if any) when the end of the scroll is reached.

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/AddAssociationFromSourceViewModel.kt
puneet-pdx: yes, the new comment is clearer

puneet-pdx (approved):
@kaushikrw changes look good.

kaushikrw:
Thanks @puneet-pdx, merging.

PR #1024, opened 2025-11-07
`FeatureForm` : Use proper formatting and casting for a fractionAlongEdge

kaushikrw:
Related to issue: [#apollo/1512](https://devtopia.esri.com/runtime/apollo/issues/1512)

### Description:

This PR fixes unneeded casting and adds formatting for the `PercentAlongEdge` control.

### Summary of changes:

- Updated the `PercentAlongEdge` to use a `double` type instead of a `float` to avoid casting to other float and back unnecessarily.
- Updated the precision of this fraction value to be to 2 digits after the decimal.

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/job/vtest/job/toolkit/1043/
  - [ ] Yes
  - [ ] No

### DIFF
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/PercentAlongControl.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/PercentAlongControl.kt
@@ -39,8 +39,6 @@ import androidx.compose.ui.unit.dp
 import com.arcgismaps.toolkit.featureforms.R
 import com.arcgismaps.utilitynetworks.UtilityAssociation
 import com.arcgismaps.utilitynetworks.UtilityElement
-import java.math.BigDecimal
-import java.math.RoundingMode
 
 /**
  * A composable that represents a percent along control of a [UtilityAssociation]. This can
@@ -84,8 +82,7 @@ internal fun PercentAlongControl(
             Slider(
                 value = value.toFloat(),
                 onValueChange = {
-                    val decimal = BigDecimal(it.toString()).setScale(2, RoundingMode.DOWN)
-                    value = decimal.toDouble()
+                    value = (it * 100).toInt() / 100.0
                 },
                 enabled = enabled,
                 modifier = Modifier
--- a/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/PercentAlongControl.kt
+++ b/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/PercentAlongControl.kt
@@ -26,7 +26,7 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Slider
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableFloatStateOf
+import androidx.compose.runtime.mutableDoubleStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
@@ -39,6 +39,8 @@ import androidx.compose.ui.unit.dp
 import com.arcgismaps.toolkit.featureforms.R
 import com.arcgismaps.utilitynetworks.UtilityAssociation
 import com.arcgismaps.utilitynetworks.UtilityElement
+import java.math.BigDecimal
+import java.math.RoundingMode
 
 /**
  * A composable that represents a percent along control of a [UtilityAssociation]. This can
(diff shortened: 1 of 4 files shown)


toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/UtilityAssociationDetails.kt
```diff
@@ -266,9 +266,7 @@ internal fun RemoveAssociationConfirmationDialog(
  */
 internal fun UtilityAssociationResult.getFractionAlongEdge(): Double? {
     return when (association.associationType) {
-        UtilityAssociationType.JunctionEdgeObjectConnectivityFromSide,
-        UtilityAssociationType.JunctionEdgeObjectConnectivityMidspan,
-        UtilityAssociationType.JunctionEdgeObjectConnectivityToSide -> {
+        UtilityAssociationType.JunctionEdgeObjectConnectivityMidspan -> {
```
kaushikrw, line 269: This change is due to the following [design](https://devtopia.esri.com/runtime/common-toolkit/blob/main/designs/Forms/UtilityAssociationsFormElement.md#:~:text=UtilityAssociation.fractionAlongEdge%20%2D%20Used%20when%20the%20association%20type%20is%20junctionEdgeObjectConnectivityMidspan.%20Shown%20as%20a%20percentage%20value%20on%20the%20second%20line.%20(e.g.%20%2235%25%22).)

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/PercentAlongControl.kt
```diff
@@ -80,9 +82,10 @@ internal fun PercentAlongControl(
                 modifier = Modifier.fillMaxWidth()
             )
             Slider(
-                value = value,
+                value = value.toFloat(),
                 onValueChange = {
-                    value = it
+                    val decimal = BigDecimal(it.toString()).setScale(2, RoundingMode.DOWN)
```
puneet-pdx: we can do this just by using arithmetic,
```suggestion
                    val decimal = kotlin.math.floor(it * 100) / 100
```
It’s efficient and avoids object allocation.

puneet-pdx (commented):
@kaushikrw just a minor suggestion. Looks good overall

toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/PercentAlongControl.kt
kaushikrw: I'm not sure this would work. The `floor` gives you the value as a `Float` but we need a `Double`. Converting this to a `Double` will result in the same precision/rounding problem as before.

kaushikrw: @puneet-pdx This gave me another idea though, we can just use implicit type conversion to a double like so, which would be more efficient as you noted.
```
value = (it * 100).toInt() / 100.0
```

puneet-pdx: I tested the logic out and I was getting the same values for value, with what you had and `math.floor` approach

```Kotlin
val decimal = BigDecimal(it.toString()).setScale(2, RoundingMode.DOWN)
                    val decimal2 = kotlin.math.floor(it * 100) / 100
                    Log.i("PercentAlongControl", "Slider value changed to it ${it.toDouble()}")
                    Log.i("PercentAlongControl", "Slider value changed to decimal - $decimal")
                    Log.i("PercentAlongControl", "Slider value changed to decimal2 - $decimal2")
                    value = decimal.toDouble()
```

>I'm not sure this would work. The floor gives you the value as a Float but we need a Double. Converting this to a Double will result in the same precision/rounding problem as before.

I am not sure what you mean here, we are converting the value toDouble with the BigDecimal too

kaushikrw: > I am not sure what you mean here, we are converting the value toDouble with the BigDecimal too

For ex, if the value `it` from the slider is `0.63167262...`

`decimal2` is of type `Float` here. The api expects a `Double`. When you convert that into a double `decimal2.toDouble()` you might get something close to `0.6300009` due to binary conversion and precision. It will never really be `0.63`.

What we actually want is a double with a value of `0.63` which is exactly what the `BigDecimal` gives us.

kaushikrw: > When you convert that into a double decimal2.toDouble() you might get something close to 0.6300009 due to binary conversion and precision. It will never really be 0.63.

This was the actual initial issue btw - https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/blob/0666e2512aedabae29116060a744518a47601c47/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/utilitynetwork/AddAssociationFromSourceViewModel.kt#L503

puneet-pdx (approved):
@kaushikrw 👍🏽

kaushikrw:
Thanks @puneet-pdx

PR #1039, opened 2025-12-06
Fail gracefully when the default browser does not support custom tabs

eri9000:
Related to issue: https://devtopia.esri.com/runtime/kotlin/issues/6659

### Summary of changes:
- Adds test to verify `AuthenticationActivity` behavior
- Modifies `AuthenticationActivity` to process incoming redirects in through `onCreate`.
- Provides better error handling when 
- Modifies `OAuthUserSignInContract` to handle errors coming back in the Result
- Creates a new Authenticator Exceptions class 

### Pre-merge Checklist

  - [ ] link:
  - [ ] Yes
  - [ ] No


toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/AuthenticationActivityExternalBrowserTest.kt
```diff
@@ -0,0 +1,138 @@
...
+    @After
+    fun tearDown() {
+        unmockkAll()
+        Intents.release()
+    }
+
+    /**
+     * Given [AuthenticationActivity] with a browser that supports Custom Tabs
+     * When [AuthenticationActivity] starts
+     * Then an intent should be fired with ACTION_VIEW, simulating a Custom Tabs launch
```
gunt0001, line 62: I'm confused - looking at `processAuthenticationFlow()`, when an intent is handled with `ACTION_VIEW`, that means we are getting the redirect from Custom Tabs. It doesn't mean we are launching Custom Tabs. If this test is meant to check that the custom tabsl intent is being launched, shouldn't we be checking for something like the intent being of type `CustomTabsIntent`?

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/AuthenticationActivityExternalBrowserTest.kt
```diff
@@ -0,0 +1,138 @@
...
+            Intent.ACTION_VIEW,
+            Uri.parse(redirectUri)
+        )
+
+        // invoke the activity with the redirect URI
+        val scenario = ActivityScenario.launchActivityForResult<AuthenticationActivity>(intent)
+        scenario.close()
+        val result = scenario.result
+        assertThat(result.resultCode).isEqualTo(RESULT_CODE_SUCCESS)
+        assertThat(result.resultData?.getStringExtra("KEY_INTENT_EXTRA_RESPONSE_URI")).isEqualTo(redirectUri)
```
gunt0001, line 142: How is this test working? In other words, what closes the Custom Tabs and redirects into `AuthenticationActivity` with `RESULT_CODE_SUCCESS`?

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/AuthenticationActivity.kt
```diff
@@ -168,48 +167,82 @@ public class AuthenticationActivity internal constructor() : ComponentActivity()
...
+    /**
+     * Processes the authentication flow based on the intent action and data.
+     * - If the action is [Intent.ACTION_VIEW] and data is present, it handles the redirect intent.
+     * - If the Auth endpoint is missing, it handles the redirect intent with an error message
+     * indicating the failure to start authentication.
+     * - If the default browser cannot launch Custom Tabs, it handles the redirect intent with
+     *   an appropriate error message.
+     * @since 300.0.0
+     */
+    private fun processAuthenticationFlow() {
```
gunt0001, line 179: Nitpick - naming suggestion:
```suggestion
    private fun promptWithCustomTabs() {
```

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/AuthenticationActivity.kt
```diff
@@ -168,48 +167,82 @@ public class AuthenticationActivity internal constructor() : ComponentActivity()
...
+        val localIntent = intent
```
gunt0001: What is the resone for reading `intent` into a local variable?

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/AuthenticationActivity.kt
```diff
@@ -168,48 +167,82 @@ public class AuthenticationActivity internal constructor() : ComponentActivity()
...
+        with(localIntent) {
+            val authEndpoint = getStringExtra(KEY_INTENT_EXTRA_URL)
+            when {
+                // This signals that we have been redirected back to the app from the browser
+                action == Intent.ACTION_VIEW && data != null -> {
```
gunt0001, line 184: Shouldn't this case only happening when we get an intent in `onNewIntent`? It seems we ar enow conflating tow different cases in `processAuthenticationFlow`:
1. launching custom tabs
2. handling the redirect from custom tabs

I think this function here should only handle the first case.

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/AuthenticationActivityExternalBrowserTest.kt
eri9000: These are the following intents that are captured by the test, 
```
AuthenticationActivityTest: Intent: action=null data=null extras=KEY_INTENT_EXTRA_URL
AuthenticationActivityTest: Intent: action=android.intent.action.VIEW data=https://example.com/auth extras=android.support.customtabs.extra.SESSION, android.support.customtabs.extra.EXTRA_ENABLE_INSTANT_APPS, androidx.browser.customtabs.extra.SHARE_STATE, com.android.browser.headers
```

As you can see, the first one is the one that the test uses to launch `AuthenticationActivity`, simulating the `OAuthUserSignInContract`. 

And the second one is the `CustomTabs` intent that  `AuthenticationActivity` launches. 
Since we are not signing in, we do not see the "redirect" intent that we would get from the browser 

>If this test is meant to check that the custom tabs intent is being launched, shouldn't we be checking for something like the intent being of type CustomTabsIntent?

I assume you mean some identifiable information that it will tell us that it is the CustomTabs Intent, I already check for `android.support.customtabs.extra.SESSION` inside the test but I could add more if you think is necessary

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/AuthenticationActivityExternalBrowserTest.kt
eri9000: This test does not open CustomTabs, it only verifies that `AuthenticationActivity` handles a "redirect" URI successfully. 
It does this by launching the activity with an intent that has a `redirect_uri` as data, simulating a browser redirecting back to the activity. 

When `AuthenticationActivity` processes a redirect URI successfully, it calls `setResult(RESULT_CODE_SUCCESS, newIntent)` [here](https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/pull/1039/changes#diff-0c8b7abb555d2fc6bfb408499740a175de565eabf3be045fdc9fec34216cd4e3R213)

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/AuthenticationActivity.kt
eri9000: i'd like to give it a more generic name since this likely will be the method to launch external browser or Auth Tabs (when we support it).

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/AuthenticationActivity.kt
eri9000: removed and just used the intent directly in `with()`

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/AuthenticationActivity.kt
eri9000: When testing different browsers sign-in, i found that some browsers might redirect with different flags and might come through `onCreate` instead of `onNewIntent`. 
This is just safe guarding this possible outcome.

eri9000:
@gunt0001 Please have another look

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/AuthenticationActivityExternalBrowserTest.kt
```diff
@@ -0,0 +1,144 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+
+/**
+ * Instrumentation test to verify [AuthenticationActivity]'s behavior.
+ *
+ * @since 300.0.0
+ */
+@RunWith(AndroidJUnit4::class)
+class AuthenticationActivityExternalBrowserTest {
```
gunt0001, line 46: Could we name it simply?
```suggestion
class AuthenticationActivityTest {
```

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/AuthenticationActivityExternalBrowserTest.kt
gunt0001: I still don't understand what this test is aiming to test. It looks like it is asserting that when launching `AuthenticationActivity` with `intent`, it receives this intent. Is this something worth testing?

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/AuthenticationActivityExternalBrowserTest.kt
```diff
@@ -0,0 +1,144 @@
...
+            )
+        }
+    }
+
+    /**
+     * Given [AuthenticationActivity] is launched with a default browser that does not support Custom Tabs
+     * When [AuthenticationActivity] starts
+     * Then the activity finishes with RESULT_CODE_CANCELED and includes an exception message in the result data
+     * @since 300.0.0
+     */
```
gunt0001, line 97: Would the following be a better description for this test?
```suggestion
    /**
     * Given a device with default browser that does not support Custom Tabs
     * When [AuthenticationActivity] receives an intent for OAuth sign-in
     * Then the activity finishes with RESULT_CODE_CANCELED and includes an exception message in the result data
     * @since 300.0.0
     */
```

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/AuthenticationActivityExternalBrowserTest.kt
```diff
@@ -0,0 +1,144 @@
...
+        val redirectUri = "kotlin-iap-test-1://auth/callback?code=123"
+        // Create an intent with ACTION_VIEW and the redirect URI, simulating the browser redirecting back to the app
+        val intent = Intent(
+            Intent.ACTION_VIEW,
+            Uri.parse(redirectUri)
+        )
+
+        // Launch the AuthenticationActivity with the redirect intent
+        val scenario = ActivityScenario.launchActivityForResult<AuthenticationActivity>(intent)
+        scenario.close()
```
gunt0001, line 137: Looking at the doc of `close`, should we be using `scenario.use { ... }` here or alternatively `ActivityScenarioRule`? The same would apply to the other tests.

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/AuthenticationActivityExternalBrowserTest.kt
gunt0001: Ok, makes sense. Would the following doc make that clearer?
```
/**
     * Given [AuthenticationActivity] receives an intent containing a valid redirect URI
     * When the activity processes the redirect intent
     * Then the activity finishes with RESULT_CODE_SUCCESS and includes the redirect URI in the result data
     * @since 300.0.0
     */
```

toolkit/authentication/src/androidTest/java/com/arcgismaps/toolkit/authentication/AuthenticationActivityExternalBrowserTest.kt
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /**
+     * Given [AuthenticationActivity] is launched with a default browser that does not support Custom Tabs
+     * When [AuthenticationActivity] starts
+     * Then the activity finishes with RESULT_CODE_CANCELED and includes an exception message in the result data
+     * @since 300.0.0
+     */
+    @Test
+    fun returnsExceptionWhenNoBrowserAvailable() {
```
gunt0001, line 99: Should this be named?
```suggestion
    fun returnsExceptionWhenNoCustomTabsAvailable() {
```

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/AuthenticationActivity.kt
gunt0001: I thought we won't support external browser? And when we move to Auth Tabs, we can rename the function to `promptWithAuthTab` at that point.

toolkit/authentication/src/main/java/com/arcgismaps/toolkit/authentication/AuthenticationActivity.kt
gunt0001: This is diverging quite a bit from the current implementation. At the minimum we will need test cases for the scenarios you mention. Lets discuss...

eri9000:
a simpler implementation is in PR https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/pull/1044 since we want to make gradual changes

PR #1066, opened 2026-02-05
Colinanderson/critical error

colinanderson:
Related to issue: https://devtopia.esri.com/runtime/red-october/issues/3094

### Description:

Adds an event to `LocalSceneView` for when the critical error changes.

image

Can be tested by trying to run the microapp in an emulator. Requires an up to date build e.g. `sdkBuildNumber=4865`.

### Summary of changes:

- add event
- add dialog to the microapp to show any critical error

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/1127/
  - [ ] Yes
  - [ ] No


microapps/LocalSceneViewApp/app/src/main/java/com/arcgismaps/toolkit/localsceneviewapp/screens/MainScreen.kt
```diff
@@ -18,9 +18,16 @@
 
 package com.arcgismaps.toolkit.localsceneviewapp.screens
 
+import android.util.Log
```
gunt0001: Unused import

gunt0001 (approved):
@colinanderson - looks good 👍

microapps/LocalSceneViewApp/app/src/main/java/com/arcgismaps/toolkit/localsceneviewapp/screens/MainScreen.kt
```diff
@@ -82,8 +89,31 @@ fun MainScreen() {
...
+                Log.e("LSV", it.toString())
+            }
+        }
     )
+
+    if (showErrorDialog) {
+        AlertDialog(
+            onDismissRequest = { showErrorDialog = false },
+            title = { Text(text = "Critical Error") },
+            text = { Text(text = criticalError?.message.toString()) },
```
eri9000, line 111: this will display "null" when the critical error's message is "null". Should we add a default error message here?

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/LocalSceneView.kt
```diff
@@ -162,6 +164,7 @@ public fun LocalSceneView(
     onTwoPointerTap: ((TwoPointerTapEvent) -> Unit)? = null,
     onPan: ((PanChangeEvent) -> Unit)? = null,
     onDrawStatusChanged: ((DrawStatus) -> Unit)? = null,
+    onCriticalErrorChanged: ((Throwable?) -> Unit)? = null,
```
eri9000, line 167: Nitpick: I think "Changed" is unnecessary. Changed implies that we should listen to this proactively as it can be different at times but I think this only will be called when initializing the view right? 

Are you expecting the views to have multiple critical errors? 

```suggestion
    onCriticalError: ((Throwable?) -> Unit)? = null,
```

eri9000 (approved):
@colinanderson 

Looks good to me 👍🏼 
Just one suggestion on the naming of the lambda but it should not hold this PR from getting merged.

microapps/LocalSceneViewApp/app/src/main/java/com/arcgismaps/toolkit/localsceneviewapp/screens/MainScreen.kt
colinanderson: Line 99 above should mean the dialog does not display if the error is `null`. At least that was the intention.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/LocalSceneView.kt
colinanderson: It isn't clear but one possible error is this which sounds like it could happen after the view has been set up

```
          name: criticalRenderingError
          value: 21000
          summary: A critical error occurred while rendering and the view is in an unusable state.
          remarks: Restarting the application is recommended to potentially resolve this problem.
          clients: [FLUTTER, GAME_ENGINE, KOTLIN, QT, SWIFT]
          since: 300.0.0
```

colinanderson: But there's no way for an app to recover from critical errors so maybe `onCriticalError` makes more sense. On the other hand it is a state flow so will give out the initial null value and thus "changes" if an error occurs.

eri9000: > On the other hand it is a state flow so will give out the initial null value and thus "changes" if an error occurs.

it's a StateFlow at the api level but here it is just a callback: `((Throwable?) -> Unit)? = null,`, which is covered by "on" prefix

colinanderson: But it will still have state behaviour so you'll can get the initial `null` and later a critical error so wouldn't changed in the name match other state changes like draw complete?
```
com...aps.toolkit.localsceneviewapp  E  Critical error: null                                                                                                                               
com...aps.toolkit.localsceneviewapp  E  Critical error: com.arcgismaps.exceptions.CriticalRenderingException: The graphics driver does not meet the Maps SDK's minimum requirements.       
                                        The graphics driver must at least meet the minimum requirements. GPU Device cannot be used for rendering. The following requirements are not met:  
                                        unsupported graphics api version: api version(3.0 < 3.2) or shaders version(3.0 < 3.2)                                                             
                                        single precision framebuffer                                                                                                                       
                                        compute and shader storage buffers. (errorCode: 21001)                                                                                             
                                                                                                                                                                                           
```

eri9000: Since the doc says this is a critical rendering error and the view becomes unusable, i'd treat it as a one-time event. In Compose, event callbacks are idiomatically named `onX`.

I don't feel strongly about either so up to you 👍🏼

microapps/LocalSceneViewApp/app/src/main/java/com/arcgismaps/toolkit/localsceneviewapp/screens/MainScreen.kt
eri9000: yes that's what i saw, if we don't think we'll ever have a case where critical error is not null then it should be fine to leave as is

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/LocalSceneView.kt
colinanderson: @eri9000 I discussed with Gunther and we decided to merge with the name as is - mainly to be consistent with other names but also to unblock - but we can take other ideas like your proxy proposal to a coffee meeting etc.

PR #1081, opened 2026-02-20
attempt to make warning and error callbacks not call due to phone rot…

sorenoid:
Related to issue: https://devtopia.esri.com/runtime/kotlin/issues/7292

### Description:

prevent error and warning callbacks from calling every time the phone is rotated.

### Summary of changes:

- use rememberSaveable, saving the Int error code of ArcGISException

### Pre-merge Checklist

  - [ ] link:
  - [ ] Yes
  - [ ] No


colinanderson (commented):
@sorenoid I haven't had a chance to run this but I don't think it will work correctly since `null` and empty list are valid states.

They might not be desirable as initial values that appear when the component is created but they might be in a running app e.g. where something is done to clear the warnings then you might want to see a lambda invoked that has empty list as its value.

PR #1092, opened 2026-03-05
Geoview focus traversal

shubham7109:
Related to issue: `#7228`

### Description:

Adds coordinated focus management across MapView, SceneView, LocalSceneView, and their Callout to support WCAG accessibility guidelines of capturing focus when modal content is displayed.

This PR introduces a shared accessibility coordinator to ensure only one of GeoView or Callout is focusable at a time and to drive focus handoff when Callout added/removed.

### Design patterns

- Centralized `GeoViewA11yCoordinator` manages shared focus state for GeoView + Callout.
- Sets GeoView explicit `focusProperties` to traverse to Callout next if available.
- Default `canFocus` to true for all GeoView composables, but continue to allow for a user driven property. 

### Changes made

- Added `GeoViewA11yCoordinator` to control focusability and accessibility events for GeoViews and Callout.
- GeoViewScope now injects the coordinator to provide callout focus through `calloutFocusRequester`
- MapView/SceneView/LocalSceneView composables attaches the coordinator's current focus state.
- Treat the Callout container layout as a focus group and manages focus placement and dispose handoff back to GeoView via `DisposableEffect`.
- Updated callout containers to use coordinator-managed focus requester.


toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
sorenoid, line 1: Generally, for this class, it should support the notion of mutually exclusive focus between the GeoView and a content lambda. not just a Callout.  Its AtomicBoolean has to do with if any content lambda is showing (of which there can be only one), and the GeoViewCompose's AtomicBoolean has to do with ensuring there is at most one Callout. 

The client should be able to turn all this stuff off by setting `canFocus` to false in the GeoView composables. I will test that.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
```diff
@@ -0,0 +1,106 @@
...
+    init {
+        // Set focus right away so initial compose AndroidView "focus interop" is set up correctly.
+        geoView.isFocusable = canFocus
+    }
+
+    /**
+     * The isCalloutBeingDisplayed property from the related GeoViewScope,
+     * Used to drive focus away from GeoView and into Callout when displayed.
+     */
+    private val isCalloutBeingDisplayed = AtomicBoolean(false)
```
sorenoid:
```suggestion
    private val isContentBeingDisplayed = AtomicBoolean(false)
```

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
```diff
@@ -0,0 +1,106 @@
...
+     */
+    val isGeoViewFocusable: Boolean
+        get() = canFocus && !isCalloutBeingDisplayed.get()
+
+    /**
+     * Used by GeoViewScope to notify here that a Callout is being displayed.
+     * Update GeoView to not be focusable and, on [View.post] of the calloutView, request focus
+     * into the Callout and send a content changed accessibility event.
+     */
+    internal fun onCalloutPlaced(calloutView: View) {
```
sorenoid:
```suggestion
    internal fun onContentComposed() {
```

a couple of things here: the name change, and also it doesn't seem necessary to use the rootView, but can just use the geoView property.

If it is necessary to use the rootView, I would change the name of the parameter to `calloutParentView` or `calloutRootView`, not `calloutView`.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
```diff
@@ -0,0 +1,106 @@
...
+        calloutView.post {
```
sorenoid: as discussed, using post guarantees the correct sequencing of focus changes on the View based GeoView.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
```diff
@@ -0,0 +1,106 @@
...
+            geoView.performAccessibilityAction(AccessibilityNodeInfo.ACTION_CLICK, null)
+        }
+    }
+
+    /**
+     * Using the current state of this coordinator, apply accessibility configuration to the GeoView.
+     *
+     * @param geoView The GeoView to apply configuration, defaults to the constructor property.
+     */
+    internal fun applyGeoViewA11yConfiguration(geoView: View = this.geoView) {
```
sorenoid: I suggest just removing this function and adding its code to onContentDisposed and onContentComposed.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -103,7 +105,9 @@ import kotlin.math.sin
  *
  * @since 200.5.0
  */
-public sealed class GeoViewScope protected constructor(private val geoView: GeoView) {
+public sealed class GeoViewScope constructor(
+    private val geoView: GeoView, private val a11yCoordinator: GeoViewA11yCoordinator
```
sorenoid: sealed classes are hard to update without breaking backwards compatibility, and unfortunately this one is public. We cannot add parameters to the primary constructor and we can't deprecate and hide the primary constructor. To get around this, you can create a protected secondary constructor, make the coordinator a nullable var property and initialize the coordinator in the secondary constructor's body.
Something like

```kotlin
    internal var a11yCoordinator: GeoViewA11yCoordinator? = null

    protected constructor(geoView: GeoView, canFocus: Boolean, focusRequester: FocusRequester): this (geoView) {
        
        this.a11yCoordinator  = GeoViewA11yCoordinator(canFocus, geoView, focusRequester)
    }
```

note, I pulled the coordinator out of the composables and instantiate it here inside the GeoViewScope. That isn't essential, just the way I did it. but either way it will need to be nullable.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -379,6 +383,7 @@ public sealed class GeoViewScope protected constructor(private val geoView: GeoV
             )
         }
         var animationDuration by remember { mutableIntStateOf(300) }
+        val rootView = LocalView.current
```
sorenoid: I am not sure this is needed. It worked for me without it, just using the geoView, as is done onContentDisposed

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -435,13 +435,20 @@ public sealed class GeoViewScope protected constructor(private val geoView: GeoV
                                 minSize = shapes.minSize
                             )
                             .animateContentSize()
-                            .focusRequester(focusRequester)
+                            .focusRequester(a11yCoordinator.calloutFocusRequester)
```
sorenoid: to deal with the nullability of the coordinator, I added a Modifier extension to deal with a nullable focusRequester

```kotlin
@Composable
private fun Modifier.optionalFocusRequester(focusRequester: FocusRequester?): Modifier {
    return if (focusRequester != null) {
        this then focusRequester(focusRequester)
    } else this
}
```

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/LocalSceneView.kt
sorenoid, line 1: Suggestions for this file apply equally to SceneView and MapView.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/LocalSceneView.kt
```diff
@@ -163,39 +166,38 @@ public fun LocalSceneView(
...
-
-    // set focus right away so the compose AndroidView "focus interop" is set up correctly.
-    localSceneView.isFocusable = canFocus
-    // TODO: remove when https://devtopia.esri.com/runtime/kotlin/issues/7178 is complete
-    localSceneView.isFocusedByDefault = canFocus
-
+    val calloutFocusRequester = remember { FocusRequester() }
+    val a11yCoordinator = remember(calloutFocusRequester,canFocus, localSceneView) {
+        GeoViewA11yCoordinator( calloutFocusRequester, canFocus, localSceneView)
+    }
```
sorenoid:
```suggestion
    val calloutFocusRequester = remember(canFocus) { FocusRequester() }
    val mapViewScope = remember(canFocus) {
        MapViewScope(mapView, canFocus, calloutFocusRequester)
    }
```

I pulled the coordinator out of the MapView entirely and brought the initialization of the MapViewScope up to the top. That isn't the most important detail. 

At a minimum, the coordinator and the focus requester need to be recreated when `canFocus` changes. The way you had it if the mapView or focusRequested change, we are in a new composition and we'd get a new coordinator anyway. With my changes, since I removed the coordinator from the view composables entirely, I just keyed on the GeoViewScope, since it owns the coordinator.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/LocalSceneView.kt
```diff
@@ -437,5 +439,6 @@ private fun ViewpointHandler(
  *
  * @since 300.0.0
  */
-public class LocalSceneViewScope internal constructor(localSceneView: LocalSceneView) :
-    GeoViewScope(localSceneView)
+public class LocalSceneViewScope internal constructor(
+    localSceneView: LocalSceneView, allyCoordinator: GeoViewA11yCoordinator
+) : GeoViewScope(localSceneView, allyCoordinator)
```
sorenoid: use secondary constructor here.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/LocalSceneView.kt
```diff
@@ -163,39 +166,38 @@ public fun LocalSceneView(
...
                 it.interactionOptions = interactionOptions
                 it.isAttributionBarVisible = isAttributionBarVisible
                 it.selectionProperties = selectionProperties
-                it.isFocusable = canFocus
-                it.isFocusedByDefault = canFocus // TODO: remove when #7178 is complete
             }
         )
 
-        val localSceneViewScope = remember { LocalSceneViewScope(localSceneView) }
+        val localSceneViewScope = remember { LocalSceneViewScope(localSceneView, a11yCoordinator) }
```
sorenoid: I moved this up, not essential. But might help in making sense of my suggestions.

sorenoid (changes requested):
Looks very good @shubham7109! My comments mostly center around preserving backwards compatibility, a suggestion to broaden the scope of the Coordinator's "mandate" from just Callouts to all trailing lambda composable content,  and a few nitpicks.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
shubham7109: I've updated the file with comments to reflect this, I like the way you described it! 
 
> exclusive focus between the GeoView and a content lambda

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
shubham7109: Works well for me without `rootView`, removed to use geoView post :)

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
shubham7109: Removing this made me realize, this coordinator also doesn't need to set `geoView.isFocusable = ..` as we can depend on the AndroidView's `update` block to take care of the current state from the coordinator.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
shubham7109: Good catch, I've updated the GeoViewScope to instead allow a secondary constructor which is now used across all GeoView types. Updated the `.api` file as well.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/LocalSceneView.kt
shubham7109: Updated the PR to allow just `canFocus` to be keyed for both the coordinator and the focus requestor. Testing on my end looks to be working as expected, although may need some time to think though the usages. 

One change I could ask to think about, moving the MapViewScope up to the top may result in behavior changes. We need to pass in a GeoView to the scope but also need to ensure that factory is called using the AndroidView prior to passing the GeoView into the scope. This I believe helps ensure `isMapViewReady` works as intended for the content lambda placement. 

This initially was the motivation behind why I introduced a new parameter to the scope (`a11yCoordinator`), which can be initialized prior to help configure the AndroidView and then can be provided later to the scope.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
shubham7109: Nice suggestion, using your feedback I've encapsulated the focusable Modifier configurations into this function which passes in the nullable coordinator.

shubham7109:
@sorenoid Thanks for the feedback, I've added the changes based on the suggestions. Ready for your review :)

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/MapView.kt
```diff
@@ -372,9 +377,9 @@ public fun MapView(
                 }
             })
 
-        val mapViewScope = remember { MapViewScope(mapView) }
-        val isMapViewReady = mapView.rememberIsReady()
+        val mapViewScope = remember { MapViewScope(mapView, a11yCoordinator) }
```
sorenoid:
```suggestion
        val mapViewScope = remember(canFocus) { MapViewScope(mapView, a11yCoordinator) }
```

sorenoid: same for all 3. If canFocus changes, then the a11yCoordinator needs to be reinstantiated to update its logic. Since the coordinator is owned by the MapViewScope, it also needs to update its a11yCoordinator property.

sorenoid (changes requested):
one last leftover

shubham7109:
@sorenoid Applied the suggestion and added few design changes: 

- Updated isGeoViewFocusable to be a mutable state, for compose to pass state changes to AndroidView. Although for `onContentDisposed`, we may want to explicitly set `geoView.isFocusable` prior to click in case if compose updates after post. 
- Added internal `preferredTargetFocusRequester` and public `Modifier.preferredContentFocusTarget()` to allow for user driven target modifier config, allowing us to control the request timing while giving user flexibly to traverse within in their app UI layout tree. 
- Updated app PR to show the usage.  

Feel free to provide any additional feedback

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
```diff
@@ -54,24 +56,42 @@ internal class GeoViewA11yCoordinator(
      */
     private val isContentBeingDisplayed = AtomicBoolean(false)
 
+    /**
+     * The preferred target element for accessibility focus when content is displayed.
```
sorenoid: can you give an example in this doc? something like `for example, the TextView title within a Callout.`

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
```diff
@@ -54,24 +56,42 @@ internal class GeoViewA11yCoordinator(
...
+
     /**
      * The main boolean condition to determine if the GeoView (not trailing content)
      * should be focusable and interact with screen readers.
      */
-    internal val isGeoViewFocusable: Boolean
-        get() = canFocus && !isContentBeingDisplayed.get()
+    private val _isGeoViewFocusable = mutableStateOf(canFocus && !isContentBeingDisplayed.get())
+    internal val isGeoViewFocusable: State<Boolean>
+        get() = _isGeoViewFocusable
```
sorenoid: I think this can be done in one line, just setting 

`internal val isGeoViewFocusable: State<Boolean> = _isGeoViewFocusable`

sorenoid (approved):
Nice work!

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -105,6 +106,18 @@ import kotlin.math.sin
  */
 public sealed class GeoViewScope protected constructor(private val geoView: GeoView) {
```
gunt0001: Instead of defining a secondary constructor as you did below in line 116 and a nullable, mutable property for `a11yCoordinator` in line 114, you could just modify this primary constructor and add a non-nullable, immutable `a11yCoordinator` property. That way you can also get rid of a bunch of null checks where referencing `a11yCoordinator`.

```suggestion
public sealed class GeoViewScope protected constructor(private val geoView: GeoView, internal val a11yCoordinator: GeoViewA11yCoordinator) {
```

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -435,19 +443,62 @@ public sealed class GeoViewScope protected constructor(private val geoView: GeoV
                                 minSize = shapes.minSize
                             )
                             .animateContentSize()
-                            .focusRequester(focusRequester)
-                            .focusable()
-                            .semantics(mergeDescendants = true) {
-                                liveRegion = LiveRegionMode.Polite
```
gunt0001, line 441: Is it ok to remove the `liveRegion` configuration? Do we expect users to set that by themselves?

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -435,19 +443,62 @@ public sealed class GeoViewScope protected constructor(private val geoView: GeoV
...
-                            },
+                            .optionalA11yFocusable(a11yCoordinator)
+                            .semantics {
+                                testTag = "CalloutContainerLayout"
+                            }
                     ) {
                         content.invoke(this)
+                        a11yCoordinator?.let { coordinator ->
```
gunt0001: This null check should be redundant with the changes to the primary constructor. There are other places where null check will be redundant but will onl comment here.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -435,19 +443,62 @@ public sealed class GeoViewScope protected constructor(private val geoView: GeoV
...
+                            DisposableEffect(coordinator) {
```
gunt0001: Assuming `coordinator` can never change, the key should be redundant here:

```suggestion
                            DisposableEffect(Unit) {
```

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -435,19 +443,62 @@ public sealed class GeoViewScope protected constructor(private val geoView: GeoV
...
     }
 
+    /**
+     * Marks a composable inside the GeoView trailing lambda content as the preferred initial focus
+     * target when toolkit accessibility focus handoff moves focus from the GeoView into the
+     * content. This is useful for content such as a Callout where initial focus should land on the
+     * first meaningful child rather than the underlying GeoView.
+     *
+     * If no preferred content focus target is provided, focus falls back to the Callout container.
+     * This modifier only has an effect when the GeoView composable's `canFocus` parameter is true.
```
gunt0001, line 466: Could we add a code snippet here that shows an example of usage?

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -435,19 +443,62 @@ public sealed class GeoViewScope protected constructor(private val geoView: GeoV
...
+     *
+     * @since 300.0.0
+     */
+    public fun Modifier.preferredContentFocusTarget(): Modifier {
+        val coordinator = a11yCoordinator
+        return if (coordinator?.canFocus == true) {
+            this
+                .focusRequester(coordinator.preferredTargetFocusRequester)
+                .focusable()
```
gunt0001, line 486: Should `focusGroup` be applied here optionally?

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
```diff
@@ -435,19 +443,62 @@ public sealed class GeoViewScope protected constructor(private val geoView: GeoV
...
+        } else {
+            this
+        }
+    }
+    /**
+     * Optional accessibility focus modifier configurations applied to the Callout container
+     * using the [a11yCoordinator].
+     */
+    private fun Modifier.optionalA11yFocusable(coordinator: GeoViewA11yCoordinator?): Modifier {
```
gunt0001: Why "optional"?

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
```diff
@@ -0,0 +1,117 @@
...
+ * @param canFocus User provided property to set underlying focus behavior.
+ * @param contentFocusRequester FocusRequester used by trailing lambda content to request focus.
+ *
+ * @since 300.0.0
+ */
+@Stable
+internal class GeoViewA11yCoordinator(
+    internal val geoView: View,
+    internal val canFocus: Boolean,
+    internal val contentFocusRequester: FocusRequester
```
gunt0001: Is there a need to pass `contentFocusRequester` as an argument to this constructor? It doesn't seem to be used outside of this class, so could be a private property?

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
```diff
@@ -0,0 +1,117 @@
...
+    init {
+        // Set focus right away so initial compose AndroidView "focus interop" is set up correctly.
+        geoView.isFocusable = canFocus
+    }
+
+    /**
+     * Tracks whether content from the related GeoViewScope is currently being displayed.
+     * Used to drive focus away from GeoView and into content when displayed.
+     */
+    private val isContentBeingDisplayed = AtomicBoolean(false)
```
gunt0001, line 49: Is there a need for this flag? Looks like it's only used in `onContentComposed`, which should only be called when the content lambda enters the composition. Since you reset this flag when leaving the composition, I wonder what this is guarding.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
```diff
@@ -0,0 +1,117 @@
...
+    internal fun onContentDisposed() {
+        isContentBeingDisplayed.set(false)
+        // Continue if user provided property is true
+        if (!canFocus) return
+        // Set focusable state to propagate an `update` to the AndroidView.
+        _isGeoViewFocusable.value = true
+        // After AndroidView post, send accessibility focus from content lambda back to GeoView.
+        geoView.post {
+            // Update isFocusable attribute in case if post was called before mutable state triggered an `update`.
+            geoView.isFocusable = true
```
gunt0001, line 127: Is this needed? It looks like setting `_isGeoViewFocusable.value` is taking care of propagating the focusable state back into the GeoView?

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/LocalSceneView.kt
```diff
@@ -163,39 +165,38 @@ public fun LocalSceneView(
...
                 it.interactionOptions = interactionOptions
                 it.isAttributionBarVisible = isAttributionBarVisible
                 it.selectionProperties = selectionProperties
-                it.isFocusable = canFocus
-                it.isFocusedByDefault = canFocus // TODO: remove when #7178 is complete
             }
         )
 
-        val localSceneViewScope = remember { LocalSceneViewScope(localSceneView) }
+        val localSceneViewScope = remember(canFocus) { LocalSceneViewScope(localSceneView, a11yCoordinator) }
```
gunt0001: Is this ok to create a new scope instance whenever `canFocus` changes? Could you test what happens when a callout is currently displayed and then `canFocus` changes? I would expect that the callout remains on the screen but I assume it will disappear if we create a new scope object here?

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
shubham7109: This may have issues for backwards compatibility, see Soren's [previous comment](https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/pull/1092#discussion_r2963296011).

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
shubham7109: `LiveRegionMode.Polite` tells services like Talkback to politely await for all other messages before announcing its region message. Such configurations can be driven by the user for the content.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
shubham7109: Could be added. My thinking was since modifiers are append-able by users, so only provide bare minimum wiring to coordinate the switching of focus between GeoView and a target/Callout container, to then allow users to tailor the layout traversal based on their `content`.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewScope.kt
shubham7109: If this modifier is not used by user (not in their composable), this means `preferredTargetFocusRequester` is not used so the `contentFocusRequester` is used instead to drive default focus switching.  Traversal just works better when user apples a target modifier, so talkback traverses down the app layout tree rather than a Toolkit tree.

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
```diff
@@ -0,0 +1,131 @@
...
+     */
+    internal fun onContentDisposed() {
+        isContentBeingDisplayed.set(false)
+        // Continue if user provided property is true
+        if (!canFocus) return
+        // Set focusable state to propagate an `update` to the AndroidView.
+        _isGeoViewFocusable.value = true
+        // After AndroidView post, send accessibility focus from content lambda back to GeoView.
+        geoView.post {
+            // Update isFocusable attribute in case if post was called before mutable state triggered an `update`.
```
gunt0001: Suggestion to make it clearer:

```suggestion
            // Update isFocusable attribute in case if post was called before mutable state _isGeoViewFocusable triggered an `update`.
```

gunt0001 (approved):

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/LocalSceneView.kt
```diff
@@ -163,39 +163,35 @@ public fun LocalSceneView(
...
         LocalSceneView(context)
     }
-
-    // set focus right away so the compose AndroidView "focus interop" is set up correctly.
-    localSceneView.isFocusable = canFocus
-    // TODO: remove when https://devtopia.esri.com/runtime/kotlin/issues/7178 is complete
-    localSceneView.isFocusedByDefault = canFocus
-
+    val localSceneViewScope = remember { LocalSceneViewScope(localSceneView, canFocus) }
+    LaunchedEffect(canFocus) { localSceneViewScope.a11yCoordinator.syncCanFocus(canFocus) }
```
sorenoid: I think it is safer to make this a `SideEffect` to prevent the syncing of canFocus from interfering with a new value of the parameter. new parameters are evaluated at recomposition and SideEffects are evaluated after recomposition, so they won't toggle each other, the way a LaunchedEffect might.
```suggestion
    if (canFocus != mapViewScope.a11yCoordinator.canFocus) {
        SideEffect { mapViewScope.a11yCoordinator.syncCanFocus(canFocus) }
    }    
```

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * Internal coordinator class to manage accessibility focus and configuration for a
+ * GeoView and mutually exclusive focus with its trailing lambda content.
+ *
+ * Track whether the GeoView can be focused and whether content is currently being displayed.
+ * Either a GeoView or its content (like Callout) can be focused at a given time, but not both.
+ *
+ * @param geoView The GeoView whose accessibility configuration is being managed.
+ * @param canFocus User provided property to set underlying focus behavior.
```
sorenoid, line 36:
```suggestion
 * @property canFocus User provided property from the parameter passed into the GeoView to set underlying focus behavior.
```

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/internal/GeoViewA11yCoordinator.kt
```diff
@@ -0,0 +1,131 @@
...
+
+
+    /**
+     * Used by GeoViewScope to notify here that trailing content has entered the composition.
+     * Update the GeoView to not be focusable and, on [View.post] of the GeoView, request focus
+     * into the trailing content.
+     */
+    internal fun onContentComposed() {
+        if (isContentBeingDisplayed.compareAndSet(false, true)) {
+            // Continue if user provided property is true
```
sorenoid, line 93:
```suggestion
            // Continue if user provided GeoView parameter is true
```

sorenoid (commented):
One final suggestion and nitpick about comments. Very nicely done.

PR #1093, opened 2026-03-06
Rom14514/consume sync test data plugin

rom14514:
Related to issue: [#6649](https://devtopia.esri.com/runtime/kotlin/issues/6649)

### Description:

Introduce a convention plugin to consume `SyncTestDataPlugin` and apply it to `:geoview-compose` for centralised Android integration test setup.

SDK PR: https://devtopia.esri.com/runtime/kotlin/pull/7365

### Summary of changes:

- Consume `SyncTestDataPlugin`
- Create `AndroidIntegrationTestingConventionPlugin`
- Apply the convention plugin to `:geoview-compose`

### Pre-merge Checklist

  - [x] link:https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/1198/
  - [ ] Yes
  - [ ] No


build-logic/build.gradle.kts
```diff
@@ -0,0 +1,12 @@
+plugins {
+    `kotlin-dsl`
+}
+
+gradlePlugin {
+    plugins {
+        create("androidIntegrationTesting") {
```
gunt0001: Could you add some comments here that explain why we need this and how it relates to `androidIntegrationTesting` in `build-logic-internal`.

build-logic-internal/convention/src/main/kotlin/AndroidIntegrationTestingConventionPlugin.kt
```diff
@@ -0,0 +1,70 @@
...
+ * Convention plugin for Android library modules that wire prerequisite tasks for
+ * instrumented integration tests.
+ *
+ * This plugin:
+ * - Registers `grantDevicePermissions` to grant runtime permissions to the test app.
+ * - Wires each `connected<Variant>` task to:
+ *   - optionally sync test data from `:build-logic-internal:syncTestData`,
+ *   - grant device permissions,
+ *   - delete IC output via `:build-logic-internal:deleteICOutput`.
+ *
```
gunt0001, line 39: Could we add here:

```suggestion
 *   - delete IC output via `:build-logic-internal:deleteICOutput`.
 *
 * Note:
 * By depending on tasks such as `syncTestData` and `deleteICOutput` from the `build-logic-internal` 
 * module, Gradle ensures that these tasks are only executed once after project evaluation. This means 
 * that test data sync and deleting IC output is only happening once for multiple modules. 
```

gunt0001 (approved):
@rom14514 - looks good, a few minor comments. I merged `v.next` into your branch and kicked off a vTest.

gradle/libs.versions.toml
```diff
@@ -96,6 +97,8 @@ dokka-versioning = { group = "org.jetbrains.dokka", name = "versioning-plugin",
 arcore = { group = "com.google.ar", name = "core", version.ref = "arcore" }
 play-services-location = { group = "com.google.android.gms", name = "play-services-location", version.ref = "playServicesLocation" }
 androidx-espresso-intents = { group = "androidx.test.espresso", name = "espresso-intents", version.ref = "androidxEspresso" }
+android-gradle-plugin = { group = "com.android.tools.build", name = "gradle", version.ref = "androidGradlePlugin" }
```
sorenoid, line 101: Can you separate these from the other libraries? It can be confusing to see plugins with version in the libraries section. The samples repo does this [here](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/blob/03cef1f19be55549ad216baa0be728224fae5029/gradle/libs.versions.toml#L99)

build-logic/build.gradle.kts
sorenoid: yes that would be helpful, I am not sure I understand what is going on here.

sorenoid (changes requested):
Looks ok, I need a bit more clarification on why the internal vs "normal" build-logic folders are needed.

build-logic/build.gradle.kts
rom14514: @gunt0001 @sorenoid I have added some comments to `settigns.gradle.kts`, `build-logic/build.gradle.kts` and `build-logic-internal/build.gradle.kts`

gradle/libs.versions.toml
rom14514: Good point, I have updated the PR to use the same approach as in the samples repo

settings.gradle.kts
```diff
@@ -17,7 +17,60 @@
...
+    // Set toolkitTestDir here because pluginManagement is evaluated before the rest of the
+    // settings.gradle.kts and build.gradle.kts files, so the variables defined in those files are not accessible here.
+    val toolkitTestDir = settingsDir.absolutePath.removeSuffix("arcgis-maps-sdk-kotlin-toolkit") + "/kotlin/toolkit-tests"
+
+    gradle.beforeProject {
+        extra.set("toolkitTestDir", toolkitTestDir)
+    }
+
+    // Include build-logic-internal for internal developers. It must be included before
+    // build-logic because build-logic contains placeholder plugins that are replaced by
```
sorenoid, line 29: Naively from looking at the code below, it looks like the empty plugin will replace the real plugin. Is the replacement behavior observed in testing or is it documented? If it doesn't break things, I suggest wrapping the public build logic include in an `else` clause.

sorenoid (commented):
I have just one further comment, but have asked @shubham7109 to look at this to ensure  it is in sync with his plans for convention plugins.

settings.gradle.kts
rom14514: The behaviour was observed in testing, unfortunately I could not find documentation for it. Including `build-logic` in an `else` block works today and wont break anything, but if we ever need a plugin to be accessible to external developers and live in `build-logic`, we would have to remove the `else` clause

build-logic-internal/convention/src/main/kotlin/AndroidIntegrationTestingConventionPlugin.kt
```diff
@@ -0,0 +1,93 @@
...
+ * that test data sync and deleting IC output is only happening once for multiple modules.
+ *
+ * Expected project setup:
+ * - `com.android.library` is applied on the target module.
+ * - Gradle property `syncTestDataBeforeInstrumentedTests` is defined (`true`/`false`).
+ */
+class AndroidIntegrationTestingConventionPlugin : Plugin<Project> {
+    override fun apply(target: Project) {
+        with(target) {
+            pluginManager.withPlugin("com.android.library") {
```
shubham7109, line 52: Could we change this to use the libs toml `android-library` here

build-logic-internal/build.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+tasks.named("grantDevicePermissions") {
+    doFirst {
+        error(
+            "This task is disabled because :build-logic-internal is not an Android module and has no testApplicationId. " +
+                    "Run a module-specific grantDevicePermissions task instead (e.g. :geoview-compose:grantDevicePermissions)."
+        )
+    }
+}
```
shubham7109, line 52: I'm struggling to understand when does this come into play? Is it needed since we use the alias `libs.plugins.arcgismaps.testdata.sync` plugin in this file to exclude `:build-logic-internal:grantDevicePermissions`?

build-logic-internal/build.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+
+// All plugins created in this module should have a corresponding placeholder plugin
+// with the same ID in build-logic. build-logic-internal is not included for
+// non-internal developers, so they can’t access these plugins and applying a missing
+// plugin would fail their builds. The placeholder lets them compile, while internal
+// developers get the real implementation from build-logic-internal.
+gradlePlugin {
+    plugins {
+        create("androidIntegrationTesting") {
```
shubham7109: Maybe try to see if you could register these plugins within the toml, not sure it would impact public users but could be much cleaner. 
```suggestion
        register("androidIntegrationTesting") {
```

build-logic-internal/gradle.properties
```diff
@@ -0,0 +1,25 @@
...
+#
+#
+
+# Relative path from the root of the Kotlin repo which should be synced as test data to the device
+# Note you should ensure the path ends with a `/` if it already exists on the device. Otherwise
+# the task will create a new directory with the same name inside the existing one.
+testDataSyncPath=test-data/
+
+# Relative path from the build-logic directory to the root of the Kotlin repo
+kotlinDirPath=../../kotlin
```
shubham7109, line 25: Feel free to use/add these setup the internal build configuration as needed. 
```suggestion
kotlinDirPath=../../kotlin
org.gradle.parallel=true
org.gradle.caching=true
org.gradle.configureondemand=true
```

build-logic-internal/convention/src/main/kotlin/AndroidIntegrationTestingConventionPlugin.kt
shubham7109, line 1: To use the standard `:convention` plugin approach, I would recommend to move this file in the folder as follows:  
- `build-logic-internal/convention/src/main/java/AndroidIntegrationTestingConventionPlugin.kt`

build-logic-internal/settings.gradle.kts
```diff
@@ -0,0 +1,106 @@
...
+        gradlePluginPortal()
+    }
+    versionCatalogs {
+        create("libs") {
+            from(files("../gradle/libs.versions.toml"))
+        }
+    }
+}
+
+rootProject.name = "build-logic-internal"
```
shubham7109, line 106: With the above change, you can include the convention plugin finally to this composite build: 
```suggestion
rootProject.name = "build-logic-internal"
include(":convention")
```

build-logic-internal/build.gradle.kts
shubham7109, line 1: Following with above change, the build gradle file also now lives in:
- `build-logic-internal/convention/build.gradle.kts`

build-logic/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
...
+// apply plugins from build-logic-internal (e.g., `android-integration-testing`)
+// would fail to compile.
+//
+// To avoid this, we register a placeholder plugin with the same ID in the
+// publicly accessible build-logic build. This lets non-internal developers
+// compile and work with the code, while internal developers get the real
+// implementation from build-logic-internal.
+gradlePlugin {
+    plugins {
+        create("androidIntegrationTesting") {
```
shubham7109:
```suggestion
        register("androidIntegrationTesting") {
```

gradle/libs.versions.toml
```diff
@@ -96,6 +97,10 @@ dokka-versioning = { group = "org.jetbrains.dokka", name = "versioning-plugin",
...
+
+# Dependencies of the included build-logic-internal
+arcgismaps-testdata-sync-gradle-plugin = { group = "com.esri.gradle-plugins.testdata.sync", name = "com.esri.gradle-plugins.testdata.sync.gradle.plugin", version.ref = "arcgismapsTestdataSync" }
```
shubham7109: Suggestion to bring the gradle plugin in the comment block: 

```suggestion
# Dependencies of the included build-logic-internal
android-gradle-plugin = { group = "com.android.tools.build", name = "gradle", version.ref = "androidGradlePlugin" }
arcgismaps-testdata-sync-gradle-plugin = { group = "com.esri.gradle-plugins.testdata.sync", name = "com.esri.gradle-plugins.testdata.sync.gradle.plugin", version.ref = "arcgismapsTestdataSync" }
```

gradle/libs.versions.toml
```diff
@@ -109,6 +114,7 @@ gradle-secrets = { id = "com.google.android.libraries.mapsplatform.secrets-gradl
 kotlin-serialization = { id = "org.jetbrains.kotlin.plugin.serialization", version.ref = "kotlin" }
 compose-compiler = { id = "org.jetbrains.kotlin.plugin.compose", version.ref = "kotlin" }
 gmazzo-test-aggregation = { id = "io.github.gmazzo.test.aggregation.results", version.ref = "gmazzo" }
+arcgismaps-testdata-sync = { id = "com.esri.gradle-plugins.testdata.sync", version.ref = "arcgismapsTestdataSync" }
```
shubham7109, line 116: You can now define the plugin in the toml to be used as an alias in modules:


```suggestion
arcgismaps-testdata-sync = { id = "com.esri.gradle-plugins.testdata.sync", version.ref = "arcgismapsTestdataSync" }
# Plugins defined by the project build-logic
android-integration-testing = { id = "android-integration-testing" }
```

toolkit/geoview-compose/build.gradle.kts
```diff
@@ -21,9 +21,10 @@ plugins {
     id("org.jetbrains.kotlin.android")
     id("org.jetbrains.kotlin.plugin.compose")
     id("artifact-deploy")
-    id("grant-test-permissions")
+    id("android-integration-testing")
```
shubham7109:
```suggestion
    alias(libs.plugins.android.integration.testing)
```

shubham7109 (changes requested):
Really nice work! Added some minor feedback comments:

build-logic-internal/convention/src/main/kotlin/AndroidIntegrationTestingConventionPlugin.kt
rom14514: It is possible, but I am not sure it is better because we cant use libs the same way we do in `build.gradle`. To use the `android-library` alias from the version catalog, it would look something like this:
``` kotlin
val libs = extensions
    .getByType(VersionCatalogsExtension::class.java)
    .named("libs")
val androidLibraryPluginId = libs.findPlugin("android-library").get().get().pluginId
pluginManager.withPlugin(androidLibraryPluginId) {
```

build-logic-internal/build.gradle.kts
rom14514: This is not required, everything will work the same without it. It would still fail if someone calls `:build-logic-internal:grantDevicePermissions`, but the error message is not helpful:
```
Execution failed for task ':build-logic-internal:grantDevicePermissions'.
> Process 'command '/Users/rom14514/Library/Android/sdk/platform-tools/adb'' finished with non-zero exit value 1
```
I added this so that if someone does call `:build-logic-internal:grantDevicePermissions`, they get a clear message about what to do instead.

build-logic-internal/gradle.properties
rom14514: The main Toolkit build doesnt have those, so I am hesitant to include them here. This might be best handled as a separate issue

build-logic-internal/convention/src/main/kotlin/AndroidIntegrationTestingConventionPlugin.kt
shubham7109: Understandable and should be okay for now, we can potentially place in a `libs` [project extension](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/blob/main/build-logic/convention/src/main/java/com/esri/arcgismaps/kotlin/build_logic/convention/ProjectExtensions.kt#L25) for all plugins to use in the future.

build-logic-internal/build.gradle.kts
rom14514: I am not sure what you mean by ‘register these plugins within the toml.’ Do you mean defining the plugin IDs in the toml and then consuming them via alias?

shubham7109: Yes, apologies for not being clear. Consuming via aliases would be ideal.

shubham7109 (approved):
Thanks for adding the changes, looks good to me 👍

sorenoid (approved):

rom14514:
Thanks for the reviews! Merging now

PR #1097, opened 2026-03-16
Expose analysis overlays on map view

rom14514:
Related to issue: #7089

### Description:

Expose `analysisOverlays` and `onAnalysisViewStatusChanged` on the composable `MapView` and `SceneView`, and add `getAnalysisViewStatus` to `GeoViewProxy`.

### Summary of changes:

- Expose `analysisOverlays` and `onAnalysisViewStatusChanged` on `MapView`
- Expose `onAnalysisViewStatusChanged` on `SceneView`
- Expose `getAnalysisViewStatus ` on `GeoViewProxy`

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/1211/
  - [ ] Yes
  - [ ] No


rom14514:
Note: I duplicated `EventUtils` because we currently can’t use extension functions from `testFixtures`. There’s an [issue](https://devtopia.esri.com/runtime/kotlin/issues/7222) tracking this, I added `EventUtils` there as something to remove once it’s resolved.

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/testUtil/EventsUtil.kt
```diff
@@ -0,0 +1,204 @@
...
+ * Copyright Laws of the United States.
+ *
+ * For additional information, contact:
+ * Environmental Systems Research Institute, Inc.
+ * Attn: Contracts Dept
+ * 380 New York Street
+ * Redlands, California, USA 92373
+ *
+ * email: contracts@esri.com
+ */
```
alan-edi, line 17: This seems to be a copy of the same file in the Kotlin repo. Is that deliberate? I don't see any uses of it from within the toolkit repo.

If it is necessary to have a copy here there will probably be some changes needed, for example:
- Use Apache license in the toolkit repo copy.
- Change all `@since`s to 300.0.0 in the toolkit repo copy.

alan-edi (commented):
@rom14514 Looks good. Just one comment about the presence of `EventsUtil.kt` in the toolkit repo

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/testUtil/EventsUtil.kt
rom14514: Yes, it is deliberate. In Toolkit we use it in new tests like this: `analysisViewStatusChanged.test(collectionTimeout = 20000) {`, the `test` is an extension function from `EventUtils`. Ideally we would want to use the same `EventUtils` from SDK, that is shared with the Toolkit using `testFixtures`, but that is not possible with the current Gradle version. There’s an [issue](https://devtopia.esri.com/runtime/kotlin/issues/7222) tracking this, I added EventUtils there as something to remove once it’s resolved.

I updated it to include the right license and use `@sinces 300.0.0`

toolkit/geoview-compose/src/main/java/com/arcgismaps/toolkit/geoviewcompose/GeoViewProxy.kt
```diff
@@ -317,4 +319,15 @@ public sealed class GeoViewProxy(className: String) {
     public fun getLayerViewState(layer: Layer): LayerViewState? {
         return geoView?.getLayerViewState(layer)
     }
+
+    /**
+     * Retrieves the analysis' status in the view.
+     *
+     * @param analysis An analysis object to get the view status for.
+     * @return A valid [AnalysisViewStatus] or null if the view doesn't contain the given analysis.
```
gunt0001: `null` could also be returned if the proxy's GeoView is not part of the composition:

```suggestion
     * @return a valid [AnalysisViewStatus] or null if the view doesn't contain the given analysis or this proxy's GeoView is not part of the composition
```

Likewise, could you also update the doc for `getLayerViewState()` above as follows:
```
@return the [LayerViewState] of the provided layer, or null if the view doesn't contain the given layer or this proxy's GeoView is not part of the composition
```

gunt0001 (approved):
@rom14514 - looks good. One minor doc comment.

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/testUtil/EventsUtil.kt
```diff
@@ -0,0 +1,206 @@
...
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
+import kotlinx.coroutines.withTimeout
+import org.junit.Assert.fail
+import java.io.Closeable
+
+/**
+ * Test cases on SharedFlows are written as extensions function of this class.
+ *
+ * The test case on the SharedFlow is itself an [extension on SharedFlow][SharedFlow.test], which
```
alan-edi: I know this is just copied from elsewhere, but is there a simple way to make it make more sense?

alan-edi (approved):
@rom14514 Looks good. Just one minor question to consider.

toolkit/geoview-compose/src/androidTest/java/com/arcgismaps/toolkit/geoviewcompose/testUtil/EventsUtil.kt
rom14514: Agreed, I simplified it to `"Test helper for collecting [Flow] emissions in order and asserting them in tests."`. I dont think we need to explain how `Flow<T>.test` works here, since that behaviour is already documented on `Flow<T>.test` itself below. If you are ok with this change, I will make the same change in SDK version of `EventUtil`.

alan-edi: Yep, looks good to me

rom14514:
@alan-edi @gunt0001 Thanks for the reviews! Merging now

PR #1107, opened 2026-03-27
Add `editSummary` support for Popup

shubham7109:
Related to issue: `#7388`

### Description:

PR to place in a summary text at the end of a popup using `Popup.getEditSummary()`.

### Pre-merge Checklist

  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/toolkit/1228
  - [ ]  https://runtime-kotlin.esri.com/view/Release/job/300.0.0/job/vtest/job/toolkit/9/


toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/screens/PopupScreen.kt
```diff
@@ -182,9 +187,33 @@ private fun PopupBody(
...
+                        color = MaterialTheme.colorScheme.onSurfaceVariant
+                    )
+                    HorizontalDivider()
+                }
+            }
+        }
     }
 }
 
+private fun getEditSummary(popup: Popup): String =
```
eri9000: @sorenoid 
Might be better to return null here indicating there's no edit summary.
We could also improve this by logging the exception?
```suggestion
private fun getEditSummaryOrNull(popup: Popup): String? =
```

sorenoid: We don't usually log things in toolkit components.

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/screens/PopupScreen.kt
```diff
@@ -182,9 +187,33 @@ private fun PopupBody(
...
+        }
     }
 }
 
+private fun getEditSummary(popup: Popup): String? =
+    try {
+        popup.getEditSummary()
+    } catch(_: Exception) {
+        null
+    }
```
eri9000, line 216:
```suggestion
private fun getEditSummary(popup: Popup): String? = runCatching { popup.getEditSummary() }.getOrNull()
```

eri9000 (approved):
LGTM 👍🏼

toolkit/popup/src/main/java/com/arcgismaps/toolkit/popup/internal/screens/PopupScreen.kt
sorenoid: `runCatching` is an Extension on `Result`, which would be immediately unwrapped. Underneath it all, the code is doing what I implemented, so I think it is simpler just to do that.

PR #1150, opened 2026-08-18
Bump to latest versions

eri9000:
Related to issue: https://devtopia.esri.com/runtime/kotlin/issues/8106

### Description:
Bump dependency versions
- gradle wrapper 9.7.0
- AGP 9.3.2

### Pre-merge Checklist

- [x] https://runtime-kotlin.esri.com/view/vTest/job/vtest/job/toolkit/1332/Test_20Report_20-_20Integration/
  - [ ] Yes
  - [ ] No


gradle/libs.versions.toml
```diff
@@ -17,29 +17,29 @@ binaryCompatibilityValidator = "0.18.1"
...
 datastorePreferences = "1.2.1"
 dokka = "2.2.0"
-hilt = "2.59.2"
+hilt = "2.60.1"
 junit = "4.13.2"
-kotlin = "2.3.21"
-kotlinxGuava = "1.10.2"
+kotlin = "2.4.10"
+kotlinxGuava = "1.11.0"
 ksp = "2.3.7"
```
shubham7109: I would request if Toolkit could use ksp `2.3.10` or `2.3.11` from the [recent release](https://github.com/google/ksp/releases).  

```suggestion
ksp = "2.3.11"
```

eri9000: Updated 👍🏼

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-9.4.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.7.1-bin.zip
```
shubham7109, line 3: Should we match with the SDK here, or update the SDK instead to match this?

build-logic/convention/src/main/kotlin/MicroappConventionPlugin.kt
```diff
@@ -122,8 +122,9 @@ class MicroappConventionPlugin : Plugin<Project> {
                 implementation(libs.findBundle("core").get())
                 implementation(libs.findLibrary("androidx-activity-compose").get())
                 implementation(libs.findLibrary("androidx-lifecycle-viewmodel-compose").get())
+
+                add("androidTestImplementation", libs.findBundle("androidXTest").get())
```
shubham7109: I'm not able to understand why was this added to support bump to latest versions, would like to know why this change was needed.

gradle/wrapper/gradle-wrapper.properties
eri9000: We ideally want to update to the latest version so let's update the SDK

shubham7109 (approved):
I see similar warnings in the kts files listed below. Could these be addressed here, or part or another PR? 

>  'fun Project.provideDelegate(any: Any?, property: KProperty<*>): PropertyDelegate' is deprecated. Use 'val property = project.property(name)' or 'val property = project.findProperty(name)' instead. See the Gradle 9.6 upgrading guide.

- build-logic-internal/build.gradle.kts
- kdoc/build.gradle.kts
- bom/build.gradle.kts

build-logic/convention/src/main/kotlin/MicroappConventionPlugin.kt
eri9000: If this line is not present, the build fails with:
```
ClassNotFoundException: androidx.test.runner.AndroidJUnitRunner
```

I imagine the AndroidJUnitRunner was being provided transitively by some other dependency and now it doesn't.

eri9000: it seems like the line isn't needed anymore. 
I removed it.

sorenoid (approved):

